PR #162, opened 2020-06-25
Add Publish functionality

alerickson:
Adding functionality for Publish-PSResource.

FinderHelper class is pulled directly from Find-PSResource.  I'm going to have Find-PSResource call into this class, but am separating out the PRs to make it easier to read through.

**Note**:  FindHelper is a WIP.  I'm primarily seeking out feedback on PublishPSResource, though feedback on the logic within all files is welcome.

Bare bones publish specs in the RFC for reference: [PSGet RFC](https://github.com/SteveL-MSFT/PowerShell-RFC/blob/08bc32943beeaad9b10a67225bb4bcd90867d80d/2-Draft-Accepted/RFCxxxx-PowerShellGet-3.0.md#publishing-resources)

### DIFF
--- a/src/FindHelper.cs
+++ b/src/FindHelper.cs
@@ -1,0 +1,1325 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+
+using System;
+using System.Management.Automation;
+using System.Collections.Generic;
+using NuGet.Configuration;
+using NuGet.Common;
+using NuGet.Protocol;
+using NuGet.Protocol.Core.Types;
+using System.Threading;
+using LinqKit;
+using MoreLinq.Extensions;
+using NuGet.Versioning;
+using System.Data;
+using System.Linq;
+using System.Net;
+using Microsoft.PowerShell.PowerShellGet.RepositorySettings;
+using System.Net.Http;
+using NuGet.Protocol.Catalog;
+using System.IO;
+using System.Threading.Tasks;
+using Newtonsoft.Json;
+using static System.Environment;
+
+
+namespace Microsoft.PowerShell.PowerShellGet.Cmdlets
+{
+
+    /// <summary>
+    /// Find helper class
+    /// </summary>
+    class FindHelper : PSCmdlet
+    {
+        // This will be a list of all the repository caches
+        public static readonly List<string> RepoCacheFileName = new List<string>();
+        // Temporarily store cache in this path for testing purposes
+        public static readonly string RepositoryCacheDir = Path.Combine(Environment.GetFolderPath(SpecialFolder.LocalApplicationData), "PowerShellGet", "RepositoryCache");
+        //public static readonly string RepositoryCacheDir = @"%APPDATA%/PowerShellGet/repositorycache";//@"%APPDTA%\NuGet";
+        //private readonly object p;
+
+        // Define the cancellation token.
+        CancellationTokenSource source;
+        CancellationToken cancellationToken;
+        List<string> pkgsLeftToFind;
+        private const string CursorFileName = "cursor.json";
+
+        string[] _name;
+        string[] _type;
+        string _version;
+        bool _prerelease = false;
+        string _moduleName;
+        string[] _tags;
+        string[] _repository;
+        PSCredential _credential;
+        SwitchParameter _includeDependencies;
+
+        public FindHelper()
+        { }
+
+        /// <summary>
+        /// </summary>
+        public List<PSObject> beginFindHelper(string[] _name, string[] _type, string _version, bool _prerelease, string _moduleName, string[] _tags, string[] _repository, PSCredential _credential, SwitchParameter _includeDependencies, bool writeToConsole)
+        {
+            this._name = _name;
+            this._type = _type;
+            this._version = _version;
+            this._prerelease = _prerelease;
+            this._moduleName = _moduleName;
+            this._tags = _tags;
+            this._repository = _repository;
+            this._credential = _credential;
+            this._includeDependencies = _includeDependencies;
+       
+
+                source = new CancellationTokenSource();
+            cancellationToken = source.Token;
+
+
+
+            var r = new RespositorySettings();
+            var listOfRepositories = r.Read(_repository);
+
+            var returnedPkgsFound = new List<IEnumerable<IPackageSearchMetadata>>();
+            pkgsLeftToFind = _name.ToList();
+            List<PSObject> returnedPSObjects = new List<PSObject>();
+            foreach (var repoName in listOfRepositories)
+            {
+
+                // We'll need to use the catalog reader when enumerating through all packages under v3 protocol.
+                // if using v3 endpoint and there's no exact name specified (ie no name specified or a name with a wildcard is specified)
+                if (repoName.Properties["Url"].Value.ToString().EndsWith("/v3/index.json") &&
+                    (_name.Length == 0 || _name.Any(n => n.Contains("*"))))//_name.Contains("*")))  /// TEST THIS!!!!!!!
+                {
+                    // right now you can use wildcards with an array of names, ie -name "Az*", "PS*", "*Get", will take a hit performance wise, though.
+                    // TODO:  add wildcard condition here
+                    ProcessCatalogReader(repoName.Properties["Name"].Value.ToString(), repoName.Properties["Url"].Value.ToString());
+                }
+
+
+
+                // if it can't find the pkg in one repository, it'll look in the next one in the list
+                // returns any pkgs found, and any pkgs that weren't found
+                returnedPkgsFound.AddRange(FindPackagesFromSource(repoName.Properties["Url"].Value.ToString(), cancellationToken));
+
+
+                // Flatten returned pkgs before displaying output returnedPkgsFound.Flatten().ToList()[0]
+                var flattenedPkgs = returnedPkgsFound.Flatten();
+                // flattenedPkgs.ToList();
+
+
+                if (flattenedPkgs.Any() && flattenedPkgs.First() != null)
+                {
+                    foreach (IPackageSearchMetadata pkg in flattenedPkgs)
+                    {
+                        //WriteObject(pkg);
+
+
+                        PSObject pkgAsPSObject = new PSObject();
+                        pkgAsPSObject.Members.Add(new PSNoteProperty("Name", pkg.Identity.Id));
+                        pkgAsPSObject.Members.Add(new PSNoteProperty("Version", pkg.Identity.Version));
+                        pkgAsPSObject.Members.Add(new PSNoteProperty("Repository", repoName.Properties["Name"].Value.ToString()));
+                        pkgAsPSObject.Members.Add(new PSNoteProperty("Description", pkg.Description));
+
+                        if (writeToConsole)
+                        {
+                            WriteObject(pkgAsPSObject);
+                        }
+                        else {
+                            returnedPSObjects.Add(pkgAsPSObject);
+                        }
+                    }
+                }
+
+                // reset found packages
+                returnedPkgsFound.Clear();
+
+                // if we need to search all repositories, we'll continue, otherwise we'll just return
+                if (!pkgsLeftToFind.Any())
+                {
+                    break;
+                }
+
+
+
+            } // end of foreach
+
+            return returnedPSObjects;
+
+        }
+
+
+        public void ProcessCatalogReader(string repoName, string sourceUrl)
+        {
+
+            // test initalizing a cursor see: https://docs.microsoft.com/en-us/nuget/guides/api/query-for-all-published-packages
+            // we want all packages published from the beginning of time
+            //       DateTime cursor = DateTime.MinValue;
+
+
+
+            // Define a lower time bound for leaves to fetch. This exclusive minimum time bound is called the cursor.
+            //var cursor = DateTime.MinValue;   /// come back to this GetCursor();
+            var cursor = GetCursor();     // come back to this
+
+            // Discover the catalog index URL from the service index.
+            var catalogIndexUrl = GetCatalogIndexUrlAsync(sourceUrl).GetAwaiter().GetResult();
+
+
+            var httpClient = new HttpClient();
+
+            // Download the catalog index and deserialize it.
+            var indexString = httpClient.GetStringAsync(catalogIndexUrl).GetAwaiter().GetResult();
+            Console.WriteLine($"Fetched catalog index {catalogIndexUrl}.");
+            var index = JsonConvert.DeserializeObject<CatalogIndex>(indexString);
+
+            // Find all pages in the catalog index that meet the time bound.
+
+            var pageItems = index
+                .Items
+                .Where(x => x.CommitTimestamp > cursor);
+
+            var allLeafItems = new List<CatalogLeafItem>();
+
+
+            foreach (var pageItem in pageItems)  /// need to process one page at a time
+            {
+                // Download the catalog page and deserialize it.
+                var pageString = httpClient.GetStringAsync(pageItem.Url).GetAwaiter().GetResult();
+                Console.WriteLine($"Fetched catalog page {pageItem.Url}.");
+                var page = JsonConvert.DeserializeObject<CatalogPage>(pageString);
+
+                // Find all leaves in the catalog page that meet the time bound.
+                var pageLeafItems = page
+                    .Items
+                    .Where(x => x.CommitTimestamp > cursor);
+
+                allLeafItems.AddRange(pageLeafItems);
+
+
+                // local
+                FindLocalPackagesResourceV2 localResource = new FindLocalPackagesResourceV2(sourceUrl);
+
+                LocalPackageSearchResource localResourceSearch = new LocalPackageSearchResource(localResource);
+                LocalPackageMetadataResource localResourceMetadata = new LocalPackageMetadataResource(localResource);
+
+                SearchFilter filter = new SearchFilter(_prerelease);
+                SourceCacheContext context = new SourceCacheContext();
+
+
+
+                // url
+                PackageSource source = new PackageSource(sourceUrl);
+                if (_credential != null)
+                {
+                    string password = new NetworkCredential(string.Empty, _credential.Password).Password;
+                    source.Credentials = PackageSourceCredential.FromUserInput(sourceUrl, _credential.UserName, password, true, null);
+                }
+                var provider = FactoryExtensionsV3.GetCoreV3(NuGet.Protocol.Core.Types.Repository.Provider);
+
+                SourceRepository repository = new SourceRepository(source, provider);
+                PackageSearchResource resourceSearch = repository.GetResourceAsync<PackageSearchResource>().GetAwaiter().GetResult();
+                PackageMetadataResource resourceMetadata = repository.GetResourceAsync<PackageMetadataResource>().GetAwaiter().GetResult();
+
+
+
+
+                // Process all of the catalog leaf items.
+                Console.WriteLine($"Processing {allLeafItems.Count} catalog leaves.");
+                foreach (var leafItem in allLeafItems)
+                {
+                    _version = leafItem.PackageVersion;
+
+                    IEnumerable<IPackageSearchMetadata> foundPkgs;
+                    if (sourceUrl.StartsWith("file://"))
+                    {
+                        foundPkgs = FindPackagesFromSourceHelper(sourceUrl, leafItem.PackageId, localResourceSearch, localResourceMetadata, filter, context).FirstOrDefault();
+                    }
+                    else
+                    {
+                        foundPkgs = FindPackagesFromSourceHelper(sourceUrl, leafItem.PackageId, resourceSearch, resourceMetadata, filter, context).FirstOrDefault();
+                    }
+
+                    // var foundPkgsFlattened = foundPkgs.Flatten();
+                    foreach (var pkg in foundPkgs)
+                    {
+                        // First or default to convert the ienumeraable object into just an IPackageSearchmetadata object
+                        // var pkgToOutput = foundPkgs.FirstOrDefault();//foundPkgs.Cast<IPackageSearchMetadata>();
+                        PSObject pkgAsPSObject = new PSObject();
+                        pkgAsPSObject.Members.Add(new PSNoteProperty("Name", pkg.Identity.Id));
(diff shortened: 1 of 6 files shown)


src/FindHelper.cs
```diff
@@ -0,0 +1,1325 @@
...
+            File.WriteAllText(CursorFileName, cursorString);
+        }
+
+
+
+
+
+    }
+
+    /*
```
JamesWTruher, line 1318: is this needed?

src/FindHelper.cs
```diff
@@ -0,0 +1,1325 @@
...
+
+            return catalogIndexUrl;
+        }
+
+
+
+        ///  Modified this
+        private static DateTime GetCursor()
+        {
+            /*
```
JamesWTruher, line 1271: is this for future?

src/PublishPSResource.cs
```diff
@@ -0,0 +1,791 @@
...
+        /// </summary>
+        [Parameter(Mandatory = true, Position = 0, ValueFromPipeline = true, ValueFromPipelineByPropertyName = true, ParameterSetName = "PathsParameterSet")]
+        [ValidateNotNullOrEmpty]
+        public string Path
+        {
+            get
+            { return _path; }
+
+            set
+            { _path = value; }
```
JamesWTruher: An easy validation you could do here is check to see whether the path exists and is a directory. You could even call `GetResolvedPSPathFromPSPath` to be sure it exists and then validate that it's a directory. I wonder about script resources though. If you provide a directory are all the files in that directory bundled up as the resource.
Is it possible to provide a path to a _file_ rather than a directory?

src/PublishPSResource.cs
```diff
@@ -0,0 +1,791 @@
...
+        {
+            get
+            { return _nuspec; }
+
+            set
+            { _nuspec = value; }
+        }
+        private string _nuspec;
+
+
```
JamesWTruher, line 297: I thought it was going to be possible to publish a `.nupkg` file. will that be coming later?

src/PublishPSResource.cs
```diff
@@ -0,0 +1,791 @@
...
+        Hashtable dependencies = new Hashtable();
+        NuGetVersion pkgVersion = null;
+        bool isScript;
+
+        /// <summary>
+        /// </summary>
+        protected override void ProcessRecord()
+        {
+            string resolvedPath = "";
+
```
JamesWTruher: ah - i see. This should probably be done in the property setter above. That way it will fail in parameter binding rather than after the pipeline gets started.

src/PublishPSResource.cs
```diff
@@ -0,0 +1,791 @@
...
+                moduleFile = System.IO.Path.Combine(resolvedPath, dirName + ".psd1");
+            }
+            else if (File.Exists(resolvedPath) && resolvedPath.EndsWith(".ps1"))
+            {
+                // Pkg to publish is a script
+                moduleFile = resolvedPath;
+                isScript = true;
+            }
+            else {
+                var message = String.Format("A .psd1 file or .ps1 file does not exist in the path '{0}'.", resolvedPath);
```
JamesWTruher: I assume these are going to be resources eventually? it might be better to do that sooner rather than later.

src/PublishPSResource.cs
```diff
@@ -0,0 +1,791 @@
...
+
+            // if user does not specify that they want to use a nuspec they've created, we'll create a nuspec
+            if (string.IsNullOrEmpty(_nuspec))
+            {
+                _nuspec = createNuspec(outputDir, moduleFile, dirName);
+            }
+            else
+            {
+                // Read the nuspec passed in to pull out the dependency information
+                using (StreamReader sr = File.OpenText(_nuspec))
```
JamesWTruher: rather than using string manipulation like this, i would suggest using an XML navigator. You'll be able to determine whether the XML file is a valid file and be able to query for the elements you want.

src/PublishPSResource.cs
```diff
@@ -0,0 +1,791 @@
...
+                        OutputDirectory = outputDirectory, 
+                        Path = _nuspec,
+                        Exclude = _exclude,
+                        Symbols = false,
+                        Logger = NullLogger.Instance
+                    },
+                    MSBuildProjectFactory.ProjectCreator,
+                    builder);
+
+            runner.BuildPackage();
```
JamesWTruher, line 430: when I looked at the NugetClient library it mentions that `BuildPackage()` is obsolete. It may be that they're trying to move away from this API. The suggestion is to use `RunPackageBuild()` instead. This may not show up in your version of the NuGet client, but it looks like it's coming.

src/PublishPSResource.cs
```diff
@@ -0,0 +1,791 @@
...
+                    0, // timeout
+                    false, // disable buffering
+                    false, // no symbols,
+                    false, // no skip duplicate
+                    false, // enable server endpoint
+                    log).GetAwaiter().GetResult();
+        }
+
+
+        private string createNuspec(string outputDir, string moduleFile, string pkgName)
```
JamesWTruher: if you were to take a `FileInfo` for `moduleFile`, you could do simple comparisons with the `.Extension` property. Also, I think you'll have trouble if the extension is capital `.PS1`, etc.

src/PublishPSResource.cs
```diff
@@ -0,0 +1,791 @@
...
+
+                    // metadata for scripts are divided into two parts
+                    string str = String.Empty;
+
+                    // read until the beginning of the metadata is hit "<#PSScriptInfo"
+                    do
+                    {
+                        str = sr.ReadLine().Trim();
+                    }
+                    while (str != "<#PSScriptInfo");
```
JamesWTruher: i don't think these things are case sensitive, so something like `<#PSSCRIPTINFO` won't be found.

src/PublishPSResource.cs
```diff
@@ -0,0 +1,791 @@
...
+                     Description here
+                    .PARAMETER Name
+                    .EXAMPLE
+                     Example cmdlet here
+                    #>
+                */
+
+                pkgName = pkgName.Remove(pkgName.Length - 4);
+
+                using (StreamReader sr = File.OpenText(moduleFile))
```
JamesWTruher: this should probably be refactored into it's own helper function and return a object which you can use. It seems too useful.

src/PublishPSResource.cs
```diff
@@ -0,0 +1,791 @@
...
+                var NoVersionFound = new ErrorRecord(ex, "NoVersionFound", ErrorCategory.InvalidArgument, null);
+
+                this.ThrowTerminatingError(NoVersionFound);
+            }
+            NuGetVersion.TryParse(version, out pkgVersion);
+
+            metadataElementsDictionary.Add("version", pkgVersion.ToNormalizedString());
+
+
+            if (parsedMetadataHash.ContainsKey("author"))
```
JamesWTruher, line 541: the usual pattern is to use `TryGetValue` with a dictionary (rather than a hashtable)
```powershell
$d = [System.Collections.Generic.Dictionary[string,string]]::new()
$d.add("one","1")
$r = $null
if( $d.TryGetValue("one",[ref]$r)) {
... add("authors",$r)
```
or some such

JamesWTruher (commented):
great start - just a few things we can talk about

src/PublishPSResource.cs
alerickson: That's possible with the -destinationpath parameter...  if you register a local repository (just an arbitrary file system path) it'll publish as a .nupkg as well.

src/FindHelper.cs
alerickson: It is... I'm going to remove this file from this PR and submit a PR with parts of find refactored and cleaned up a bit.

src/PublishPSResource.cs
alerickson: I do validate the file path in ProcessRecord on line 273.  Path accepts both a directory (for modules) and a file (for scripts), so it just validates the path first, and then checks if there's a .psd1 available within the path or if the path itself is a path to a .ps1 file.

src/PublishPSResource.cs
alerickson: Ah, okay, I'll make that change.

src/PublishPSResource.cs
alerickson: Agree with both those comments.  Didn't think about passing FileInfo, but that would be much more appropriate here (and elsewhere).

src/PublishPSResource.cs
alerickson: Yup, I'll go through and make sure strings are case insensitive.

src/PublishPSResource.cs
```diff
@@ -0,0 +1,773 @@
...
+            { return _repository; }
+
+            set
+            { _repository = value; }
+        }
+        private string _repository;
+
+
+        /// <summary>
+        /// Can be used to publish the a nupkg locally.
```
daxian-dbw: > the a

Typo.
Today we have to create a local repository to generate the nupkg for it through `publish-module`. Does this parameter allow doing that without creating a local repository beforehand?

src/PublishPSResource.cs
```diff
@@ -0,0 +1,773 @@
...
+        /// </summary>
+        [Parameter()]
+        [ValidateNotNullOrEmpty]
+        public string DestinationPath
+        {
+            get
+            { return _destinationPath; }
+
+            set
+            { _destinationPath = value; }
```
daxian-dbw: No validation on the path. Is the validation done elsewhere?

src/PublishPSResource.cs
```diff
@@ -0,0 +1,773 @@
...
+        [Parameter(Mandatory = true, Position = 0, ValueFromPipeline = true, ValueFromPipelineByPropertyName = true, ParameterSetName = "PathsParameterSet")]
+        [ValidateNotNullOrEmpty]
+        public string Path
+        {
+            get
+            { return _path; }
+
+            set
+            {
+                string resolvedPath = "";
```
daxian-dbw: I don't see where `value` is used in this setter ... did you test this property setter?

src/PublishPSResource.cs
```diff
@@ -0,0 +1,773 @@
...
+            { return _literalPath; }
+
+            set
+            {
+                // Get the .psd1 file or .ps1 file
+                dirName = new DirectoryInfo(value).Name;
+                if (File.Exists(System.IO.Path.Combine(value, dirName + ".psd1")))
+                {
+                    // Pkg to publish is a module
+                    moduleFileInfo = new FileInfo(System.IO.Path.Combine(value, dirName + ".psd1"));
```
daxian-dbw: > System.IO.Path.Combine(value, dirName + ".psd1")

Let's avoid duplicate code by assign this to a local variable.

src/PublishPSResource.cs
```diff
@@ -0,0 +1,773 @@
...
+        public string LiteralPath
+        {
+            get
+            { return _literalPath; }
+
+            set
+            {
+                // Get the .psd1 file or .ps1 file
+                dirName = new DirectoryInfo(value).Name;
+                if (File.Exists(System.IO.Path.Combine(value, dirName + ".psd1")))
```
daxian-dbw: Maybe the check should be
```c#
if (Directory.Exists(value))
{
    ...
}
else if (File.Exists(value) && value.EndsWith(".ps1", ...))
{
   ...
}
```

src/PublishPSResource.cs
```diff
@@ -0,0 +1,773 @@
...
+        [ValidateNotNullOrEmpty]
+        public string LiteralPath
+        {
+            get
+            { return _literalPath; }
+
+            set
+            {
+                // Get the .psd1 file or .ps1 file
+                dirName = new DirectoryInfo(value).Name;
```
daxian-dbw: I may not fully understand the logic -- what if `value` is not a directory path? `dirName` is not reset to null in that case.

src/PublishPSResource.cs
```diff
@@ -0,0 +1,773 @@
...
+                    // Pkg to publish is a module
+                    moduleFileInfo = new FileInfo(System.IO.Path.Combine(value, dirName + ".psd1"));
+                }
+                else if (File.Exists(value) && value.EndsWith(".ps1", StringComparison.OrdinalIgnoreCase))
+                {
+                    // Pkg to publish is a script
+                    moduleFileInfo = new FileInfo(value);
+                    isScript = true;
+                }
+                _literalPath = value;
```
daxian-dbw: same question, is validation needed on `value`? `_literalPath` is set even if `value` is a non-existing path.

src/PublishPSResource.cs
```diff
@@ -0,0 +1,773 @@
...
+
+            set
+            { _releaseNotes = value; }
+        }
+        private string _releaseNotes;
+
+
+        /// <summary>
+        ///  Adds one or more tags to the resource that you are publishing.
+        ///  Note-- this applies only to the nuspec.
```
daxian-dbw, line 192: > ///  Note-- this applies only to the nuspec.

Then does it make sense to have a separate parameter set for this parameter?

src/PublishPSResource.cs
```diff
@@ -0,0 +1,773 @@
...
+
+            set
+            { _tags = value; }
+        }
+        private string[] _tags;
+
+
+        /// <summary>
+        ///  Specifies the URL of licensing terms for the resource you want to publish.
+        ///  Note-- this applies only to the nuspec.
```
daxian-dbw: same question here

src/PublishPSResource.cs
```diff
@@ -0,0 +1,773 @@
...
+
+            set
+            { _licenseUrl = value; }
+        }
+        private string _licenseUrl;
+
+
+        /// <summary>
+        ///  Specifies the URL of an icon for the resource.
+        ///  Note-- this applies only to the nuspec.
```
daxian-dbw: same here.

src/PublishPSResource.cs
```diff
@@ -0,0 +1,773 @@
...
+
+            set
+            { _iconUrl = value; }
+        }
+        private string _iconUrl;
+
+
+        /// <summary>
+        /// Specifies the URL of a webpage about this project.
+        /// Note-- this applies only to the nuspec.
```
daxian-dbw: same here.

src/PublishPSResource.cs
```diff
@@ -0,0 +1,773 @@
...
+        }
+        private string[] _exclude = System.Array.Empty<string>();
+
+
+        /// <summary>
+        /// Specifies a nuspec file rather than relying on this module to produce one.
+        /// </summary>
+        [Parameter()]
+        [ValidateNotNullOrEmpty]
+        public string Nuspec
```
daxian-dbw, line 287: Is this one coupled with all other parameters commented with `Note-- this applies only to the nuspec`?
Does it make sense to have them in a separate parameter set?

src/PublishPSResource.cs
```diff
@@ -0,0 +1,773 @@
...
+            { _nuspec = value; }
+        }
+        private string _nuspec;
+
+
+        Hashtable dependencies = new Hashtable();
+        NuGetVersion pkgVersion = null;
+        bool isScript;
+        FileInfo moduleFileInfo;
+        string dirName;
```
daxian-dbw: I'm not aware of the convention used in this module, but it's inconsistent with other private fields defined in this file.
Please change them to the `private xxx _dirName` format.

src/PublishPSResource.cs
```diff
@@ -0,0 +1,773 @@
...
+        /// </summary>
+        protected override void ProcessRecord()
+        {
+            var resolvedPath = !string.IsNullOrEmpty(_path) ? _path : _literalPath;
+
+            string outputDir = "";
+            // if there's no specified destination path to publish the nupkg, we'll just create a temp folder and delete it later
+            if (!string.IsNullOrEmpty(_destinationPath))
+            {
+                outputDir = SessionState.Path.GetResolvedPSPathFromPSPath(_destinationPath).FirstOrDefault().Path;
```
daxian-dbw: It seems `_path` is resolved in the property setter, so maybe do the same for `_destinationPath`?

src/PublishPSResource.cs
```diff
@@ -0,0 +1,773 @@
...
+            }
+            else
+            {
+                // create temp directory 
+                outputDir = System.IO.Path.Combine(System.IO.Path.GetTempPath(), Guid.NewGuid().ToString());
```
daxian-dbw: > System.IO.Path.GetTempPath()

Please remove the redundant namespace prefix, since `using System.IO;` is declared at the top (search for other instances).

src/PublishPSResource.cs
```diff
@@ -0,0 +1,773 @@
...
+            }
+            else
+            {
+                // Read the nuspec passed in to pull out the dependency information
+                XDocument doc = XDocument.Load(_nuspec);
+
+                // ex: <version>2.2.1</version>
+                var versionNode = doc.Descendants("version");
+                NuGetVersion version;
+                NuGetVersion.TryParse(versionNode.FirstOrDefault().Value, out version);
```
daxian-dbw:
```suggestion
                NuGetVersion.TryParse(versionNode.FirstOrDefault().Value, out NuGetVersion version);
```

src/PublishPSResource.cs
```diff
@@ -0,0 +1,773 @@
...
+            
+            // Push the nupkg to the appropriate repository 
+
+            // to get the pkg version we need to open the .ps1 file or .psd1 file and parse out the version 
+            // if .ps1 file, version gets parsed already
+            // if .psd1 file and we create nuspec, version gets parsed already,
+
+            var fullNupkgPath = System.IO.Path.Combine(outputDirectory, pkgName + "." + pkgVersion.ToNormalizedString() + ".nupkg" );
+
+            var repoURL = repositoryUrl.FirstOrDefault().Properties["Url"].Value.ToString();
```
daxian-dbw: `FirstOrDefault()` is used at a few places in this file. Here (maybe other places too), you have `.Properties["Url"]` directly following it, but be noted that that method may return `null` (default), which will potentially cause NRE. If you know for sure the enumerable is not empty, then use `First()` instead.

Please review other instances too.

src/PublishPSResource.cs
```diff
@@ -0,0 +1,773 @@
...
+            var publishLocation = repoURL.EndsWith("/v2", StringComparison.OrdinalIgnoreCase) ? repoURL + "/package" : repoURL;
+
+            var settings = NuGet.Configuration.Settings.LoadDefaultSettings(null, null, null);
```
daxian-dbw, line 440: > LoadDefaultSettings(null, null, null)

Please use named parameters here, for readability.

src/PublishPSResource.cs
```diff
@@ -0,0 +1,773 @@
...
+                                key = str.Substring(1, str.Length - 1).ToLower();
+                            }
+
+                            try
+                            {
+                                parsedMetadataHash.Add(key, value);
+                            }
+                            catch (Exception e)
+                            {
+                                WriteDebug(String.Format("Failed to add key '{0}' and value '{1}' to hashtable", key, value));
```
daxian-dbw: Is it intended to not throw an error?
Same question goes to all other `WriteDebug` usage.

src/PublishPSResource.cs
```diff
@@ -0,0 +1,773 @@
...
+                 Synopsis description here
+                .DESCRIPTION
+                 Description here
+                .PARAMETER Name
+                .EXAMPLE
+                 Example cmdlet here
+                #>
+            */
+
+            using (StreamReader sr = File.OpenText(moduleFileInfo.FullName))
```
daxian-dbw, line 669: It looks to me you will read the whole file looking for these 2 sections instead of requiring these 2 sections appear at the top of a script. Is that intentional? Also, the parsing logic seems pretty loose on validating the schema of those 2 comment sections, are you going to improve it?

src/PublishPSResource.cs
```diff
@@ -0,0 +1,773 @@
...
+        NuGetVersion pkgVersion = null;
+        bool isScript;
+        FileInfo moduleFileInfo;
+        string dirName;
+
+        /// <summary>
+        /// </summary>
+        protected override void ProcessRecord()
+        {
+            var resolvedPath = !string.IsNullOrEmpty(_path) ? _path : _literalPath;
```
daxian-dbw: nit: you can use `_path` in both `Path` and `LiteralPath` properties, since they are exclusive.

src/PublishPSResource.cs
alerickson: Yes, that's correct.

src/PublishPSResource.cs
alerickson: Yes, the validation is done elsewhere, but I'll move it to the property setter to be congruent with the other parameters.

src/PublishPSResource.cs
alerickson: Ah that's correct, this was a recent change and I hadn't tested.

src/PublishPSResource.cs
alerickson: If it's not a directory path, it will return the name of the file itself.  dirName may not be an adequate variable name, I changed the variable name to pkgFileOrDir.  I refactored a bit so it's no longer in the property setter.

src/PublishPSResource.cs
alerickson: Because of the 'Path' parameter, there's ambiguity between the two and I need to specify System.IO.

src/PublishPSResource.cs
alerickson: I updated this to throw errors.

src/PublishPSResource.cs
alerickson: These two sections are required to be at the beginning of the file. My intention was to read each line until it hits the end of the 2 sections and then exit.  To be quite honest, I'm not sure how to improve the parsing logic.  I need the logic to be backwards compatible with PowerShell script metadata formats.  Currently there's the option of having all metadata within one comment block, or having it split between two.  The logic is a bit funky and I'd certainly like to improve it, but not quite sure how just yet. Do you have any suggestions?

daxian-dbw: I suggest to use `Parser.ParseFile` to get the tokens. The tokens are generated in the order of scanning the file. So once you get the token array, start from the first token, check if its `Kind` property to see if it's a `NewLine` token or a `Comment` token. If the metadata comment blocks are required to be at the top, then the first 2 non-newline tokens must be the comment tokens you want.

You can get the text of a token through token.Text. Then you can take a look at https://github.com/PowerShell/PowerShell/blob/master/src/System.Management.Automation/help/HelpCommentsParser.cs#L712 as an example of parsing comment texts.

PR #216, opened 2020-08-07
Add Documentation 

SydneyhSmith:
Docs Updated:
- Find-PSResource
- Get-PSResourceRepository
- Install-PSResource (some areas need to be finished)


alerickson:
@SydneyhSmith I'm going to close this PR since we're incorporating help docs as we refactor cmdlets, but we can refer back to these docs when pushing the new changes

Issue #220, opened 2020-08-26
Publish-PSResource throws 'InvalidCastException' when parsing manifest with RequiredModules


PaulHigin:
The following line throws:
https://github.com/PowerShell/PowerShellGet/blob/development/src/PublishPSResource.cs#L489
`Unable to cast object of type 'System.Object[]' to type 'System.Collections.Hashtable`

This happens for a PowerShell module package with a manifest (psd1) file has a 'RequiredModules' field containing another hashtable.  This may be a peculiarity of PowerShell Ast.Find where the returned object is sometimes wrapped in an array object, even though a single result is returned.  If this is the case, the code will need to detect and unwrap.

Repro:
```powershell
<#
# TestModule.psd1
@{
    ModuleVersion = '1.0'
    Author = 'paul'
    Description = 'None'
    RootModule = '.\TestModule.psm1'
    FunctionsToExport = @('HelloFunction')
    RequiredModules = @(
        @{ ModuleName = 'PSScriptAnalyzer'; ModuleVersion = '1.18.0' }
    )
}
#>

<#
# TestModule.psm1
function HelloFunction { "Hello" }
#>

Publish-PSResource -APIKey 'none' -Repository local-repository -Path .\TestModule\
```

Results:
```powershell
Publish-PSResource: Unable to cast object of type 'System.Object[]' to type 'System.Collections.Hashtable'
```

Expected:
```powershell
# Success.  No error
```

PR #221, opened 2020-08-27
Fix cast error

PaulHigin:
This fix attempts to address issue #220 by detecting a returned array and unwrapping it.  It might be better to use LanguagePrimitives.ConvertTo(), but it is not clear to me it will do the unwrapping.  But I will try that in a different PR.


PaulHigin:
This is not where the cast error occurred.  Closing.

PR #222, opened 2020-08-27
Publish bug fix

alerickson:
-  Fix bug related to error thrown when 'RequiredModules' in .psd1 is parsed as an object array instead of hashtable (see: issue #220 )
- Add error handling for when repository a resource is attempting to publish to does not exist


alerickson:
Closing, first issue handled in PR #224 ,
second issue handled in PR #225

PR #223, opened 2020-08-27
Fixed bug with Find-PSResource Version param, added tests for Find

anamnavi:
Resolved bug with Find-PSResource Version parameter. Bug occurred if the resource being installed had dependencies for another resource which may not have a specified version, in which case version range would be registered as `(, )` and the `OriginalString` for the version would be null and print an error/warning. Added check to handle this case and its warning.

Also added tests for Find-PSResource. At the moment this only contains test for the Module resource type, and more will be added in.


alerickson (approved):
Looks good!  Thanks for fixing that bug!

PR #229, opened 2020-09-01
Resource repository improvements

alerickson:
* Make repositories unique by name
* Add tab completion for -Name parameter in Get-PSResource, Set-PSResource, and Unregister-PSResource
* Remove credential argument from Register-PSResourceRepository


src/ArgumentCompleter.cs
```diff
@@ -0,0 +1,38 @@
+﻿using Microsoft.PowerShell.PowerShellGet.RepositorySettings;
+using System;
+using System.Collections;
+using System.Collections.Generic;
+using System.Management.Automation;
+using System.Management.Automation.Language;
+
+internal class NameCompleter : IArgumentCompleter
```
rjmholt: I might name this something more specific, like `RepositoryNameCompleter`

src/ArgumentCompleter.cs
```diff
@@ -0,0 +1,38 @@
...
+        return CompleteRepositoryName(wordToComplete);
+    }
+
+
+    private IEnumerable<CompletionResult> CompleteRepositoryName(string wordToComplete)
+    {
+        var res = new List<CompletionResult>();
+
+        var r = new RespositorySettings();
+        var listOfRepositories = r.Read(new string[]{ });
```
rjmholt: Reading the definition of this method, you can save an allocation here by passing `null` or by providing an overload that takes no argument

src/ArgumentCompleter.cs
```diff
@@ -0,0 +1,38 @@
...
+    {
+        return CompleteRepositoryName(wordToComplete);
+    }
+
+
+    private IEnumerable<CompletionResult> CompleteRepositoryName(string wordToComplete)
+    {
+        var res = new List<CompletionResult>();
+
+        var r = new RespositorySettings();
```
rjmholt:
```suggestion
        var repositorySettings = new RespositorySettings();

```

src/ArgumentCompleter.cs
```diff
@@ -0,0 +1,38 @@
...
+        var listOfRepositories = r.Read(new string[]{ });
+
+        foreach (var repo in listOfRepositories)
```
rjmholt:
```suggestion
        foreach (PSObject repo in listOfRepositories)

```

src/ArgumentCompleter.cs
rjmholt:
```suggestion
        IReadOnlyList<PSObject> listOfRepositories = r.Read(new string[]{ });

```

src/ArgumentCompleter.cs
```diff
@@ -0,0 +1,38 @@
...
+         {
+            if ((repo.Properties["Name"].Value.ToString()).StartsWith(wordToComplete, StringComparison.OrdinalIgnoreCase))
+            {
+                res.Add(new CompletionResult(repo.Properties["Name"].Value.ToString()));
+            }
```
rjmholt:
```suggestion
            string repoName = repo.Properties["Name"].Value.ToString();
            if ((repoName.StartsWith(wordToComplete, StringComparison.OrdinalIgnoreCase))
            {
                res.Add(new CompletionResult(repoName));
            }
```

src/ArgumentCompleter.cs
```diff
@@ -0,0 +1,38 @@
...
+        }
+
+        return res;
+    }
+}
```
rjmholt, line 39:
```suggestion
}

```

src/ArgumentCompleter.cs
```diff
@@ -20,16 +20,17 @@ internal class NameCompleter : IArgumentCompleter
 
     private IEnumerable<CompletionResult> CompleteRepositoryName(string wordToComplete)
     {
-        var res = new List<CompletionResult>();
+        List<CompletionResult> res = new List<CompletionResult>();
```
rjmholt, line 23: This `var` makes sense: the type is explicit on the RHS
```suggestion
        var res = new List<CompletionResult>();

```

rjmholt (approved):

PR #233, opened 2020-09-02
Fixed bug with ModuleName param for Find-PSResource, added tests for it

anamnavi:
Bug would arise when using the Find-PSResource cmdlet with ModuleName parameter (and Version parameter along with it) for an incorrect resource type (i.e script resource). It would still return a resource. Resolved this bug and added error message in this current change.


src/FindPSResource.cs
```diff
@@ -840,11 +841,14 @@ private ExpressionStarter<DataRow> BuildPredicate(string repository)
             ExpressionStarter<DataRow> starter2 = PredicateBuilder.New<DataRow>(true);
             if (_moduleName != null)
             {
+                Console.WriteLine("builds predicate for ModuleNAme"); //Anam
```
alerickson: Can you change this to a comment, or a debug statement?

src/FindPSResource.cs
```diff
@@ -841,6 +842,7 @@ private ExpressionStarter<DataRow> BuildPredicate(string repository)
             if (_moduleName != null)
             {
                 predicate = predicate.And(pkg => pkg.Field<string>("Name").Equals(_moduleName));
+                // predicate = predicate.And(pkg => pkg.Field<string>("Type").Equals("Module")); //Anam
```
alerickson: Can you remove this comment as well, unless you think this change will be needed in the future for caching

src/FindPSResource.cs
```diff
@@ -995,26 +996,33 @@ private List<IEnumerable<IPackageSearchMetadata>> FindPackagesFromSourceHelper(s
 
 
 
-            // Check version first to narrow down the number of pkgs before potential searching through tags
+            // Check version first to narrow down the number of pkgs before potentially searching through tags
             if (_version != null)
             {
+                //use either ModuleName or Name (whichever not null) to prevent id error
+                var nameVal = name == null ? _moduleName : name;
```
alerickson: Will this throw any errors/cause any issues if both name and _moduleName are null?

alerickson (changes requested):

src/FindPSResource.cs
anamnavi: Good question! I wasn't sure if we could have a case where Name and ModuleName parameters are not specified as but a Version parameter is specified, and if that would be allowable behavior. I assumed it could be, so in the code I checked for cases where name and _moduleName aren't null, i.e later lines L1009, L1022. 1050 and just added implementation for those cases. So adding code for that first case would be easy if it ends up being allowable behavior.


I realize that behavior of name and _moduleName being null but version isn't null probably isn't meaningful (as it could create a large search and return list), so I've changed the code to not account for this case. And I'll just have one overarching check for if nameVal is null.

alerickson (approved):

anamnavi:
Fixed ModuleName bug

PR #250, opened 2020-09-10
Remaining find tests

anamnavi:
added find tests that mirror v2 tests for Find-Command, Find-DSCResource, Find-RoleCapability, Find-Script. Made request PR changes. I'll create a new PR to switch to using test modules/our own modules instead of production ones and to remove code explaining each test.


test/Pester.CommandFindResourceTests.ps1
```diff
@@ -0,0 +1,203 @@
+# This is a Pester test suite to validate Find-PSResource.
+#
+# Copyright (c) Microsoft Corporation, 2020
```
SteveL-MSFT: Use standard copyright header:

```csharp
// Copyright (c) Microsoft Corporation.
```

test/Pester.CommandFindResourceTests.ps1
```diff
@@ -0,0 +1,203 @@
...
+
+Import-Module "$psscriptroot\PSGetTestUtils.psm1" -WarningAction SilentlyContinue -force
+Import-Module "C:\code\PowerShellGet\src\bin\Debug\netstandard2.0\publish\PowerShellGet.dll" -force
```
SteveL-MSFT: This path is hardcoded to your computer and will fail on other systems.  Importing the newly built module should be part of `PSGetTestUtils` module.  Also the import should be on the psd1 and not the dll directly.

test/Pester.CommandFindResourceTests.ps1
```diff
@@ -0,0 +1,203 @@
...
+
+
+$PSGalleryName = 'PSGallery'
+$PSGalleryLocation = 'https://www.powershellgallery.com/api/v2'
```
SteveL-MSFT: These should also be defined in the test module as I expect many tests to use this info

test/Pester.CommandFindResourceTests.ps1
```diff
@@ -0,0 +1,203 @@
...
+
+$PoshTestGalleryName = 'PoshTestGallery'
+$PostTestGalleryLocation = 'https://www.poshtestgallery.com/api/v2'
+
+
+$TestLocalDirectory = 'TestLocalDirectory'
+$tmpdir = Join-Path -Path ([System.IO.Path]::GetTempPath()) -ChildPath $TestLocalDirectory
```
SteveL-MSFT: Pester has a `$TestDrive` you should use instead of directly using the temp folder as it gets cleaned up after test is finished

test/Pester.CommandFindResourceTests.ps1
```diff
@@ -0,0 +1,203 @@
...
+        $repo.Priority | Should be 50
+    }
+
+    # Purpose: to register PoshTestGallery resource repo and check it registered successfully
+    #
+    # Action: Register-PSResourceRepository PoshTestGallery -URL https://www.poshtestgallery.com/api/v2 -Trusted
+    #
+    # Expected Result: PoshTestGallery resource repo has registered successfully
+    It 'register the Poshtest Repository When -URL is a Website and Installation Policy is Trusted' {
+        # Register-PSResourceRepository $PoshTestGalleryName -URL $PostTestGalleryLocation -Trusted
```
SteveL-MSFT: Remove commented out code that is not used

test/Pester.CommandFindResourceTests.ps1
```diff
@@ -0,0 +1,203 @@
...
+
+        $repo = Get-PSResourceRepository $PoshTestGalleryName
+        $repo.Name | should be $PoshTestGalleryName
```
SteveL-MSFT: Use consistent casing and switch format making the code easier to read:

$repo.Name | Should -Be ...

Applies everywhere in this PR

test/Pester.CommandFindResourceTests.ps1
```diff
@@ -0,0 +1,203 @@
...
+        $repo.URL | should be $PostTestGalleryLocation
+        $repo.Trusted | should be true
```
SteveL-MSFT: Booleans should be `$true` or `$false`, applies everywhere.  But when using Pester, it should be: `Should -BeTrue` or `Should -BeFalse`

test/Pester.CommandFindResourceTests.ps1
```diff
@@ -0,0 +1,203 @@
...
+        $res = Find-PSResource -Name NonExistantCommand
+        $res | Should -BeNullOrEmpty
+    }
+
+    # Purpose: find Command resource with exact version, given Version parameter -> [4.3.0.0]
+    #
+    # Action: Find-PSResource -Name Az.Compute -Version "[4.3.0.0]"
+    #
+    # Expected Result: should return Az.Compute resource with version 4.3.0.0
+    It "find Command resource with exact version, given Version parameter -> [4.3.0.0]" {
```
SteveL-MSFT: Many of these test cases are very similar, you should use `-TestCases`.  See https://github.com/PowerShell/PowerShell/blob/master/test/powershell/engine/Basic/Attributes.Tests.ps1 for an example

test/Pester.CommandFindResourceTests.ps1
```diff
@@ -0,0 +1,203 @@
...
+
+    # Purpose: find Command resource with latest-nonpreview versions, by excluding Prerelease parameter
+    #
+    # Action: Find-PSResource -Name Az.Accounts
+    #
+    # Expected Result: should return latest non-prerelease/non-preview version of Az.Accounts resource
+    It "find Command resource with latest-nonpreview versions, by excluding Prerelease parameter" {
+        $res = Find-PSResource -Name Az.Accounts
+        $res.Name | Should -Be "Az.Accounts"
+        $res.Version | Should -Be "1.9.3.0"
```
SteveL-MSFT: Won't this version change if new version of Az.Accounts get published?

SteveL-MSFT (changes requested):

test/Pester.CommandFindResourceTests.ps1
alerickson: @anamnavi we can use test modules to have a stable module that checks for specific scenarios.  I have a handful already up, but it's easy to publish your own too, we can discuss offline

test/Pester.CommandFindResourceTests.ps1
```diff
@@ -0,0 +1,203 @@
...
+### Find-PSResource ###
+##########################
+Describe 'Test Find-PSResource for Command' {
+    
+    # Purpose: to check if v3 installs the PSGallery repo by default
+    #
+    # Action: Get-PSResourceRepository PSGallery
+    #
+    # Expected Result: Should find that the PSGallery resource repo is already registered in v3
+    It 'find the Default Registered PSGallery' {
```
alerickson: I think this test should go under 'Repository' tests instead of find

test/Pester.CommandFindResourceTests.ps1
```diff
@@ -0,0 +1,203 @@
...
+        $repo.Trusted | should be true
+    }
+
+    # Purpose: find Command resource given Name paramater
+    #
+    # Action: Find-PSResource -Name Az.Compute
+    #
+    # Expected Result: returns Az.Compute resource
+    It "find Command resource given Name parameter" {
+        $res = Find-PSResource -Name Az.Compute
```
alerickson: In general we should be using test modules instead of production modules just for more control and stability

test/PSGetTestUtils.psm1
```diff
@@ -6,6 +6,25 @@
 
 #."$PSScriptRoot\uiproxy.ps1"
 
+# import the .dll:
+$baseP = $PSScriptRoot -split "\\" #PSScriptRoot should be : C:\Users\annavied\Documents\PowerShellGet\test
```
alerickson: Be careful when using specific delimiters because this will change depending on the system... you can use something like `[io.path]::DirectorySeparatorChar` instead

test/PSGetTestUtils.psm1
```diff
@@ -6,6 +6,25 @@
 
 #."$PSScriptRoot\uiproxy.ps1"
 
+# import the .dll:
+$baseP = $PSScriptRoot -split "\\" #PSScriptRoot should be : C:\Users\annavied\Documents\PowerShellGet\test
+$basePWithoutTest = $baseP[0..($baseP.Length-2)] #basePWithoutTest should be: C:\Users\annavied\Documents\PowerShellGet
+$basePWithoutTest = $basePWithoutTest -Join "\"
```
alerickson: Do you need the -Join "\\" if you're running Join-Path below?

test/PSGetTestUtils.psm1
```diff
@@ -6,6 +6,25 @@
 
 #."$PSScriptRoot\uiproxy.ps1"
 
+# import the .dll:
+$baseP = $PSScriptRoot -split "\\" #PSScriptRoot should be : C:\Users\annavied\Documents\PowerShellGet\test
+$basePWithoutTest = $baseP[0..($baseP.Length-2)] #basePWithoutTest should be: C:\Users\annavied\Documents\PowerShellGet
+$basePWithoutTest = $basePWithoutTest -Join "\"
+$script:fullPath = Join-Path $basePWithoutTest -ChildPath "\src\bin\Debug\netstandard2.0\publish\PowerShellGet.dll"
```
alerickson: Also be cautious of the directory separators here

test/Pester.CommandFindResourceTests.ps1
anamnavi: moved.

test/Pester.CommandFindResourceTests.ps1
anamnavi: fixed this.

test/Pester.CommandFindResourceTests.ps1
anamnavi: fixed this.

test/Pester.CommandFindResourceTests.ps1
anamnavi: removed it as my tests don't actually rely on this. Noted what you said about `$TestDrive` for future scenarios though :)

test/PSGetTestUtils.psm1
anamnavi: thanks for pointing this out! I used Join-Path now which uses appropriate directory separator chars

test/PSGetTestUtils.psm1
anamnavi: good point thanks!

test/Pester.CommandFindResourceTests.ps1
anamnavi: will submit this change in a separate PR for better readability

test/CommandFindResource.Tests.ps1
```diff
@@ -0,0 +1,171 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+Import-Module "$psscriptroot\PSGetTestUtils.psm1" -WarningAction SilentlyContinue -Force
```
SteveL-MSFT: I don't think there's a reason to have `-WarningAction` here?

test/CommandFindResource.Tests.ps1
```diff
@@ -0,0 +1,171 @@
...
+
+Describe 'Test Find-PSResource for Command' {
+    
+    # Purpose: find Command resource given Name paramater
+    #
+    # Action: Find-PSResource -Name Az.Compute
+    #
+    # Expected Result: returns Az.Compute resource
+    It "find Command resource given Name parameter" {
+        $res = Find-PSResource -Name Az.Compute
```
SteveL-MSFT: I thought we would use a test module rather than a real module for the tests?  In this case, Az.Compute is likely to be there, but could also be unlisted if they rename or refactor it.

SteveL-MSFT: Or at least use a module that our team owns

test/CommandFindResource.Tests.ps1
```diff
@@ -0,0 +1,171 @@
...
+        $res | Should -BeNullOrEmpty
+    }
+
+    # Purpose: find Command resource with exact version, given Version parameter
+    #
+    # Action: Find-PSResource -Name Az.Compute -Version "[4.3.0.0]"
+    #
+    # Expected Result: should return Az.Compute resource with version 4.3.0.0
+    It "find Command resource given Name to <Reason>" -TestCases @(
+        @{Version="[4.3.0.0]";          ExpectedVersion="4.3.0.0"; Reason="validate version, exact match"},
```
SteveL-MSFT: For completeness, we should have a test case for each of the examples here: https://docs.microsoft.com/en-us/nuget/concepts/package-versioning#version-ranges

test/CommandFindResource.Tests.ps1
```diff
@@ -0,0 +1,171 @@
...
+        $res.Version | Should -Be $ExpectedVersion
+    }
+
+    
+    # Purpose: find Command resource with wildcard version, given Version parameter -> '*' 
+    #
+    # Action: Find-PSResource -Name Az.Compute -Version "*"
+    #
+    # Expected Result: should return all Az.Compute resources (versions in descending order)
+    It "find Command resource given Name to validate version wilcard match " {
```
SteveL-MSFT: Does the wildcard also work with partial versions like 4.2.* or 4.2.*.0 or 4.*.*.0?

test/CommandFindResource.Tests.ps1
```diff
@@ -0,0 +1,171 @@
...
+
+    # Purpose: find Command resource with latest version (including preview versions), with Prerelease parameter
+    #
+    # Action: Find-PSResource -Name Az.Accounts -Prerelease
+    #
+    # Expected Result: should return latest version (including preview versions) of Az.Accounts resource
+    It "find Command resource with latest version (including preview versions), with Prerelease parameter" {
+        $res = Find-PSResource -Name Az.Accounts -Prerelease
+        $res.Name | Should -Be "Az.Accounts"
+        $res.Version | Should -Be "2.0.1.0"
```
SteveL-MSFT: This is unreliable as a future version of preview Az.Accounts would have different version number

test/CommandFindResource.Tests.ps1
```diff
@@ -0,0 +1,171 @@
...
+        $res = Find-PSResource -ModuleName "Az.Compute" -Repository (Get-PoshTestGalleryName)
+        $res.Name | Should -Be "Az.Compute"
+    }
+
+    # Purpose: find command resource when given ModuleName and any Version parameter
+    #
+    # Action: Find-PSResource -Name "Az.Accounts" -Version [2.0.0.0] -Repository "PoshTestGallery"
+    #
+    # Expected Result: should find a command resource when given ModuleName and any version value
+    It "find Command resource when given ModuleName to <Reason>" -TestCases @(
```
SteveL-MSFT: How is this different from the tests in line 34 above?

test/CommandFindResource.Tests.ps1
```diff
@@ -0,0 +1,171 @@
...
+        $res.Name | Should -Be "Az.Compute"
+        $res.Version | Should -Be $ExpectedVersion
+    }
+
+    # Purpose: find Command resource given ModuleName to validate version match
+    #
+    # Action: Find-PSResource -ModuleName Az.Compute -Version "*"
+    #
+    # Expected Result: should return all Az.Compute resources (versions in descending order)
+    It "find Command resource given ModuleName to validate version wildcard match " {
```
SteveL-MSFT: this test also seems like a duplicate?

test/CommandFindResource.Tests.ps1
```diff
@@ -0,0 +1,171 @@
...
+        $res.Count | Should -BeGreaterOrEqual 40
+    }
+
+    # Purpose: find resource with tag, given single Tags parameter
+    #
+    # Action: Find-PSResource -Tags "Azure" -Repository PoshTestGallery | Where-Object { $_.Name -eq "Az.Accounts" }
+    #
+    # Expected Result: should return Az.Accounts resource
+    It "find resource with single tag, given Tags parameter" {
+        $tagValue = "Azure"
```
SteveL-MSFT: Seems better to use test modules particularly when we know exactly how many modules should be returned

test/CommandFindResource.Tests.ps1
```diff
@@ -0,0 +1,171 @@
...
+    # Purpose: find resource with tags, given multiple Tags parameter values
+    #
+    # Action: Find-PSResource -Tags "Azure","Authentication","ARM" -Repository PoshTestGallery | Where-Object { $_.Name -eq "Az.Accounts" }
+    #
+    # Expected Result: should return Az.Accounts resource
+    It "find resource with multiple tags, given Tags parameter" {
+        $tagValue1 = "Azure"
+        $tagValue2 = "Authentication"
+        $tagValue3 = "ARM"
+        $res = Find-PSResource -Tags $tagValue1,$tagValue2,$tagValue3 -Repository (Get-PoshTestGalleryName) | Where-Object { $_.Name -eq "Az.Accounts" }
```
SteveL-MSFT: Unnecessary to have tags defined as variables that are only used once, you can just inline them and quotes not needed if there isn't whitespace

```powershell
$res = Find-PSResource -Tags Azure,Authentication,ARM -Repository (Get-PoshTestGalleryName) | Where-Object { $_.Name -eq "Az.Accounts" }
```

SteveL-MSFT (changes requested):

test/CommandFindResource.Tests.ps1
anamnavi: the tests in line 34 use the `Name` parameter. I added similar tests for the `ModuleName` parameter, because the logic in FindPSResource.cs on this [line ](https://github.com/PowerShell/PowerShellGet/blob/development/src/FindPSResource.cs#L1077) where it detects if a package has a specific version and adds it to the list of packages returned is handled with different logic if a `Name` or `ModuleName` was given. (For `ModuleName` it can only add the package if the package returned was of type module and not script). So I wanted to test both.

anamnavi: the comment above was incorrect, so I fixed that. (the comments will be removed in the next PR though)

test/CommandFindResource.Tests.ps1
anamnavi: Good question. That was in the other v3 tests that I copied the format from, the reason is that without `-WarningAction SilentlyContinue` when you run the Pester tests the warning `WARNING: The names of some imported commands from the module 'PSGetTestUtils' include unapproved verbs that might make them less discoverable. To find the commands with unapproved verbs, run the Import-Module command again with the Verbose parameter. For a list of approved verbs, type Get-Verb.` arises. It's about the naming of method I think Amber created for use in another test file, but I can go ahead and modify that.

test/CommandFindResource.Tests.ps1
anamnavi: Yes, the plan is to use a test module (or module that our team owns) instead of a production one. I was going to change that in the next PR I create today. Should I create those modules and change it now instead?

test/CommandFindResource.Tests.ps1
SteveL-MSFT: You should remove the warning action from this file.  After this PR is merged, I can submit a new PR to fix the verb used so that the warning message doesn't come up.

test/CommandFindResource.Tests.ps1
SteveL-MSFT: Ok, make all those changes in a different PR

test/CommandFindResource.Tests.ps1
anamnavi: I've covered all the cases for Version from this now. For the case of `Minimum version, inclusive`, their syntax is different than the one we use in V3. For version x such that x >= 1.0, They would use `-Version "1.0"` but the V3 syntax to do this would be `-Version "[1.0,)"`.  Doing it their way would give an exact version match in V3. Should I submit a PR to their documentation? Or maybe we make a note about the difference in V3 somewhere in our documentation?

test/CommandFindResource.Tests.ps1
anamnavi: Wildcard implementation for version parameter isn't as expected. The version parsing relies on `NuGetVersion.TryParse()` so I need to look into that and/or add further wildcard implementation and will do so in the next PR.

anamnavi: Currently if someone added `-Version 4.2.*` to their Find-PSResource command they would likely expect it to find an exact version or list of exact versions that start with 4.2 . Instead, it seems to get parsed as a range and the latest version is returned (so if there's a version later than 4.2.* it would return that. For the other 2 cases you listed it successfully prints an error message about those not being valid strings and doesn't return anything.

test/CommandFindResource.Tests.ps1
anamnavi: Agreed. This will be resolved by using our test modules versus production modules in the next PR

test/CommandFindResource.Tests.ps1
anamnavi: while similar to the tests on line 56, this tests the `-ModuleName` parameter and the similar one tests the `-Name` parameter. The logic and implementation for these two parameters are kind of separated in the code so I tried to test both.

test/CommandFindResource.Tests.ps1
anamnavi: Agreed. Will resolve this in next PR.

test/CommandFindResource.Tests.ps1
anamnavi: that makes sense 👍 I incorporated this change.

test/CommandFindResource.Tests.ps1
anamnavi: okay done.

test/CommandFindResource.Tests.ps1
SteveL-MSFT: If that is a change in nuget v3, then please open a doc bug against them

test/CommandFindResource.Tests.ps1
SteveL-MSFT: 4.2.* should definitely be equivalent to maxversion < 4.3.  If this is not currently the case, please open a new issue.

test/CommandFindResource.Tests.ps1
```diff
@@ -0,0 +1,179 @@
...
+        $res | Should -BeNullOrEmpty
+    }
+
+    # Purpose: find Command resource with exact version, given Version parameter
+    #
+    # Action: Find-PSResource -Name Az.Compute -Version "[4.3.0.0]"
+    #
+    # Expected Result: should return Az.Compute resource with version 4.3.0.0
+    It "find Command resource given Name to <Reason>" -TestCases @(
+        @{Version="[4.3.0.0]";          ExpectedVersion="4.3.0.0"; Reason="validate version, exact match"},
```
SteveL-MSFT: Would "4.3.0.0" now be an exact match instead of minimum match or something else?  Seems like that case should be covered even if it is an error case.  Seems like there should be a separate test case for validating incorrectly formatted version string syntax with many test variants.

test/ModuleFindResource.Tests.ps1
```diff
@@ -0,0 +1,207 @@
...
+        @{Version="[1.0.0.0, 2.5.0.0]"; ExpectedVersion="2.5.0.0"; Reason="validate version, exact range inclusive"},
+        @{Version="(1.0.0.0, 2.5.0.0)"; ExpectedVersion="2.0.0.0"; Reason="validate version, exact range exclusive"},
+        @{Version="(1.0.0.0,)";         ExpectedVersion="2.5.0.0"; Reason="validate version, minimum version exclusive"},
+        @{Version="[1.0.0.0,)";         ExpectedVersion="2.5.0.0"; Reason="validate version, minimum version inclusive"},
+        @{Version="(,1.5.0.0)";         ExpectedVersion="1.0.0.0"; Reason="validate version, maximum version exclusive"},
+        @{Version="(,1.5.0.0]";         ExpectedVersion="1.5.0.0"; Reason="validate version, maximum version inclusive"},
+        @{Version="[1.0.0.0, 2.5.0.0)"; ExpectedVersion="2.0.0.0"; Reason="validate version, mixed inclusive minimum and exclusive maximum version"}
+    ) {
+        param($Version, $ExpectedVersion)
+        $res = Find-PSResource -Name "ContosoServer" -Version $Version -Repository (Get-PoshTestGalleryName)
```
SteveL-MSFT: It seems that with PowerShellGet v2, it has a difference in behavior when `-Repository` is explicitly used.  The only difference should be an optimization where PowerShellGet doesn't look through all repositories.  It seems we should have an explicit test for that.

test/ModuleFindResource.Tests.ps1
```diff
@@ -0,0 +1,207 @@
...
+        $res | Should -BeNullOrEmpty
+    }
+
+    # Purpose: find resources when given Name, Version not null --> '*'
+    #
+    # Action: Find-PSResource -Name ContosoServer -Version "*" -Repository PoshTestGallery
+    #
+    # Expected Result: returns 4 ContosoServer resources (of all versions in descending order)
+    It "find resources when given Name, Version not null --> '*'" {
+        $res = Find-PSResource -Name ContosoServer -Version "*" -Repository (Get-PoshTestGalleryName)
```
SteveL-MSFT: Instead of calling `Get-PoshTestGalleryName` within each test, it would be better to have a `BeforeAll{}` block call that helper function and store result in $testGalleryName variable that is used by each test that needs it

test/ModuleFindResource.Tests.ps1
```diff
@@ -0,0 +1,207 @@
...
+    }
+
+    # Purpose: to find a prerelease resource with the prerelease param
+    #
+    # Action: Find-PSResource -Name "PSGTEST-PublishPrereleaseModule-3139" -Prerelease
+    #
+    # Expected Result: without prerelease param latest nonpreview version (or null, if none exists)
+    # should be returned. With it the module should be returned and valid
+    It "find Resource with Prerelease param" {
+        $res = Find-PSResource -Name "PSGTEST-PublishPrereleaseModule-3139"
```
SteveL-MSFT: Does the `3139` mean something?  Seems like we should use a namespace to avoid collisions and make the test module names descriptive without any "magic numbers".  Like `Microsoft.PowerShell.Test.PrereleaseModule`

test/PSGetTestUtils.psm1
```diff
@@ -117,6 +130,23 @@ function Get-PSGetLocalAppDataPath {
...
+}
+
+function Get-PoshTestGalleryName {
+    return $script:PoshTestGalleryName
+}
+
+function Get-PoshTestGalleryLocation {
+    return $script:PostTestGalleryLocation
+}
+
```
SteveL-MSFT, line 209: You don't have to clean the whole file, but we should not have extra whitespace where it's not needed.  Just delete the ones near code you added/modified.

test/RoleCapabilityFindResource.Tests.ps1
```diff
@@ -0,0 +1,141 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+Import-Module "$psscriptroot\PSGetTestUtils.psm1" -force
+
+Describe 'Test Find-PSResource for Role Capability' {
```
SteveL-MSFT: Would be better to have a file name convention like: `FindPSResource.RoleCapability.Tests.ps1` that way if someone wanted to find all the `Find-PSResource` tests they can just do `ls FindPSResource.*` or if they do a listing they would be grouped together if sorted.

SteveL-MSFT (changes requested):
Getting better, almost there :)

test/CommandFindResource.Tests.ps1
anamnavi: yes that would be an exact match. I added tests for invalid versions

test/ModuleFindResource.Tests.ps1
anamnavi: good point, I added this change

test/ModuleFindResource.Tests.ps1
anamnavi: The `3139` doesn't mean anything, I just hadn't thought about removing magic numbers from names but I can see why using a namespace is better practice here. I swapped out this module for a test one for other reasons, but I've made note of using namespaces to make the module name more descriptive and free of random numbers.

test/ModuleFindResource.Tests.ps1
anamnavi: I've added a performance test for this now

test/RoleCapabilityFindResource.Tests.ps1
anamnavi: good point, I hadn't considered file discovery from a user point of view. I made this change.

test/ModuleFindResource.Tests.ps1
anamnavi: performance tests with timing and with Verbose messages will be added in next PR. Currently there is a [test ](https://github.com/anamnavi/PowerShellGet/blob/remaining-find-tests/test/FindResource.Command.Tests.ps1#L202) to check that the repository with higher priority is checked first for a resource. I set the repositories' configurations (including priorities) in the BeforeAll section of the tests.

test/FindResource.Command.Tests.ps1
```diff
@@ -0,0 +1,228 @@
...
+    #
+    # Expected Result: should return Az.Accounts resource
+    It "find resource with single tag, given Tags parameter" {
+        $res = Find-PSResource -Tags "Azure" -Repository $TestGalleryName | Where-Object { $_.Name -eq "Az.Accounts" }
+        $res | Should -Not -BeNullOrEmpty
+        $res.Name | Should -Be "Az.Accounts"
+
+    }
+
+
```
SteveL-MSFT: Extra newline

```suggestion
```

test/PSGetTestUtils.psm1
```diff
@@ -117,9 +128,178 @@ function Get-PSGetLocalAppDataPath {
...
+    $PublishModuleBase = Join-Path $script:TempModulesPath $PublishModuleName
+    $null = New-Item -Path $PublishModuleBase -ItemType Directory -Force
+
+    $PublishModuleBase = Join-Path $script:TempModulesPath $PublishModuleName
+    $version = "1.0"
+    New-ModuleManifest -Path (Join-Path -Path $PublishModuleBase -ChildPath "$PublishModuleName.psd1") -ModuleVersion $version -Description "$PublishModuleName module"  -NestedModules "$PublishModuleName.psm1"
+    Publish-PSResource -path  $PublishModuleBase -Repository psgettestlocal
+}
+
+function RemoveTmpdir
```
SteveL-MSFT: This should not be needed as tests should be using $TestDrive which is auto-deleted

test/FindResource.Module.Tests.ps1
```diff
@@ -0,0 +1,261 @@
...
+    #
+    # Action: Find-PSResource ModuleWithDependencies1 -IncludeDependencies
+    #
+    # Expected Result: should return resource specified and all associated dependecy resources
+    It "find resource with IncludeDependencies parameter" {
+        $res = Find-PSResource ModuleWithDependencies1 -IncludeDependencies -Version "[1.0,2.0]"
+        $res.Count | Should -BeGreaterOrEqual 11
+    }
+
+
```
SteveL-MSFT: Just do a search for all extra newlines and remove

test/PSGetTestUtils.psm1
```diff
@@ -117,9 +128,178 @@ function Get-PSGetLocalAppDataPath {
...
+    return $script:PoshTestGalleryName
+}
+
+function Get-PoshTestGalleryLocation {
+    return $script:PostTestGalleryLocation
+}
 
+function Get-NewPSResourceRepositoryFile {
+    # register our own repositories with desired priority
+    $originalXmlFilePath = Join-Path -Path $env:LOCALAPPDATA -ChildPath "PowerShellGet" -AdditionalChildPath "PSResourceRepository.xml"
```
SteveL-MSFT: That env var doesn't exist on non-Windows.  If the intent of this is to overwrite the psrepository registration file, then it would be better to export the current psrepositories on the system, register the test ones, then on cleanup remove all the repositories and put back the originals.

test/PSGetTestUtils.psm1
```diff
@@ -117,9 +128,178 @@ function Get-PSGetLocalAppDataPath {
...
+    Remove-Item -Path $originalXmlFilePath -Force
+    
+    Copy-Item -Path $tempXmlFilePath -Destination $originalXmlFilePath
+
+    Remove-Item $tempXmlFilePath
+}
+
+function Get-LocalRepoSetUp
+{
+    $repoURLAddress = Join-Path -Path ([System.IO.Path]::GetTempPath()) -ChildPath "testdir"
```
SteveL-MSFT: Use $TestDrive instead of creating own TempPath

test/PSGetTestUtils.psm1
```diff
@@ -117,9 +128,178 @@ function Get-PSGetLocalAppDataPath {
...
+
+    $PublishModuleName = $roleCapName
+    $PublishModuleBase = Join-Path $TempModulesPath $PublishModuleName
+    $null = New-Item -Path $PublishModuleBase -ItemType Directory -Force
+
+    $PublishModuleBase = Join-Path $TempModulesPath $PublishModuleName
+    $version = "1.0"
+
+    New-PSRoleCapabilityFile -Path (Join-Path -Path $PublishModuleBase -ChildPath "$PublishModuleName.psrc")
+    New-ModuleManifest -Path (Join-Path -Path $PublishModuleBase -ChildPath "$PublishModuleName.psd1") -ModuleVersion $version -Description "$PublishModuleName module"  -NestedModules "$PublishModuleName.psm1"
```
SteveL-MSFT:
```suggestion
    New-ModuleManifest -Path (Join-Path -Path $PublishModuleBase -ChildPath "$PublishModuleName.psd1") -ModuleVersion $version -Description "$PublishModuleName module" -NestedModules "$PublishModuleName.psm1"

```

test/PSGetTestUtils.psm1
```diff
@@ -117,9 +128,178 @@ function Get-PSGetLocalAppDataPath {
...
+
+    Get-LocalRepoSetUp
+
+    $PublishModuleName = $dscName
+    $PublishModuleBase = Join-Path $script:TempModulesPath $PublishModuleName
+    $null = New-Item -Path $PublishModuleBase -ItemType Directory -Force
+
+    $PublishModuleBase = Join-Path $script:TempModulesPath $PublishModuleName
+    $version = "1.0"
+    New-ModuleManifest -Path (Join-Path -Path $PublishModuleBase -ChildPath "$PublishModuleName.psd1") -ModuleVersion $version -Description "$PublishModuleName module"  -NestedModules "$PublishModuleName.psm1" -DscResourcesToExport @('DefaultGatewayAddress', 'WINSSetting') -Tags @('PSDscResource_', 'DSC')
```
SteveL-MSFT:
```suggestion
    New-ModuleManifest -Path (Join-Path -Path $PublishModuleBase -ChildPath "$PublishModuleName.psd1") -ModuleVersion $version -Description "$PublishModuleName module" -NestedModules "$PublishModuleName.psm1" -DscResourcesToExport @('DefaultGatewayAddress', 'WINSSetting') -Tags @('PSDscResource_', 'DSC')

```

test/PSGetTestUtils.psm1
```diff
@@ -117,9 +128,178 @@ function Get-PSGetLocalAppDataPath {
...
+    )
+    Get-LocalRepoSetUp
+
+    $PublishModuleName = $cmdName
+    $PublishModuleBase = Join-Path $script:TempModulesPath $PublishModuleName
+    $null = New-Item -Path $PublishModuleBase -ItemType Directory -Force
+
+    $PublishModuleBase = Join-Path $script:TempModulesPath $PublishModuleName
+    $version = "1.0"
+    New-ModuleManifest -Path (Join-Path -Path $PublishModuleBase -ChildPath "$PublishModuleName.psd1") -ModuleVersion $version -Description "$PublishModuleName module"  -NestedModules "$PublishModuleName.psm1" -CmdletsToExport @('Get-Test', 'Set-Test')
```
SteveL-MSFT:
```suggestion
    New-ModuleManifest -Path (Join-Path -Path $PublishModuleBase -ChildPath "$PublishModuleName.psd1") -ModuleVersion $version -Description "$PublishModuleName module" -NestedModules "$PublishModuleName.psm1" -CmdletsToExport @('Get-Test', 'Set-Test')

```

test/testRepositories.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="utf-8"?>
+<configuration>
+  <Repository Name="psgettestlocal"  Url="file:///c:/code/testdir"                  Priority="50" Trusted="false" />
+  <Repository Name="PSGallery"       Url="https://www.powershellgallery.com/api/v2" Priority="50" Trusted="false" />
+  <Repository Name="PoshTestGallery" Url="https://www.poshtestgallery.com/api/v2"   Priority="10" Trusted="true"  />
+</configuration>
```
SteveL-MSFT: Best practice is to have a trailing newline in all text files

SteveL-MSFT (changes requested):

test/FindResource.Command.Tests.ps1
anamnavi: removed

test/PSGetTestUtils.psm1
anamnavi: switched to using $TestDrive and didn't need this so removed it.

test/FindResource.Module.Tests.ps1
anamnavi: removed extra newlines

test/PSGetTestUtils.psm1
anamnavi: switched to using $TestDrive

test/PSGetTestUtils.psm1
anamnavi: made this change

test/PSGetTestUtils.psm1
anamnavi: made this change

test/PSGetTestUtils.psm1
anamnavi: made this change

test/testRepositories.xml
anamnavi: added newline

test/PSGetTestUtils.psm1
anamnavi: Currently we don't have functionality to take an entire snapshot of the repository like this. (We would just be able to save it's name, Repository, Description and Version - but other info like packages published to it, etc can't be saved. Also command to return the repositories returns a PSObject that going to change in the future so I talked to Amber and she suggested that wouldn't be good to rely on.) Instead I used the .NET Environment method. I checked and this works cross platform on Linux/WSL

SteveL-MSFT (approved):

PR #252, opened 2020-09-11
Fix warnings in build and update build script

SteveL-MSFT:
Fix all warnings
Update build script to treat warnings as errors
Fix build script to fail early if build fails
Add `-Clean` switch to build script

`dotnet publish` will do a restore and build so those individual steps aren't needed

### DIFF
--- a/src/InstallHelper.cs
+++ b/src/InstallHelper.cs
@@ -179,7 +179,6 @@ namespace Microsoft.PowerShell.PowerShellGet.Cmdlets
                 {
                     // TODO:  implement after implementing publish
                     throw new Exception("This feature is not yet implemented");
-                    return;
                 }
                 else if (resolvedReqResourceFile.EndsWith(".json"))
                 {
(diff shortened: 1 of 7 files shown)


src/InstallHelper.cs
```diff
@@ -193,7 +192,7 @@ public void ProcessInstallParams(string[] _name, string _version, bool _prerelea
                     {
                         pkgsinJson = JsonConvert.DeserializeObject<Dictionary<string, PkgParams>>(_requiredResourceJson, new JsonSerializerSettings { MaxDepth = 6 });
                     }
-                    catch (Exception e)
```
PaulHigin, line 196: I think we should include the exception message in the reported error message, as it may help understand why conversion fails.

src/PublishPSResource.cs
```diff
@@ -805,7 +805,7 @@ .PARAMETER Name
                         {
                             parsedMetadataHash.Add(key, value);
                         }
-                        catch (Exception e)
```
PaulHigin, line 808: Add exception message to error message?

PaulHigin (changes requested):

src/InstallHelper.cs
SteveL-MSFT: added

src/PublishPSResource.cs
SteveL-MSFT: added

PaulHigin (approved):
LGTM

alerickson (approved):

PR #256, opened 2020-09-14
Fix Register-PSResourceRepository to accept a non-Windows filesystem path

SteveL-MSFT:
Unix file paths don't start with a drive letter, so the Url API fails.  Fix is to use an appropriate URL scheme which is `file://` in this case denoting it as a filesystem path.  This will work on both Windows and Unix systems.

No added test for this as I'm fixing the Register-PSRepository tests in a separate PR which requires more extensive changes.

Also added a luanch.json file so that VSCode can attach the pwsh process to debug PowerShellGet.dll

### DIFF
--- a/src/RegisterPSResourceRepository.cs
+++ b/src/RegisterPSResourceRepository.cs
@@ -59,7 +59,7 @@ namespace Microsoft.PowerShell.PowerShellGet.Cmdlets
                 if (!Uri.TryCreate(value, string.Empty, out url))
                 {
                     // Try the URL as a file path
-                    var resolvedPath = string.Format(CultureInfo.InvariantCulture, "file://{0}", SessionState.Path.GetResolvedPSPathFromPSPath(value.ToString()).FirstOrDefault().Path);
+                    var resolvedPath = string.Format(CultureInfo.InvariantCulture, "{0}{1}{2}", Uri.UriSchemeFile, Uri.SchemeDelimiter, SessionState.Path.GetResolvedPSPathFromPSPath(value.ToString()).FirstOrDefault().Path);
                     if (!Uri.TryCreate(resolvedPath, UriKind.Absolute, out url))
                     {
                         var message = string.Format(CultureInfo.InvariantCulture, "The URL provided is not valid: {0}", value);
--- a/src/RegisterPSResourceRepository.cs
+++ b/src/RegisterPSResourceRepository.cs
@@ -56,16 +56,13 @@ namespace Microsoft.PowerShell.PowerShellGet.Cmdlets
             set
             {
                 Uri url;
-                Uri.TryCreate(value, string.Empty, out url);
-
-                if (url == null)
+                if (!Uri.TryCreate(value, string.Empty, out url))
                 {
                     // Try the URL as a file path
-                    var resolvedPath = string.Format("file://{0}", SessionState.Path.GetResolvedPSPathFromPSPath(value.ToString()).FirstOrDefault().Path);
-                    Uri.TryCreate(resolvedPath, UriKind.Absolute, out url);
-                    if (url == null)
+                    var resolvedPath = string.Format(CultureInfo.InvariantCulture, "file://{0}", SessionState.Path.GetResolvedPSPathFromPSPath(value.ToString()).FirstOrDefault().Path);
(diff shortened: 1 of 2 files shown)


src/RegisterPSResourceRepository.cs
```diff
@@ -60,8 +60,16 @@ public Uri URL
 
                 if (url == null)
```
PaulHigin: It looks like this should be ?
```
if (!Uri.TryCreate(value, string.Empty, out url))
{
...
}

File: src/RegisterPSResourceRepository.cs [hunk 2]
```diff
@@ -60,8 +60,16 @@ public Uri URL
 
                 if (url == null)
                 {
-                    var resolvedPath = SessionState.Path.GetResolvedPSPathFromPSPath(value.ToString()).FirstOrDefault().Path;
-                    Uri.TryCreate(value, resolvedPath, out url);
+                    // Try the URL as a file path
+                    var resolvedPath = string.Format("file://{0}", SessionState.Path.GetResolvedPSPathFromPSPath(value.ToString()).FirstOrDefault().Path);
```
PaulHigin: I believe string.Format should include (CultureInfo.InvariantCulture).

src/RegisterPSResourceRepository.cs
```diff
@@ -60,8 +60,16 @@ public Uri URL
 
                 if (url == null)
                 {
-                    var resolvedPath = SessionState.Path.GetResolvedPSPathFromPSPath(value.ToString()).FirstOrDefault().Path;
-                    Uri.TryCreate(value, resolvedPath, out url);
+                    // Try the URL as a file path
+                    var resolvedPath = string.Format("file://{0}", SessionState.Path.GetResolvedPSPathFromPSPath(value.ToString()).FirstOrDefault().Path);
+                    Uri.TryCreate(resolvedPath, UriKind.Absolute, out url);
```
PaulHigin: if (!Uri.TryCreate()) ?

src/RegisterPSResourceRepository.cs
```diff
@@ -60,8 +60,16 @@ public Uri URL
...
                 if (url == null)
                 {
-                    var resolvedPath = SessionState.Path.GetResolvedPSPathFromPSPath(value.ToString()).FirstOrDefault().Path;
-                    Uri.TryCreate(value, resolvedPath, out url);
+                    // Try the URL as a file path
+                    var resolvedPath = string.Format("file://{0}", SessionState.Path.GetResolvedPSPathFromPSPath(value.ToString()).FirstOrDefault().Path);
+                    Uri.TryCreate(resolvedPath, UriKind.Absolute, out url);
+                    if (url == null)
+                    {
+                        var message = String.Format("The URL provided is not valid: {0}", value);
```
PaulHigin: string.Format should include Culture.InvariantCulture

PaulHigin (changes requested):

src/RegisterPSResourceRepository.cs
TylerLeonhardt: Also:

image

TylerLeonhardt: and

image

PaulHigin (approved):
LGTM

alerickson (approved):

anamnavi (approved):

src/RegisterPSResourceRepository.cs
```diff
@@ -56,12 +56,17 @@ public Uri URL
...
                 Uri url;
-                Uri.TryCreate(value, string.Empty, out url);
-
-                if (url == null)
+                if (!Uri.TryCreate(value, string.Empty, out url))
                 {
-                    var resolvedPath = SessionState.Path.GetResolvedPSPathFromPSPath(value.ToString()).FirstOrDefault().Path;
-                    Uri.TryCreate(value, resolvedPath, out url);
+                    // Try the URL as a file path
+                    var resolvedPath = string.Format(CultureInfo.InvariantCulture, "{0}{1}{2}", Uri.UriSchemeFile, Uri.SchemeDelimiter, SessionState.Path.GetResolvedPSPathFromPSPath(value.ToString()).FirstOrDefault().Path);
```
TylerLeonhardt, line 62: I'm concerned about UNC paths...

PR #257, opened 2020-09-15
Refactor Save-PSResource

alerickson:
Refactor Save-PSResource to call into the same class that Install and Update use.  This allows some added functionality and reduces redundancy.


src/InstallHelper.cs
```diff
@@ -22,6 +22,10 @@
 using NuGet.Protocol;
 using NuGet.Protocol.Core.Types;
 using NuGet.Versioning;
+using NuGet.Packaging;
```
SteveL-MSFT: NIT: These should be in alphabetical order

src/InstallHelper.cs
```diff
@@ -72,91 +80,92 @@ public void ProcessInstallParams(string[] _name, string _version, bool _prerelea
...
-                myDocumentsPath = Path.Combine(Environment.GetFolderPath(SpecialFolder.LocalApplicationData), "powershell");
-                programFilesPath = Path.Combine("/usr", "local", "share", "powershell");
+                // If PS6+ on Windows
+                if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+                {
+                    var id = System.Security.Principal.WindowsIdentity.GetCurrent();
+                    consoleIsElevated = (id.Owner != id.User);
 
-                using (System.Management.Automation.PowerShell pwsh = System.Management.Automation.PowerShell.Create())
+                    myDocumentsPath = Path.Combine(Environment.GetFolderPath(SpecialFolder.MyDocuments), "PowerShell");
```
SteveL-MSFT: I would think these paths are used in many places in PowerShellGet.  If so, should probably have some helper class with a method to retrieve this info and have that helper differentiate Windows/non-Windows and probably a helper to determine if user is root/elevated.  This way the install code is the same whether it's Windows or not.

src/InstallHelper.cs
```diff
@@ -47,22 +52,25 @@ class InstallHelper : PSCmdlet
...
+            cmdletPassedIn.WriteDebug(string.Format("Parameters passed in >>> Name: '{0}'; Version: '{1}'; Prerelease: '{2}'; Repository: '{3}'; Scope: '{4}'; AcceptLicense: '{5}'; Quiet: '{6}'; Reinstall: '{7}'; TrustRepository: '{8}'; NoClobber: '{9}';", string.Join(",", _name), _version != null ? _version : string.Empty, _prerelease.ToString(), _repository != null ? string.Join(",", _repository) : string.Empty, _scope != null ? _scope : string.Empty, _acceptLicense.ToString(), _quiet.ToString(), _reinstall.ToString(), _trustRepository.ToString(), _noClobber.ToString()));
 
-            var consoleIsElevated = false;
-            var isWindowsPS = false;
-            cmdletPassedIn.WriteDebug("Entering InstallHelper::ProcessInstallParams");
+            if (!save)
```
SteveL-MSFT: It seems like the only difference between Install and Save is the destination.  If so, then perhaps it makes sense for the InstallHelper to always do a save, but when used via Install-PSResource it gets passed in the destination depending on the parameters used.

src/InstallHelper.cs
```diff
@@ -72,91 +80,92 @@ public void ProcessInstallParams(string[] _name, string _version, bool _prerelea
...
+                // If no scope is specified (whether or not the console is elevated) default installation will be to CurrentUser
+                // If running under admin on Windows with PowerShell less than PS6, default will be AllUsers
+                if (string.IsNullOrEmpty(_scope))
                 {
-                    _scope = "AllUsers";
+                    // If non-Windows or non-elevated default scope will be current user
+                    _scope = "CurrentUser";
+
+                    // If Windows and elevated default scope will be all users 
+                    if (isWindowsPS && consoleIsElevated)
```
SteveL-MSFT: I'm not sure if this is the desired behavior.  Might be better to require `-Scope AllUsers` to be explicit.

src/InstallHelper.cs
```diff
@@ -72,91 +80,92 @@ public void ProcessInstallParams(string[] _name, string _version, bool _prerelea
...
-                cmdletPassedIn.WriteVerbose(string.Format("Newly created PowerShell scripts path is: {0}", psScriptsPath));
-                Directory.CreateDirectory(psScriptsPath);
-            }
-            if (!Directory.Exists(psInstalledScriptsInfoPath))
-            {
-                cmdletPassedIn.WriteVerbose(string.Format("Newly created PowerShell installed scripts info path  is: {0}", psInstalledScriptsInfoPath));
-                Directory.CreateDirectory(psInstalledScriptsInfoPath);
-            }
+                // Create PowerShell modules and scripts paths if they don't already exist
+                if (!Directory.Exists(psModulesPath))
```
SteveL-MSFT: Since this code is almost the same, you could just have a foreach loop here

SteveL-MSFT (changes requested):
Will wait for initial feedback to be addressed before continuing review

src/InstallHelper.cs
alerickson: That's fair. I'll pull out this section which is all related to particular paths, PS editions, and elevations permissions.  It'll also help make this file specifically about installation.

src/InstallHelper.cs
alerickson: We can change this behavior, I just tried to stay consistent with v2, but it makes more sense for this to be consistent with what happens in PSCore since we have the opportunity to modify it

src/InstallHelper.cs
alerickson: Save also does not include the XML unless specified, and can be saved as a nupkg or unpacked.  I'll think about more at a high level and see which is easier

src/InstallHelper.cs
PaulHigin: I feel a foreach loop is unnecessarily complex here.

src/InstallHelper.cs
```diff
@@ -72,91 +80,92 @@ public void ProcessInstallParams(string[] _name, string _version, bool _prerelea
...
+                {
+                    cmdletPassedIn.WriteVerbose(string.Format("Newly created PowerShell modules path is: {0}", psModulesPath));
```
PaulHigin: Verbose message should be: "Creating PowerShell modules path...", etc.

src/InstallHelper.cs
```diff
@@ -325,9 +334,11 @@ public void ProcessRepositories(string[] packageNames, string version, bool prer
                 if (sourceTrusted || yesToAll)
                 {
                     cmdletPassedIn.WriteDebug("Untrusted repository accepted as trusted source.");
+                    // Check if repository is a local repository
+                    localRepo = repo.Properties["Url"].Value.ToString().StartsWith(Uri.UriSchemeFile + Uri.SchemeDelimiter) ? true : false;
```
anamnavi: just a small thing but is the ternary operator needed for this line? if `repo.Properties["Url"].Value.ToString().StartsWith(Uri.UriSchemeFile + Uri.SchemeDelimiter)` returns `true` or `false` anyways and `localRepo` is of type bool, then I think `localRepo = repo.Properties["Url"].Value.ToString().StartsWith(Uri.UriSchemeFile + Uri.SchemeDelimiter);` could suffice?

src/InstallHelper.cs
```diff
@@ -72,91 +80,92 @@ public void ProcessInstallParams(string[] _name, string _version, bool _prerelea
...
-            if (!Directory.Exists(psInstalledScriptsInfoPath))
-            {
-                cmdletPassedIn.WriteVerbose(string.Format("Newly created PowerShell installed scripts info path  is: {0}", psInstalledScriptsInfoPath));
-                Directory.CreateDirectory(psInstalledScriptsInfoPath);
-            }
+                // Create PowerShell modules and scripts paths if they don't already exist
+                if (!Directory.Exists(psModulesPath))
+                {
+                    cmdletPassedIn.WriteVerbose(string.Format("Newly created PowerShell modules path is: {0}", psModulesPath));
+                    Directory.CreateDirectory(psModulesPath);
```
PaulHigin: File operations can fail in a number of ways.  I wonder if it makes sense to wrap in try/catch here and provide a general error message: "Unable to create <name> with error: <exception error>", for a better user experience?

src/InstallHelper.cs
```diff
@@ -383,10 +403,21 @@ public List<string> InstallPkgs(string repositoryName, string repositoryUrl, Lis
                     // TODO: proper error handling
                     try
                     {
-                        filteredFoundPkgs = (resourceMetadata.GetMetadataAsync(pkgName, prerelease, false, context, NullLogger.Instance, cancellationToken).GetAwaiter().GetResult()
-                            .OrderByDescending(p => p.Identity.Version, VersionComparer.VersionRelease)
-                            .FirstOrDefault());
+                        if (localRepo)
+                        {
+                            var localResource = new FindLocalPackagesResourceV2(repositoryUrl);
+                            localPkgMetadataResource = new LocalPackageMetadataResource(localResource);
```
PaulHigin: It looks like localPkgMetadataResource can be obtained outside the foreach loop.

src/InstallHelper.cs
```diff
@@ -358,14 +369,23 @@ public List<string> InstallPkgs(string repositoryName, string repositoryUrl, Lis
 
             // TODO:  proper error handling here
             PackageMetadataResource resourceMetadata = null;
+            LocalPackageMetadataResource localPkgMetadataResource = null;
```
PaulHigin: Is there a base class that encompasses both PackageMetadataResource and LocalPackageMetadataResource?  It would be nice to create one package resource object since they are both used in the same way.

src/InstallHelper.cs
```diff
@@ -358,14 +369,23 @@ public List<string> InstallPkgs(string repositoryName, string repositoryUrl, Lis
 
             // TODO:  proper error handling here
             PackageMetadataResource resourceMetadata = null;
+            LocalPackageMetadataResource localPkgMetadataResource = null;
             try
             {
-                resourceMetadata = repository.GetResourceAsync<PackageMetadataResource>().GetAwaiter().GetResult();
+                if (localRepo)
+                {
```
PaulHigin: I think a single helper function can be created here that returns 'filteredFoundPkgs' from local or other repos and handles version range as needed.  It would make readability better and handle errors in a uniform way.

src/InstallHelper.cs
```diff
@@ -383,10 +403,21 @@ public List<string> InstallPkgs(string repositoryName, string repositoryUrl, Lis
...
                     {
-                        filteredFoundPkgs = (resourceMetadata.GetMetadataAsync(pkgName, prerelease, false, context, NullLogger.Instance, cancellationToken).GetAwaiter().GetResult()
-                            .OrderByDescending(p => p.Identity.Version, VersionComparer.VersionRelease)
-                            .FirstOrDefault());
+                        if (localRepo)
+                        {
+                            var localResource = new FindLocalPackagesResourceV2(repositoryUrl);
+                            localPkgMetadataResource = new LocalPackageMetadataResource(localResource);
 
+                            filteredFoundPkgs = (localPkgMetadataResource.GetMetadataAsync(pkgName, prerelease, false, context, NullLogger.Instance, cancellationToken).GetAwaiter().GetResult()
```
PaulHigin: Please include comments when using nuget APIs.  How they work is not obvious (to me) and a short explanation would be helpful.

PaulHigin (changes requested):
Just beginning review.  This is pretty complex code so I am glad to see it being refactored (good job!).
Overall I feel the readability can be improved by breaking functional pieces into separate helper methods.
Also please think about errors and how they can be presented to the user that makes sense.
It is not always clear to me when an error should be terminating or just emitted and allow processing to continue.

src/InstallHelper.cs
```diff
@@ -713,226 +806,315 @@ public List<string> InstallPkgs(string repositoryName, string repositoryUrl, Lis
...
+                                    includesFunction.Add(tag.Remove(0, psFunction.Length));
+                                }
+                                else if (tag.StartsWith(psRoleCapability))
                                 {
-                                    pkgsWithCommandClobber.Add(results.Where(pkg => ((ReadOnlyCollection<PSModuleInfo>)pkg.Properties["ExportedCommands"].Value).Where(ec => ec.Name.Equals(command, StringComparison.InvariantCultureIgnoreCase)).Any()));
+                                    includesRoleCapability.Add(tag.Remove(0, psRoleCapability.Length));
                                 }
-                                if (pkgsWithCommandClobber.Any())
+                                else if (!tag.StartsWith("PSWorkflow_") && !tag.StartsWith("PSCmdlet_") && !tag.StartsWith("PSIncludes_")
+                                    && !tag.Equals("PSModule") && !tag.Equals("PSScript"))
```
anamnavi: I'm guessing that because Find-PSResource wasn't working for filtering by the PSScript tag when you search for all packages without a name specified, some tests or cases of install might break here, but once I get the Type parameter figured out I can provide more insight on whether checking for PSScript tag may actually be an issue or not.

anamnavi (approved):

src/InstallHelper.cs
alerickson: Those two variables are redundant so I'm just removing them

alerickson:
There's still more comments I need to address, but I pushed changes to address some of the major issues

PaulHigin (approved):

Issue #217, opened 2020-08-10
Uninstall-PSResouce fails if 'C:\Program Files\PowerShell\Modules' does not exist


ThomasNieto:
PSGet beta 8 fails when trying to uninstall a module if an optional install location does not exist.

```powershell
C:\> Get-Module pester -ListAvailable

    Directory: C:\Users\Thomas\Documents\PowerShell\Modules

ModuleType Version    PreRelease Name                                PSEdition ExportedCommands
---------- -------    ---------- ----                                --------- ----------------
Script     5.0.3                 Pester                              Desk      {Invoke-Pester, Describe, Context, It…}
Script     5.0.2                 Pester                              Desk      {Invoke-Pester, Describe, Context, It…}
Script     4.9.0                 Pester                              Desk      {Describe, Context, It, Should…}

C:\> Uninstall-PSResource Pester-Verbose
Uninstall-PSResource: Could not find a part of the path 'C:\Program Files\PowerShell\Modules'.

C:\> Test-Path "C:\Program Files\PowerShell\Modules"
False
```

alerickson:
Thanks @ThomasNieto!  This will get resolved in the next release

ThomasNieto:
Thanks @alerickson. Might also want to check to see if this issue also occurs on install or any of the checked folders does not exist.

alerickson:
@ThomasNieto yup, it was happening in install as well, will do a more thorough check to make sure this isn't happening elsewhere.

SydneyhSmith:
Closing this as resolved

ThomasNieto:
@alerickson / @SydneyhSmith when was this issue resolved? I still get the error with `Get-PSResource` on beta 10.

```powershell
C:\> Get-PSResource

Get-PSResource: Could not find a part of the path 'C:\Program Files\PowerShell\Modules'.
```

SydneyhSmith:
Thanks @ThomasNieto I can repro this as well, I will re-open the issue, apologies for closing it!

alerickson:
This will be resolved with the release of Preview 11.

PR #272, opened 2020-10-01
Fixes Issue 217

Persistent13:
This fixes the error that occurs when a PSModulePath directory is not present on the system when running Get-PSResource mentioned in #217


alerickson:
Thanks for this PR!  We're generally not actively taking PRs but this is a small add.  If you can resolve the conflict we'll merge this in.

alerickson:
@Persistent13 I'm going to incorporate this change into our Get-InstalledPSResource refactor (see: https://github.com/PowerShell/PowerShellGet/pull/351) and close this PR.  Thanks again for this fix!

PR #297, opened 2020-10-26
Refactor tests testmodules

### LINKED ISSUE #263: Use test modules instead of production ones in tests
Per feedback on this [PR](https://github.com/PowerShell/PowerShellGet/pull/250), refactor tests to use our own test modules instead of production modules. Create those test modules and publish them to PoshTestGallery, but also add code in PSGetTestUtils to detect if these modules still exist before running the individual tests and if not then to publish them again. 

When using own test modules, account for various possible cases of parameters and resource type. Create a black box test plan document/design for these tests.

anamnavi:
Refactored the Find tests to use test modules instead of production modules. Since the team has control and ownership over these test modules, testing for an exact version or anything else that is dependent upon the package not being updated is less error prone than before. This makes the tests more reliable as a failing test would indicate an error with the code, not dependent packages having changed. Resolves #263


test/FindResource.Command.Tests.ps1
```diff
@@ -14,221 +14,148 @@ Describe 'Test Find-PSResource for Command' {
...
-    # Purpose: not find Command resource given unavailable name
-    #
-    # Action: Find-PSResource -Name NonExistantCommand
-    #
-    # Expected result: should not return NonExistantCommand resource
-    It "should not find Command resource given unavailable name" {
-        $res = Find-PSResource -Name NonExistantCommand
-        $res | Should -BeNullOrEmpty
+    It "find Command resource given Name parameter" {
+        $res = Find-PSResource -Name "test_command_module"
```
SteveL-MSFT, line 22: I think test modules should use the namespace: Microsoft.PowerShell.Test.*

test/FindResource.Command.Tests.ps1
```diff
@@ -14,221 +14,148 @@ Describe 'Test Find-PSResource for Command' {
...
-        @{Version='[4-2-1-0]';       Description="version formatted with invalid delimiter"},
-        @{Version='[4.*.0]';         Description="version with wilcard in middle"},
-        @{Version='[*.2.1.0]';       Description="version with wilcard at start"},
-        @{Version='[4.*.1.0]';       Description="version with wildcard at second digit"},
-        @{Version='[4.2.*.0]';       Description="version with wildcard at third digit"}
-        @{Version='[4.2.1.*]';        Description="version with wildcard at end"},
-        @{Version='[4..1.0]';        Description="version with missing digit in middle"},
-        @{Version='[4.2.1.]';        Description="version with missing digit at end"},
-        @{Version='[4.2.1.0.0]';     Description="version with more than 4 digits"}
+        @{Version='(2.5.0.0)';       Description="exlcusive version (4.2.1.0)"},
```
SteveL-MSFT: Description doesn't match test?  There's also a typo

test/FindResource.Command.Tests.ps1
```diff
@@ -14,221 +14,148 @@ Describe 'Test Find-PSResource for Command' {
...
+        @{Version='[2.5.*.0]';       Description="version with wildcard at third digit"}
+        @{Version='[2.5.0.*]';       Description="version with wildcard at end"},
+        @{Version='[2..0.0]';        Description="version with missing digit in middle"},
+        @{Version='[2.5.0.]';        Description="version with missing digit at end"},
+        @{Version='[2.5.0.0.0]';     Description="version with more than 4 digits"}
     ) {
         param($Version, $Description)
 
         $res = $null
         try {
```
SteveL-MSFT: What is purpose of try/catch?  If you ever have a generic catch, it must have a comment indicating why.  I realize this is existing code, but we should fix this as it might be catching errors that should cause the test to fail.  Applies everywhere.

test/FindResource.Command.Tests.ps1
```diff
@@ -14,221 +14,148 @@ Describe 'Test Find-PSResource for Command' {
...
-    # Purpose: find Command resource given ModuleName to validate version match
-    #
-    # Action: Find-PSResource -ModuleName Az.Compute -Version "*"
-    #
-    # Expected Result: should return all Az.Compute resources (versions in descending order)
     It "find Command resource given ModuleName to validate version wildcard match " {
-        $res = Find-PSResource -ModuleName "Az.Compute" -Version "*"
-        $res.Count | Should -BeGreaterOrEqual 40
+        $res = Find-PSResource -ModuleName "test_command_module" -Version "*" -Repository $TestGalleryName
+        $res.Count | Should -Be 4
```
SteveL-MSFT: Should have a loop to validate each of the results is correct module, applies to all cases

test/FindResource.Module.Tests.ps1
```diff
@@ -15,75 +15,40 @@ Describe 'Test Find-PSResource for Module' {
...
     }
 
-    # Purpose: to find all resources when no parameters are specified
-    #
-    # Action: Find-PSResource
-    #
-    # Expected-Result: finds all (more than 1) resources in PSGallery
     It "find Resources Without Any Parameter Values" {
         $psGetItemInfo = Find-PSResource
         $psGetItemInfo.Count | Should -BeGreaterThan 1
```
SteveL-MSFT: Does this actually query for all the modules in the gallery?  Seems like an unnecessary load on psgallery.  Perhaps the cmdlet should require some input before returning results?  Should open an issue to discuss, even requiring an explicit wildcard like `*` would indicate it was a deliberate action by the user rather than just trying `Find-PSResource` and incurring a load on psgallery.

test/FindResource.Module.Tests.ps1
```diff
@@ -98,166 +63,111 @@ Describe 'Test Find-PSResource for Module' {
...
-    # Purpose: find resources when given ModuleName, Version not null --> '*'
-    #
-    # Action: Find-PSResource -ModuleName ContosoServer -Version "*" -Repository PoshTestGallery
-    #
-    # Expected Result: returns 4 ContosoServer resources (of all versions in descending order)
     It "find resources when given ModuleName, Version not null --> '*'" {
-        $res = Find-PSResource -ModuleName ContosoServer -Version "*" -Repository $TestGalleryName
-        $res.Count | Should -BeGreaterOrEqual 4
+        $res = Find-PSResource -ModuleName "test_module" -Version "*" -Repository $TestGalleryName
+        $res.Count | Should -Be 7
```
SteveL-MSFT: Instead of the magic number 7, should have somewhere (probably in the utils module) a hashtable that contains all the test module versions and then using a count to match

test/FindResource.RoleCapability.Tests.ps1
```diff
@@ -78,41 +53,21 @@ Describe 'Test Find-PSResource for Role Capability' {
...
 
-    # Purpose: find Role Capability resource with wildcard version, given Version parameter
-    #
-    # Action: Find-PSResource -Name "TestRoleCapModule" -Version "*"
-    #
-    # Expected Result: returns all versions of DSCTestModule (versions in descending order)
     It "find Role Capability resource with wildcard version, given Version parameter -> '*' " {
-        $res = Find-PSResource -Name "DscTestModule" -Version "*" -Repository $TestGalleryName
+        $res = Find-PSResource -Name "test_rolecap_module" -Version "*" -Repository $TestGalleryName
         $res.Count | Should -BeGreaterOrEqual 1
```
SteveL-MSFT: The count should be exact since we control the test modules

test/FindResource.Script.Tests.ps1
```diff
@@ -134,101 +82,64 @@ Describe "Test Find-PSResource for Script" {
...
     It "find resource with latest (including prerelease) version given Prerelease parameter" {
         # test_script resource's latest version is a prerelease version, before that it has a non-prerelease version
-        $res = Find-PSResource -Name "test_script"
+        $res = Find-PSResource -Name "test_script" -Repository $TestGalleryName
         $res.Version | Should -Be "2.5.0.0"
 
-        $resPrerelease = Find-PSResource -Name "test_script" -Prerelease
-        $resPrerelease.Version | Should -Be "3.0.0.0"        
+        $resPrerelease = Find-PSResource -Name "test_script" -Prerelease -Repository $TestGalleryName
+        $resPrerelease.Version | Should -Be "3.0.0.0"
```
SteveL-MSFT: There's trailing whitespace, you should have VSCode setup to automatically strip it

test/testRepositories.xml
```diff
@@ -1,6 +1,8 @@
 <?xml version="1.0" encoding="utf-8"?>
 <configuration>
-  <Repository Name="psgettestlocal"  Url="file:///c:/code/testdir"                  Priority="50" Trusted="false" />
+
```
SteveL-MSFT: unnecessary whitespace?

test/testRepositories.xml
```diff
@@ -1,6 +1,8 @@
...
+  <Repository Name="psgettestlocal"  Url="file:///c:/code/testdir"                  Priority="40" Trusted="false" />
```
SteveL-MSFT: these file paths only work on Windows and only if that machine has C: drive which isn't guaranteed

SteveL-MSFT (changes requested):

test/FindResource.Module.Tests.ps1
anamnavi: added in

test/FindResource.RoleCapability.Tests.ps1
anamnavi: I forgot to change this test to be against the exact count. Updated it now to check count against that of the role capability test module hashtable stored in PSGetTestUtils

test/FindResource.Module.Tests.ps1
anamnavi: Created issue, will discuss in PSGet triage meeting on Thursday.
https://github.com/PowerShell/PowerShellGet/issues/302

test/FindResource.Command.Tests.ps1
anamnavi: Changed this to catch the ArgumentException being thrown.

test/FindResource.Command.Tests.ps1
anamnavi: I discussed this with Amber and her view was that we may be better off changing this when we move to NuGet gallery or reach the situation when we need the names to have the Microsoft namespace to stand out and not conflict with generic test modules. The way to change the name from something like `test_module` to `Microsoft.PowerShell.Test.*` is to either create a new module with this name or to update it in the database/from the backend which Amber could do. It's not a simple update by publishing the module with an updated version (with the name change made in). So her suggestion was to update once when we know for sure what we want it to be like.

test/testRepositories.xml
anamnavi: Good point! My thought was that this filepath gets changed before it's ever used, but I see that it's better to not have to rely on that alone. Better to not introduce something that could cause an issue later or if misused. Additionally, this really helped my understanding of how PSGet registers the default repos and which ones. I removed this and instead register the test repo and unregister it in the PSGetTestUtils file now, with Join-Path to create the path.

test/FindResource.Command.Tests.ps1
```diff
@@ -8,227 +8,168 @@ Describe 'Test Find-PSResource for Command' {
     BeforeAll{
        $TestGalleryName = Get-PoshTestGalleryName
        $PSGalleryName = Get-PSGalleryName
+       $CommandTest = Get-CommandTestModule
        Get-NewPSResourceRepositoryFile
+       Get-RegisterLocalRepos
```
JamesWTruher:
```suggestion
       Register-LocalRepos
```

test/FindResource.Command.Tests.ps1
```diff
@@ -8,227 +8,168 @@ Describe 'Test Find-PSResource for Command' {
...
-    # Purpose: find Command resource given Name paramater
-    #
-    # Action: Find-PSResource -Name Az.Compute
-    #
-    # Expected Result: returns Az.Compute resource
-    It "find Command resource given Name parameter" {
-        $res = Find-PSResource -Name Az.Compute
-        $res | Should -Not -BeNullOrEmpty
-        $res.Name | Should -Be "Az.Compute"
+       Get-UnregisterLocalRepos
```
JamesWTruher:
```suggestion
       Unregister-LocalRepos
```
`Register` and `Unregister` are approved verbs

test/FindResource.Command.Tests.ps1
```diff
@@ -8,11 +8,14 @@ Describe 'Test Find-PSResource for Command' {
     BeforeAll{
        $TestGalleryName = Get-PoshTestGalleryName
        $PSGalleryName = Get-PSGalleryName
+       $CommandTest = Get-CommandTestModule
        Get-NewPSResourceRepositoryFile
+       Get-RegisterLocalRepos
```
SteveL-MSFT: `Register` is a verb.  Also, nouns are expected to be singular, so perhaps `Get-LocalPSRepository`

test/FindResource.Command.Tests.ps1
```diff
@@ -8,11 +8,14 @@ Describe 'Test Find-PSResource for Command' {
...
     }
 
     AfterAll {
        Get-RevertPSResourceRepositoryFile
+       Get-UnregisterLocalRepos
```
SteveL-MSFT: Same as above

test/FindResource.Command.Tests.ps1
```diff
@@ -50,19 +61,17 @@ Describe 'Test Find-PSResource for Command' {
...
         param($Version, $Description)
-
-        $res = $null
-        try {
-            $res = Find-PSResource -Name "test_command_module" -Version $Version -Repository $TestGalleryName -ErrorAction Ignore
-        }
-        catch {}
-
-        $res | Should -BeNullOrEmpty
+        {Find-PSResource -Name "test_command_module" -Version $Version -Repository $TestGalleryName -ErrorAction Ignore} | Should -Throw -ExceptionType ([ArgumentException])
```
SteveL-MSFT:
```suggestion
        { Find-PSResource -Name "test_command_module" -Version $Version -Repository $TestGalleryName -ErrorAction Ignore } | Should -Throw -ExceptionType ([ArgumentException])
```

test/FindResource.Command.Tests.ps1
```diff
@@ -50,19 +61,17 @@ Describe 'Test Find-PSResource for Command' {
...
     }
 
-    It "find Command resource given Name to validate version wilcard match " {
-        $res = Find-PSResource -Name "test_command_module" -Version "*" -Repository $TestGalleryName
-        $res.Count | Should -Be 4
+    It "find resources when given Name, Version not null --> '*'" {
+        $expectedModules = @()
```
SteveL-MSFT: Typically `expected` is what you validate against, so this is really `$actualModules`

test/FindResource.Command.Tests.ps1
```diff
@@ -50,19 +61,17 @@ Describe 'Test Find-PSResource for Command' {
...
+        Find-PSResource -Name "test_command_module" -Version "*" -Repository $TestGalleryName | ForEach-Object {
+            if($_.Name -eq "test_command_module") {
+                $expectedModules += $_.Name
+            }
```
SteveL-MSFT, line 72: Is there a reason to filter out incorrect results that may be returned?  Seems like this test should validate that each returned module has the right name

test/FindResource.DSCResource.Tests.ps1
```diff
@@ -51,19 +62,17 @@ Describe 'Test Find-PSResource for DSC Resource' {
...
         param($Version, $Description)
-
-        $res = $null
-        try {
-            $res = Find-PSResource -Name "test_dsc_module" -Version $Version -Repository $TestGalleryName -ErrorAction Ignore
-        }
-        catch {}
-
-        $res | Should -BeNullOrEmpty
+        {Find-PSResource -Name "test_dsc_module" -Version $Version -Repository $TestGalleryName -ErrorAction Ignore} | Should -Throw -ExceptionType ([ArgumentException])
```
SteveL-MSFT:
```suggestion
        { Find-PSResource -Name "test_dsc_module" -Version $Version -Repository $TestGalleryName -ErrorAction Ignore } | Should -Throw -ExceptionType ([ArgumentException])
```

test/FindResource.Command.Tests.ps1
SteveL-MSFT: Why is `-ErrorAction Ignore` used here?

test/FindResource.Command.Tests.ps1
```diff
@@ -38,8 +41,16 @@ Describe 'Test Find-PSResource for Command' {
...
 
     It "not find resource with incorrectly formatted version such as <Description>" -TestCases @(
-        @{Version='(2.5.0.0)';       Description="exlcusive version (4.2.1.0)"},
-        @{Version='[2-5-0-0]';       Description="version formatted with invalid delimiter"},
+        @{Version='(2.5.0.0)';       Description="exclusive version (2.5.0.0)"},
+        @{Version='[2-5-0-0]';       Description="version formatted with invalid delimiter"}
+    ) {
+        param($Version, $Description)
+
+        $res = Find-PSResource -Name "test_command_module" -Version $Version -Repository $TestGalleryName -ErrorAction Ignore
```
SteveL-MSFT, line 49: I thought we were going to use namespaced test module names?

test/FindResource.Module.Tests.ps1
```diff
@@ -93,8 +97,13 @@ Describe 'Test Find-PSResource for Module' {
     }
 
     It "find resources when given ModuleName, Version not null --> '*'" {
-        $res = Find-PSResource -ModuleName "test_module" -Version "*" -Repository $TestGalleryName
-        $res.Count | Should -Be 7
+        $expectedModules = @()
+        Find-PSResource -ModuleName "test_module" -Version "*" -Repository $TestGalleryName | ForEach-Object {
```
SteveL-MSFT, line 101: This code is similar in several cases, you can use `-TestCases` with splatting to cover different parameter variations and use the same validation code.  Something like:

```powershell
It "... with parameter '<Parameter>'" -TestCases @(
  @{ Parameter = 'Name' }
  @{ Parameter = 'ModuleName' }
) {
  param($parameter)
  ...
  $parameters = @{
    $parameter = 'test_module'
    Version = '*'
    Repository = $TestGalleryName
  }

  Find-PSResource @parameters
```

test/PSGetTestUtils.psm1
```diff
@@ -30,6 +30,50 @@ $script:PSGalleryLocation = 'https://www.powershellgallery.com/api/v2'
 $script:PoshTestGalleryName = 'PoshTestGallery'
 $script:PostTestGalleryLocation = 'https://www.poshtestgallery.com/api/v2'
 
+$script:command_test_module = @(
+    @{Name="test_command_module"; Version="2.5.0.0"; Repository="PoshTestGallery"; Description="this a test command resource of type module"},
```
SteveL-MSFT: lots of repetition, so recommend storing repeated values in a variable in case they need to change in the future or reused

SteveL-MSFT (changes requested):

test/PSGetTestUtils.psm1
anamnavi: stored in variable.

PR #321, opened 2020-12-10
WIP: Catalog v3 implementation

anamnavi:
This PR is a work in progress. For V3 protocol, currently a user cannot search for a package that has a wildcard in the name and the repository has a V3 endpoint.

My goal with this PR is that with Find-PSResource, user should be able to:
- [x] search for a package with wildcard(s) in the Name
- [x] support prerelease versions when searching for the wildcard package (using -Prerelease switch)
- [x] write the found package's info in a way consistent with existing search (i.e should return name, version, description, repository)
- [x] display progress bar when filtering through the packages, with SecondsRemaining and PercentComplete metrics

I would appreciate feedback on **which technique of searching and filtering packages should be kept**. Technique 1 uses an Dictionary to store unique packages with the latest version. Technique 2 uses PLINQ to sort through the packages in parallel and obtain a list of unique packages with the latest versions.

Technique 2 takes less time for each wildcard permutation I've tried (i.e for input "Far*et, technique 2 takes 01.34 (secs.ms) whereas technique 1 takes 03.51 (secs.ms). My only concern with technique 2 is that when doing WriteProgress and updating the count for PercentComplete and SecondsRemanining I'm not confident that the code is thread-safe. I've left a comment explaining this near the code [here](https://github.com/PowerShell/PowerShellGet/pull/321/files#diff-6d6ebe8b16cd1e4d4887171b9802156242aef292ecb8ff3c9b6de8bfe429ab63R373).

For a full report of runtimes between both techniques and wildcard cases covered see [here](https://docs.google.com/document/d/1wbquguQZJ0Il0EFxrnU_G0rm-DxWxR2XN9lKAGTsdhw/edit?usp=sharing)


src/code/FindPSResource.cs
```diff
@@ -206,62 +228,181 @@ protected override void ProcessRecord()
...
+                int totalPkgsCount = allPkgs.Count();
+                int currentPkg = 0;
+
+                DateTime start = DateTime.Now;
+                int trackedPercent = 0;
+                uniquePkgsWithLatestVersion = allPkgs.AsParallel()
+                    .Select(x => {
+                        TimeSpan elapsedTime = DateTime.Now - start;
+                        int currentPercent =(int)(100.0 * Interlocked.Increment(ref currentPkg)/totalPkgsCount);
+                        // I'm not 100% confident in this code being thread-safe. variables currentPkg and trackedPercent are the ones
```
anamnavi, line 369: Please see the note I've left here

src/code/FindPSResource.cs
```diff
@@ -57,6 +57,24 @@ public string[] Name
         }
         private string[] _name = new string[0];
 
+        /// <summary>
+        /// corresponds to which wildcard implementation technique we want to use/test out
+        /// added this for testing wildcard feature thru CLI
+        /// </summary>
+        [Parameter(ValueFromPipeline = true, ValueFromPipelineByPropertyName = true)]
```
SteveL-MSFT, line 64: Is this temporary and only for initial testing?  Should use an enum instead of Int.  With PowerShell 7, you can mark this as experimental so that users understand that it may change (https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_experimental_features?view=powershell-7.1)

src/code/FindPSResource.cs
```diff
@@ -206,62 +224,181 @@ protected override void ProcessRecord()
             foreach (var repoName in listOfRepositories)
             {
                 WriteDebug(string.Format("Searching in repository '{0}'", repoName.Properties["Name"].Value.ToString()));
-                // We'll need to use the catalog reader when enumerating through all packages under v3 protocol.
-                // if using v3 endpoint and there's no exact name specified (ie no name specified or a name with a wildcard is specified)
-                if (repoName.Properties["Url"].Value.ToString().EndsWith("/v3/index.json") && 
-                    (_name.Length == 0 || _name.Any(n => n.Contains("*"))))//_name.Contains("*")))  /// TEST THIS!!!!!!!
+
+                if (repoName.Properties["Url"].Value.ToString().EndsWith("/v3/index.json") &&
```
SteveL-MSFT, line 228: Is case sensitively a concern for the URL?

src/code/FindPSResource.cs
```diff
@@ -206,62 +224,181 @@ protected override void ProcessRecord()
...
+                    else if(_technique == 2)
+                    {
+                        pkgsWithDescriptions = ProcessCatalogWithPLINQ(_name[0], repoName.Properties["Url"].Value.ToString(), cancellationToken);
+                    }
+
+                    if(pkgsWithDescriptions.Any())
+                    {
+                        foreach(CatalogEntry pkg in pkgsWithDescriptions.Keys)
+                        {
+                            PSObject pkgAsObject = new PSObject();
```
SteveL-MSFT, line 245: We should probably declare a class for this rather than returning a PSObject with noteproperties

src/code/FindPSResource.cs
```diff
@@ -206,62 +224,181 @@ protected override void ProcessRecord()
...
+                    // returns any pkgs found, and any pkgs that weren't found
+                    returnedPkgsFound.AddRange(FindPackagesFromSource(repoName.Properties["Name"].Value.ToString(), repoName.Properties["Url"].Value.ToString(), cancellationToken));
 
-                // if it can't find the pkg in one repository, it'll look in the next one in the list
-                // returns any pkgs found, and any pkgs that weren't found
-                returnedPkgsFound.AddRange(FindPackagesFromSource(repoName.Properties["Name"].Value.ToString(), repoName.Properties["Url"].Value.ToString(), cancellationToken));
 
+                    // Flatten returned pkgs before displaying output returnedPkgsFound.Flatten().ToList()[0]
+                    var flattenedPkgs = returnedPkgsFound.Flatten();
+                    // flattenedPkgs.ToList();
```
SteveL-MSFT, line 269: Commented out code should be removed

src/code/FindPSResource.cs
```diff
@@ -206,62 +224,181 @@ protected override void ProcessRecord()
...
+                    // if we need to search all repositories, we'll continue, otherwise we'll just return
+                    if (!pkgsLeftToFind.Any())
+                    {
+                        break;
+                    }
                 }
+            }
+        }
+
+        public Dictionary<CatalogEntry, string> ProcessCatalogWithDict(string repoName, string sourceUrl, CancellationToken cancellationToken){
```
SteveL-MSFT, line 302: STYLE: opening braces should always be on a newline

src/code/FindPSResource.cs
```diff
@@ -206,62 +224,181 @@ protected override void ProcessRecord()
...
-            } // end of foreach
+                int trackedPercent = 0;
+                foreach(CatalogEntry pkg in my_list)
+                {
+                    TimeSpan elapsedTime = DateTime.Now - start;
+                    int currentPercent = (int)((100.0 * (currentPkg++))/totalPkgsCount);
+                    if(trackedPercent < currentPercent)
+                    {
+                        trackedPercent = currentPercent;
+                        Host.UI.WriteProgress(
```
SteveL-MSFT, line 326: Is there a reason you are calling Host.UI.WriteProgress() directly instead of using WriteProgress() that is part of Cmdlet class?

src/code/FindPSResource.cs
```diff
@@ -57,6 +57,24 @@ public string[] Name
...
         private string[] _name = new string[0];
 
+        /// <summary>
+        /// corresponds to which wildcard implementation technique we want to use/test out
+        /// added this for testing wildcard feature thru CLI
+        /// </summary>
+        [Parameter(ValueFromPipeline = true, ValueFromPipelineByPropertyName = true)]
+        [ValidateNotNull]
+
+        public int Technique
```
PaulHigin, line 67: Maybe 'Technique' should be an enum type with allow list.  But I assume this is just for testing so may not be worth changing.

SteveL-MSFT (changes requested):
Perhaps have tests written first before code to show how this feature would work with the expected results?

src/code/FindPSResource.cs
```diff
@@ -206,62 +224,181 @@ protected override void ProcessRecord()
             foreach (var repoName in listOfRepositories)
             {
                 WriteDebug(string.Format("Searching in repository '{0}'", repoName.Properties["Name"].Value.ToString()));
-                // We'll need to use the catalog reader when enumerating through all packages under v3 protocol.
-                // if using v3 endpoint and there's no exact name specified (ie no name specified or a name with a wildcard is specified)
-                if (repoName.Properties["Url"].Value.ToString().EndsWith("/v3/index.json") && 
-                    (_name.Length == 0 || _name.Any(n => n.Contains("*"))))//_name.Contains("*")))  /// TEST THIS!!!!!!!
+
+                if (repoName.Properties["Url"].Value.ToString().EndsWith("/v3/index.json") &&
+                    (_name.Length != 0 || _name.Any(n => n.Contains("*"))))
```
PaulHigin, line 229: Is '*' the only wildcard character supported by v3 protocol?  If so, we might want to see how we can support PowerShell wildcard characters with an extra layer.

src/code/FindPSResource.cs
```diff
@@ -206,62 +224,181 @@ protected override void ProcessRecord()
...
+                    // if it can't find the pkg in one repository, it'll look in the next one in the list
+                    // returns any pkgs found, and any pkgs that weren't found
+                    returnedPkgsFound.AddRange(FindPackagesFromSource(repoName.Properties["Name"].Value.ToString(), repoName.Properties["Url"].Value.ToString(), cancellationToken));
 
-                // if it can't find the pkg in one repository, it'll look in the next one in the list
-                // returns any pkgs found, and any pkgs that weren't found
-                returnedPkgsFound.AddRange(FindPackagesFromSource(repoName.Properties["Name"].Value.ToString(), repoName.Properties["Url"].Value.ToString(), cancellationToken));
 
+                    // Flatten returned pkgs before displaying output returnedPkgsFound.Flatten().ToList()[0]
+                    var flattenedPkgs = returnedPkgsFound.Flatten();
```
PaulHigin, line 268: Why do packages need to be flattened?  Shouldn't 'FindPackagesFromSource' return a simple list?  Do we ever want hierarchical lists?  If not then it seems helper functions should return data with desired format.

src/code/FindPSResource.cs
```diff
@@ -206,62 +224,181 @@ protected override void ProcessRecord()
...
 
 
-                        PSObject pkgAsPSObject = new PSObject();
-                        pkgAsPSObject.Members.Add(new PSNoteProperty("Name", pkg.Identity.Id));
-                        // Version.Version ensures type is System.Version instead of type NuGetVersion
-                        pkgAsPSObject.Members.Add(new PSNoteProperty("Version", pkg.Identity.Version.Version));
-                        pkgAsPSObject.Members.Add(new PSNoteProperty("Repository", repoName.Properties["Name"].Value.ToString()));
-                        pkgAsPSObject.Members.Add(new PSNoteProperty("Description", pkg.Description));
+                            PSObject pkgAsPSObject = new PSObject();
+                            pkgAsPSObject.Members.Add(new PSNoteProperty("Name", pkg.Identity.Id));
```
PaulHigin, line 279: I feel we should have a regular .NET type for returning data (like PSRepositoryItemInfo), and not create a PSCustomObject.

src/code/FindPSResource.cs
PaulHigin, line 302: Should this be a 'private' accessible method?

Actually, I think both these functions should be part of a utility or helper class.

src/code/FindPSResource.cs
```diff
@@ -206,62 +224,181 @@ protected override void ProcessRecord()
...
+                    {
+                        break;
+                    }
                 }
+            }
+        }
+
+        public Dictionary<CatalogEntry, string> ProcessCatalogWithDict(string repoName, string sourceUrl, CancellationToken cancellationToken){
+            var feed = new Uri(sourceUrl);
+            Dictionary<string, CatalogEntry> uniquePkgs = new Dictionary<string, CatalogEntry>();
```
PaulHigin, line 304: Nitpick.  You can use 'var' in these delcarations:
```csharp
var uniquePkgs = new Dictionary<string, CatalogEntry>();
```

src/code/FindPSResource.cs
```diff
@@ -206,62 +224,181 @@ protected override void ProcessRecord()
...
+                    if (!pkgsLeftToFind.Any())
+                    {
+                        break;
+                    }
                 }
+            }
+        }
+
+        public Dictionary<CatalogEntry, string> ProcessCatalogWithDict(string repoName, string sourceUrl, CancellationToken cancellationToken){
+            var feed = new Uri(sourceUrl);
```
PaulHigin, line 303: Style: No need to declare this variable here, and can insert in 'new CatalogReader(feed)' below, since it seems to be used only there.

src/code/FindPSResource.cs
```diff
@@ -206,62 +224,181 @@ protected override void ProcessRecord()
...
+            Dictionary<string, CatalogEntry> uniquePkgs = new Dictionary<string, CatalogEntry>();
+            Dictionary<CatalogEntry, string> pkgsWithDesc = new Dictionary<CatalogEntry, string>();
+
+            using(var catalog = new CatalogReader(feed))
+            {
+                IReadOnlyList<CatalogEntry> my_list = catalog.GetFlattenedEntriesAsync(cancellationToken).GetAwaiter().GetResult();
```
PaulHigin, line 309: Style: `my_list` should benamed `cataloglist`?  To be more descriptive.

src/code/FindPSResource.cs
```diff
@@ -206,62 +224,181 @@ protected override void ProcessRecord()
...
 
-            } // end of foreach
+                int trackedPercent = 0;
+                foreach(CatalogEntry pkg in my_list)
+                {
+                    TimeSpan elapsedTime = DateTime.Now - start;
+                    int currentPercent = (int)((100.0 * (currentPkg++))/totalPkgsCount);
+                    if(trackedPercent < currentPercent)
+                    {
+                        trackedPercent = currentPercent;
```
PaulHigin, line 325: Can this progress logic be wrapped into a helper method?  It may be needed for other cmdlets or cases.

src/code/FindPSResource.cs
```diff
@@ -206,62 +224,181 @@ protected override void ProcessRecord()
...
+                        Host.UI.WriteProgress(
+                            sourceId:1,
+                            new ProgressRecord(activityId:0, "Filtering packages...",  $"Filtering through {currentPkg} package")
+                            {
+                                PercentComplete = trackedPercent,
+                                SecondsRemaining = (int) ((elapsedTime.TotalMilliseconds/currentPkg) * (totalPkgsCount - currentPkg)),
+                            }
+                        );
+                    }
+                    if(v3Pattern.IsMatch(pkg.Id) && (_prerelease || !pkg.Version.IsPrerelease) && ((!uniquePkgs.TryGetValue(pkg.Id, out CatalogEntry entry)) || (pkg.Version > entry.Version)))
```
PaulHigin, line 335: We usually add a comment to help clarify the filter logic.  Something like:
```csharp
// Add unique package if
//    Pattern matches
//    Not prerelease version or prerelease versions allowed
//    Package version does not already exist in list
```

src/code/FindPSResource.cs
```diff
@@ -206,62 +224,181 @@ protected override void ProcessRecord()
...
+                    {
```
PaulHigin, line 336: The last filter condition `(pkg.Version > entry.Version)` doesn't seem right.  Presumably there can be multiple packages in `uniquePkgs` with different versions, and an new one should only be added if the version is different from any other package version (with same id).  This seems to rely on order, which Dictionary does not guarantee.

I wonder if the uniquePkgs dictionary should be keyed with a key generated from id and version.

src/code/FindPSResource.cs
```diff
@@ -206,62 +224,181 @@ protected override void ProcessRecord()
...
+                        uniquePkgs[pkg.Id] = pkg;
+                    }
+                }
+                foreach(string pkgName in uniquePkgs.Keys)
+                {
+                    CatalogEntry current = uniquePkgs[pkgName];
+                    pkgsWithDesc.Add(current, current.GetNuspecAsync().GetAwaiter().GetResult().GetDescription());
```
PaulHigin, line 343: I assume this is done in a separate loop because this retrieve has performance implications?  A comment should be added to clarify.

src/code/FindPSResource.cs
```diff
@@ -206,62 +224,181 @@ protected override void ProcessRecord()
...
+                DateTime start = DateTime.Now;
+                int trackedPercent = 0;
+                uniquePkgsWithLatestVersion = allPkgs.AsParallel()
+                    .Select(x => {
+                        TimeSpan elapsedTime = DateTime.Now - start;
+                        int currentPercent =(int)(100.0 * Interlocked.Increment(ref currentPkg)/totalPkgsCount);
+                        // I'm not 100% confident in this code being thread-safe. variables currentPkg and trackedPercent are the ones
+                        // we're locking on to ensure different threads can't modify at same time.
+                        // Could a thread update currentPkg on L 372 but before reaching L380 another thread hits line 372 and increments currentPkg?
+                        // A possible solution could be to use the commented conditional code below that nests Interlocked operations
```
PaulHigin, line 372: I agree that this is worrisome.  It is always hard to reason about multi-threaded code.  I am pretty sure `Host.UI.WriteProgress()` is not thread safe, plus it probably must only be called with monotonically increasing values.

I think the best solution is to have a worker thread that writes to `Host.UI.WriteProgress()`, and only writes when a volatile 'currentprecent' variable is incremented.

However, I am not sure it is worth the effort just for progress tracking.

src/code/FindPSResource.cs
```diff
@@ -206,62 +224,181 @@ protected override void ProcessRecord()
...
+                foreach(string pkgName in uniquePkgs.Keys)
+                {
+                    CatalogEntry current = uniquePkgs[pkgName];
+                    pkgsWithDesc.Add(current, current.GetNuspecAsync().GetAwaiter().GetResult().GetDescription());
+                }
+            }
+            return pkgsWithDesc;
+        }
+
+        public Dictionary<CatalogEntry, string> ProcessCatalogWithPLINQ(string repoName, string sourceUrl, CancellationToken cancellationToken)
```
PaulHigin, line 349: It is not clear to me that the perf gain is worth the trouble of multi-threaded code here.  But if we find it is, I would rather use .Net Tasks rather than PLINQ.

PaulHigin (changes requested):
Overall looks good.  Let's chat about some of these comments.

src/code/FindPSResource.cs
rjmholt: I know it's been a while, but I just want to chime in to say I suggested the code here on the basis that `Host.UI.WriteProgress()` is threadsafe ([this](https://github.com/PowerShell/PowerShell/blob/9aede0558a4c26a4f0769a51d8f0bb3e5fcfa744/src/Microsoft.PowerShell.ConsoleHost/host/msh/ConsoleHostUserInterface.cs#L1322) suggests at least that ConsoleHost's UI is written with this intention).

Given that @anamnavi discovered that parallelising the catalog presented a significant speedup, I tried to come up with a relatively lightweight way to preserve that parallelisation while also informing the user of progress. So rather than juggling locks or event queues, this just increments an integer across threads and whichever thread pushes the percentage over a given interval updates the progress.

The downsides are:
- It's not terribly straightforward code, although I'm not sure another solution would be more obvious
- The `sourceId` parameter is cumbersome
- The thread-safety relies on a monotonicity condition within the logic itself, rather than a locking primitive, so changing it could break things
- There is indeed a race-condition when the global percentage is updated, although because it's just about deciding to write progress it's fairly harmless

I'll see if I can come up with a consumer-thread solution for the sake of comparison

rjmholt: Here's an alternative, event-queue-based write-progress demo:

```csharp
protected override void EndProcessing()
{
    // We need to be able to cancel in case something happens in the tasks
    var tasksErrored = new CancellationTokenSource();
    // We'll use this as an event queue to write progress
    var progressEvents = new BlockingCollection<bool>();
    // This is our global monotonically increasing count of task progress across threads
    int count = 0;

    // Launch our worker tasks
    Task workersDone = Task.WhenAll(
        Enumerable.Range(0, 100000)
            .Select(async i => {
                try
                {
                    // Simulate some work here
                    await Task.Delay(1);
                    // There are 100,000 tasks, and every 1000th will send a progress update, for 100 total
                    if (Interlocked.Increment(ref count) % 1000 == 0)
                    {
                        progressEvents.Add(true);
                    }
                }
                catch
                {
                    // If anything happens, we cancel the awaiter
                    // but also rethrow so the task awaiter will pick up the exception
                    tasksErrored.Cancel();
                    throw;
                }
            // We also want to signal when we're done so the awaiter knows to end
            // we could do this with the cancellation token
            // but that would create a race condition between the workers and the awaiter
            // so instead we just send a different message to the event queue
            })).ContinueWith(_ => progressEvents.Add(false));

    // Now we synchronously run an awaiter loop on the pipeline thread
    // Every time we receive an event, we update the progress
    // The challenge here is ending properly so we don't throw or get stuck waiting for events that will never come
    int progressPercent = 0;
    try
    {
        foreach (bool shouldContinue in progressEvents.GetConsumingEnumerable(tasksErrored.Token))
        {
            // We will be sent false at the end, so that's when we break
            if (!shouldContinue)
            {
                break;
            }

            // For every event, we just update the percentage by 1 and write progress
            WriteProgress(
                new ProgressRecord(activityId: 1, "Counting", "Almost done!")
                {
                    PercentComplete = progressPercent++
                });
        }
    }
    // This is how cancellation will occur, so we just catch it to proceed
    catch (OperationCanceledException)
    {
    }

    // We await the task here so that any exceptions that occur are thrown properly on our thread
    workersDone.GetAwaiter().GetResult();
}
 ```

The problems I see with it are:

- It's heavier weight than the counter solution
- It's more code with more moving parts and is conceptually complex when it comes to dealing with the edge cases and error modes
- The semantics of an unawaited task in .NET aren't great, but we can't both listen for events and await the task, so we're forced to do the first and then the second and trust that things will go as we hoped.

I can imagine my demo could be improved a fair amount, although at this point I'd also guess that a simple `lock` around some counter state in the original code would also work quite well, provide similar performance and be more readable.

Comment (PaulHigin) [reply src/code/FindPSResource.cs:372]: @rjmholt If you look at the remoting code, you will see that a lot of effort is expended ensuring all data streams (including progress) are marshaled onto a single thread on the client.  So I would be hesitant to trust WriteProgress.  Frankly, I feel the effort is probably not worth it, at least for the initial implementation.

Comment (rjmholt) [reply src/code/FindPSResource.cs:372]: Example using a lock (really a better configuration of the original):

```csharp
protected override void EndProcessing()
{
    // Lock for incrementing percentage and writing progress
    var progressLock = new object();
    // This is our global monotonically increasing count of task progress across threads
    int count = 0;
    // How much progress we've made
    int percentage = 0;

    Enumerable.Range(0, 1000)
        .AsParallel()
        .ForAll(i => {
            // Simulate work
            Task.Delay(1).Wait();

            // Each thread can atomically increment the count and get a unique result
            // If this thread is one of the 100 that ticks over a 0, it writes progress
            if (Interlocked.Increment(ref count) % 10 == 0)
            {
                // Now we lock around both writing progress and incrementing percentage
                lock (progressLock)
                {
                    Host.UI.WriteProgress(
                        sourceId: 1,
                        new ProgressRecord(activityId: 1, "Counting", "Almost done!")
                        {
                            PercentComplete = percentage++
                        });
                }

                // If we decided that Host.UI.WriteProgress is threadsafe, we could do this equivalently and save needing to use a lock:
                // Host.UI.WriteProgress(
                //     sourceId: 1,
                //     new ProgressRecord(activityId: 1, "Counting", "Almost done!")
                //     {
                //         PercentComplete = Interlocked.Increment(ref percentage),
                //     });
            }
        });
}
```

rjmholt: One thing I noticed doing this is that if you have any kind of async operation in your task body, you're much better off with a `Select(async _ => { })` rather than `.AsParallel().ForAll()` or similar, just because being able to `await` means you won't block the threadpool

PaulHigin: Again, I wonder if it is worth the effort/overhead to parallelize this.  The lock re-serializes the threads, making them less efficient.  Testing will show any benefit.  Also, the new x-plat, inline progress indicator may be more performant.

PR #327, opened 2021-01-13
Refactor all tests to Pester 5 support

JustinGrote:
This PR refactors all tests to work with Pester 5, enabling tests to be run on a per-test basis, especially with the Pester Test explorer in vscode by @TylerLeonhardt 

image

I focused primarily on converting the tests and not "fixing" them where broken, for tests that are clearly or unintentionally broken they were reflagged as pending with a Because reason. For example:
image

# Test Results
### WinPS 5.1
image
### Powershell 7.1
image

# Major Caveat
I did not update the DSC tests to 5, so they must be excluded when running pester, and the build process would need to be updated for that. I recommend installing both pester 4 and pester 5, and running the pester 4 DSC tests as a Start-Job. This PR also converts that repository to a git submodule because there's no need to maintain the code separately in this codebase since it isn't being forked.


alerickson:
@JustinGrote unfortunately we can't move onto Pester 5 just yet because PSPackageProject does not support it.  
I opened up an issue for us to track and get back to: https://github.com/TravisEz13/PSPackageProject/issues/51 
@PaulHigin looked into this recently: https://github.com/PowerShell/PowerShellGet/pull/358

I'm going to close this PR for now but we can circle back to this.

PR #329, opened 2021-01-15
Add design doc and PR template

PaulHigin:
Proposal for a design document to help define and implement cmdlets, and write tests.

### DIFF
--- a/Docs/DesignDoc.md
+++ b/Docs/DesignDoc.md
@@ -1,0 +1,307 @@
+# PowerShellGet V3 Module Design
+
+## Description
+
+PowerShellGet V3 is an upgrade to the currently available V2 module.
+The V2 module is completely script based, and has dependencies on other PowerShell modules(PackageManagement).  
+
+This version directly uses NuGet APIs, via NuGet managed code binaries.  
+
+For more information, see [Re-architecting PowerShellGet - the PowerShell package manager]().  
+
+## Goals
+
+- Works side by side with current PowerShellGet V2 module
+
+- Remove dependency on PackageManagement module, and directly use NuGet APIs
+
+- Leverage the latest NuGet V3 APIs
+
+- Provide cmdlets that perform similar functions but do not interfere with V2 cmdlets
+
+- Implement as binary cmdlets and minimize use of PowerShell scripts
+
+- Remove unneeded components (DscResources)
+
+- Minimize binary dependencies
+
+- Work over all PowerShell supported platforms
+
+- Minimize code duplication
+
+- Have only one .NET dependency (net461?) for Windows 5.x compatibility
+
+## Compatibility Module
+
+### Update module as needed
+
+### Write/update tests as needed
+
+## Summary of work estimates
+
+### Cmdlet work estimates
+
+TODO:
+
+### Compatibility Module work estimates
+
+TODO:
+
+## Cmdlets
+
+### Find-PSResource
+
+The `Find-PSResource` cmdlet combines the `Find-Module, Find-Script` cmdlets from V2.
+It performs a search on a repository (server) based on the `-Name` parameter argument.
+It returns `PSRepositoryItemInfo` objects which describe each resource item found.
+Other parameters allow the returned results to be filtered by item type and tags.
+
+#### Syntax
+
+``` PowerShell
+Find-PSResource [[-Name] <string[]>] [-Type <string[]>] [-Version <string>] [-Prerelease] [-Tags <string[]>]
+[-Repository <string[]>] [-Credential <pscredential>] [-IncludeDependencies] [-WhatIf] [-Confirm] [<CommonParameters>]
+```
+
+#### Parameters
+
+##### -Name
+
+Name of a resource or resources to find.
+Accepts wild card characters.
+
+```yml
+Type: string[]
+```
+
+#### -Type
+
+Specifies one or more resource types to find.
+Two resource types are supported: Module and Script.
+
+```yml
+Type: string[]
+AllowedValues: 'Module','Script'
+```
+
+#### -Version
+
+Specifies the version of the resource to be returned.
+
+```yml
+Type: string
+```
+
+#### -Prerelease
+
+When specified, includes prerelease versions in search.
+
+```yml
+Type: SwitchParameter
+```
+
+#### -Tags
+
+Filters search results for resources that include one or more of the specified tags.
+
+```yml
+Type: string[]
+```
+
+#### -Repository
+
+Specifies one or more repository names to search.
+If not specified, search will include all currently registered repositories.
+
+```yml
+Type: string[]
+```
+
+#### -Credential
+
+Optional credentials to be used when accessing a repository.
+
+```yml
+Type: PSCredential
+```
+
+#### -IncludeDependencies
+
+When specified, search will return all matched resources along with any resources the matched resources depends on.
+
+```yml
+Type: SwitchParameter
+```
+
+#### Outputs
+
+```json
+"PSRepositoryItemInfo" : {
+    "Name",
+    "Version",
+    "Type",
+    "Description",
+    "Author",
+    "CompanyName",
+    "Copyright",
+    "PublishedDate",
+    "InstalledDate",
+    "UpdatedDate",
+    "LicenseUri",
+    "ProjectUri",
+    "IconUri",
+    "Tags",
+    "Includes",
+    "PowerShellGetFormatVersion",
+    "ReleaseNotes",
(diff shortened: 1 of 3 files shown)


Docs/DesignDoc.md
```diff
@@ -0,0 +1,307 @@
+# PowerShellGet V3 Module Design
+
+## Description
+
+PowerShellGet V3 is an upgrade to the currently available V2 module.
+The V2 module is completely script based, and has dependencies on other PowerShell modules(PackageManagement).  
+
+This version directly uses NuGet APIs, via NuGet managed code binaries.  
+
+For more information, see [Re-architecting PowerShellGet - the PowerShell package manager]().
```
alerickson: RFC: https://github.com/PowerShell/PowerShell-RFC/pull/185

Docs/DesignDoc.md
```diff
@@ -0,0 +1,307 @@
...
+
+- Remove unneeded components (DscResources)
+
+- Minimize binary dependencies
+
+- Work over all PowerShell supported platforms
+
+- Minimize code duplication
+
+- Have only one .NET dependency (net461?) for Windows 5.x compatibility
```
alerickson, line 32: It'll just be netstandard2.0 now

alerickson: It'll just be netstandard2.0, compatible with net461

Docs/DesignDoc.md
```diff
@@ -0,0 +1,307 @@
...
+
+### Compatibility Module work estimates
+
+TODO:
+
+## Cmdlets
+
+### Find-PSResource
+
+The `Find-PSResource` cmdlet combines the `Find-Module, Find-Script` cmdlets from V2.
```
anamnavi: Would we need to make a note here about Find-PSResource also combining Find-Command and Find-RoleCapability? (as you said earlier DSC resources will be made obsolete, so perhaps no need to mention Find-DscResource)

Or perhaps saying that Find-PSResource is able to find resources that previously would have been categorized as type Command or RoleCapability, or would have been discovered via Find-Command and Find-RoleCapability are all able to be discovered via the new Find-PSResource cmdlet so that won't be a breaking change.

Docs/DesignDoc.md
```diff
@@ -0,0 +1,307 @@
...
+It returns `PSRepositoryItemInfo` objects which describe each resource item found.
+Other parameters allow the returned results to be filtered by item type and tags.
+
+#### Syntax
+
+``` PowerShell
+Find-PSResource [[-Name] <string[]>] [-Type <string[]>] [-Version <string>] [-Prerelease] [-Tags <string[]>]
+[-Repository <string[]>] [-Credential <pscredential>] [-IncludeDependencies] [-WhatIf] [-Confirm] [<CommonParameters>]
+```
+
```
anamnavi: Want to add a note that we will need to add one syntax set for when Find-PSResource is run with a wildcard name (-Name includes *) because not all the parameter switches will be able to be used. I'm not sure if there's a way to conditionally specify "when Name has *"

for example, `[[-Name if contains wildcard]<string[]>] [-Prerelease][-Repository]`

Docs/DesignDoc.md
```diff
@@ -0,0 +1,307 @@
...
+I believe V2 APIs do not support wildcard name searches, and that V3 APIs provide limited support.
+But PowerShell has its own wildcard behavior that may not be compatible with the NuGet APIs.
+PowerShellGet V2 appears to simply retrieve all resources from the server, and then filter as necessary, when a wildcard or filter parameter is used.
+This results in very slow operation as PowerShellGallery has more than 6000 resource items.
+Search strategy will initially follow V2s behavior.
+But some time needs to be spent to see if V3 wildcard support can be used to do more filtering on the server.  
+
+Search can be performed on remote or local (file based) repositories, depending on the repository Url.  
+
+Does `-Version` string parameter handle wild card characters?
```
anamnavi: -Version does not support wildcard characters.

Docs/DesignDoc.md
```diff
@@ -0,0 +1,307 @@
...
+We should assume that a specific resource package version is identical over any repository from which it is retrieved.  
+
+Should a search with no name or 'all' wildcard result in all repository items returned?
+Should there be a default limit for number of items returned, and if so can it be bypassed?  
+
+#### Tests
+
+Most search tests can be performed on a local repository.  
+
+Some tests should be performed on remote repository (PSGallery) to verify remote operation, but can be limited.
```
anamnavi: currently, those tests would be performed to PoshTestGallery, if verifying remote operation.

Docs/DesignDoc.md
```diff
@@ -0,0 +1,307 @@
...
+##### -Type param
+
+- Validate correct resource type returned
+
+##### -Version param
+
+- Validate correct resource version returned
+- Validate wild card (if supported), correct version range returned
+- Errors: Not found
+- Errors: Invalid version string format
```
anamnavi: Version has to be of syntax: "x.x.x.x" or "[x.x.x.x]" or "(x.x.x.x]" or "(x.x.x.x)" (not sure if this is obvious or worth specifying)

anamnavi (approved):

Docs/DesignDoc.md
```diff
@@ -0,0 +1,307 @@
...
+
+- Remove dependency on PackageManagement module, and directly use NuGet APIs
+
+- Leverage the latest NuGet V3 APIs
+
+- Provide cmdlets that perform similar functions but do not interfere with V2 cmdlets
+
+- Implement as binary cmdlets and minimize use of PowerShell scripts
+
+- Remove unneeded components (DscResources)
```
alerickson: We're still keeping DSCResources in the module, though unsure of whether this will be mof based or not (or possibly allowing for different types of DSCResources)

Docs/DesignDoc.md
```diff
@@ -0,0 +1,307 @@
...
+### Compatibility Module work estimates
+
+TODO:
+
+## Cmdlets
+
+### Find-PSResource
+
+The `Find-PSResource` cmdlet combines the `Find-Module, Find-Script` cmdlets from V2.
+It performs a search on a repository (server) based on the `-Name` parameter argument.
```
alerickson: We should add that a repository can be remote or local (server or file path).

Docs/DesignDoc.md
```diff
@@ -0,0 +1,307 @@
...
+    "Repository",
+    "PackageManagementProvider",
+    "AdditionalMetadata"
+}
+```
+
+#### Notes
+
+Search strategy will depend on what NuGet APIs support on the server.
+I believe V2 APIs do not support wildcard name searches, and that V3 APIs provide limited support.
```
alerickson: V2 APIs do support wildcard searches and V3 apis provided very limited support (Anam is providing extensive support client side).

Docs/DesignDoc.md
```diff
@@ -0,0 +1,307 @@
...
+But PowerShell has its own wildcard behavior that may not be compatible with the NuGet APIs.
+PowerShellGet V2 appears to simply retrieve all resources from the server, and then filter as necessary, when a wildcard or filter parameter is used.
```
alerickson: Most of the filtering is done server side (via OData queries), which is only available with v2

Docs/DesignDoc.md
alerickson: Definitely worth specifying since this type of version syntax will be new to PSGet

Docs/DesignDoc.md
PaulHigin: I think we can ignore parameters as appropriate.  But we need to make this clear in the documentation.

Docs/DesignDoc.md
PaulHigin: Yes, we can add this to a 'Notes' section since it will apply to many cmdlets.

Docs/DesignDoc.md
PaulHigin: Will add note that this needs to be discussed.

Docs/DesignDoc.md
PaulHigin: Thanks.  I'll you and Anam update so that it is accurate.  We may want a separate search strategy section, since it is pretty complex.

Docs/DesignDoc.md
PaulHigin: Good to know.  Will update.

Docs/DesignDoc.md
PaulHigin: Do we know which api version we want to use, and why?

Issue #279, opened 2020-10-04
Cmdlets have SupportsShouldProcess but don't actually implement it


ThomasNieto:
Cmdlets have `SupportsShouldProcess` but don't actually implement it for `-WhatIf` and `-Confirm` parameters. This gives users a false sense of security.

* Install-PSResource
* Publish-PSResource
* Save-PSResource
* Update-PSResource
* Uninstall-PSResource
* Register-PSResourceRepository
* Unregister-PSResourceRepository
* Set-PSResourceRepository

These don't need `SupportsShouldProcess` since they are non-state changing cmdlets.

* Get-PSResourceRepository
* Find-PSResource
* Get-PSResource

SydneyhSmith:
Thanks @ThomasNieto for calling this out, we do plan to implement

alerickson:
This is complete for all of the mentioned cmdlets except for Save-PSResource.  We'll add this in shortly.

alerickson:
Resolved with PR #447

Issue #258, opened 2020-09-16
First run experience is... lacking. Cannot find XML


Windos:
Testing PowerShellGet 3.0.0-beta10 with CompatPowerShellGet 0.0.2, on both PowerShell 7.0.3 and 7.1.0-preview.7.

First thing I did with the new module was try and install the SecretManagement module.

```powershell
Install-Module -Name Microsoft.PowerShell.SecretManagement -AllowPrerelease
```

I'm told Install-Module is deprecated (expected), but then get an error about not finding XML and being told to register a repo.

```output
WARNING: The cmdlet 'Install-Module' is deprecated, please use 'Install-PSResource'.
Install-PSResource: Was not able to successfully find xml. Try running 'Register-PSResourceRepository -PSGallery'
```

It seems that this error occurs when running *any* PowerShellGet command, including `Get-PSResourceRepository`.

As a first run experience with the module and expecting it to "just work", this is jarring. Surley PSGallery should be a default (as I'm sure most users will be using that, and only that... advanced users should be more than capable of removing it if needed?)

alerickson:
Thanks for the feedback!  I completely agree with you, we do if it on our short list of things to add and should get this out with the next release

Issue #249, opened 2020-09-10
Register-PSResourceRepository -Repositories Parameter Type


ThomasNieto:
The parameter type for `Register-PSResourceRepository -Repositories` is a `<List[hashtable]>` instead of the PowerShell standard of `[hashtable[]]` array.

SydneyhSmith:
Thanks @ThomasNieto happy to make the change, although curious is this is impacting your functionality in any way or are just looking for consistency

ThomasNieto:
@SydneyhSmith Its for consistency and ease of use. I would say that most PowerShell users don't know what generic types are and how to create them since none of the built-in cmdlets use them.

Issue #197, opened 2020-07-27
Parameters do not support wildcards


ThomasNieto:
In PowerShellGet v3 beta 7 parameters that users would expect to have wildcard support don't. For example `Get-PSResource -Name PS*` returns nothing same with `Get-PSResourceRepository -Name PS*`.

alerickson:
This will be resolved with the release of Preview 11.

PR #346, opened 2021-03-02
Refactoring *-PSResourceRepository cmdlets (Register, Get, Unregister)

anamnavi:
This PR includes refactoring for Register-PSResourceRepository, Get-PSResourceRepository, and Unregister-PSResourceRepository. Each of these cmdlets performs some CRUD operation on the repository store (APIs for this are in the RepositorySettings class). For the cmdlets that produce output I've created a strongly typed output class (PSRespositoryInfo). I've also added tests for each cmdlet. 

Notes: 
- I kept the preexisting method of deserialization for the repository xml file contents.  We control the contents of the xml object we deserialize and create them consistently but am open to a more consistent deserialization method congruent with how we will be doing it for the *-PSResource cmdlets.
- I will create a separate PR for the help docs for these cmdlets.

# PR Summary

- refactored Register-PSResourceRepository, Get-PSResourceRepository, Unregister-PSResourceRepository
- created PSRespositoryItem class
- refactored Add(), Read() and Remove() method in RepositorySettings.cs to return PSRepositoryItem objects (where needed)
- resolves issues:
  - #132 (`PassThru` parameter support for Get-PSResourceRepository cmdlet) 
  - #249 (`Repositories` parameter is of type `Hashtable[]`
  - #258 (Get-PSResourceRepository and Unregister-PSResourceRepository will create repository store xml file if it doesn't exist)
  - #197 (Get-PSResourceRepository supports wildcards)
  - #28 (Register-PSResourceRepository ensures repository uniqueness on `Name` parameter.)
  - #279 (Register-PSResourceRepository and Unregister-PSResourceRepository support and implement ShouldProcess, Get-PSResourceRepository does NOT)

## PR Checklist

    - [ ] None
    - **OR**
        - [ ] Issue filed: 
- **User-facing changes**
    - [ ] Not Applicable
    - **OR**
        - [ ] Issue filed: 
    - **OR**
- **Tooling**
    - **OR**


src/code/Get-PSResourceRepository.cs
```diff
@@ -0,0 +1,61 @@
...
+            set
+            { _name = value; }
+        }
+        private string[] _name = new string[0];
+        #endregion
+
+        #region Methods
+        protected override void ProcessRecord()
+        {
+            var r = new RepositorySettings();
```
alerickson: Something Paul had previously suggested  (that I definitely think we should do) is to make RepositorySettings a static class

src/code/RegisterPSResourceRepository.cs
```diff
@@ -0,0 +1,295 @@
...
+            { _priority = value; }
+        }
+        private int _priority = defaultPriority;
+
+        /// <summary>
+        /// Specifies a proxy server for the request, rather than a direct connection to the internet resource.
+        /// </summary>
+        [Parameter(ValueFromPipelineByPropertyName = true)]
+        [ValidateNotNullOrEmpty]
+        public Uri Proxy
```
alerickson: We should also be sure to document which parameters are not yet working yet.. I believe the Proxy/Proxy Credential params should be included in this.

src/code/ArgumentCompleter.cs
```diff
@@ -0,0 +1,44 @@
...
+    {
+        return CompleteRepositoryName(wordToComplete);
+    }
+
+
+    private IEnumerable<CompletionResult> CompleteRepositoryName(string wordToComplete)
+    {
+        List<CompletionResult> res = new List<CompletionResult>();
+
+        RepositorySettings repositorySettings = new RepositorySettings();
```
alerickson: Same comment as earliery about using a static Repository Settings class here.

src/code/ArgumentCompleter.cs
```diff
@@ -0,0 +1,44 @@
...
+        List<PSRepositoryItem> listOfRepositories = repositorySettings.Read(null);
+
+        foreach (PSRepositoryItem repo in listOfRepositories)
+         {
+            // string repoName = repo.Properties["Name"].Value.ToString();
```
alerickson: I think we can get rid of this comment

src/code/ArgumentCompleter.cs
```diff
@@ -0,0 +1,44 @@
...
+using System.Collections.Generic;
+using System.Management.Automation;
+using System.Management.Automation.Language;
+using Microsoft.PowerShell.PowerShellGet.RepositorySettings;
+using Microsoft.PowerShell.PowerShellGet.PSRepositoryItem;
+
+internal class RepositoryNameCompleter : IArgumentCompleter
+{
+    public IEnumerable<CompletionResult> CompleteArgument(
+        string commandName,  // For cmdlets Get-PSResource, Set-PSResource, and Unregister-PSResource
```
alerickson: We can either update the comment here or just remove the comment all together since it's not terribly helpful

src/code/ArgumentCompleter.cs
```diff
@@ -0,0 +1,44 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System;
+using System.Collections;
+using System.Collections.Generic;
+using System.Management.Automation;
+using System.Management.Automation.Language;
+using Microsoft.PowerShell.PowerShellGet.RepositorySettings;
+using Microsoft.PowerShell.PowerShellGet.PSRepositoryItem;
```
alerickson: Is this assembly needed?  Just a general comment to any unnecessary assemblies (these all very well may be needed here, but just wanted to mention that since I know there's a handful of files that have extraneous assemblies).

src/code/Get-PSResourceRepository.cs
```diff
@@ -0,0 +1,61 @@
...
+            List<PSRepositoryItem> items = r.Read(_name);
+
+            foreach (var repo in items)
```
alerickson: We should try to explicitly type variables that were previously implicitly types,  this 'var' for example

src/code/PSRepositoryItem.cs
```diff
@@ -0,0 +1,80 @@
...
+    /// This class contains information for a repository item.
+    public class PSRepositoryItem
+    {
+        private string _name;
+        private Uri _url;
+        private bool _trusted= false;
+        private int _priority = 50;
+
+
+        public PSRepositoryItem(string name, Uri url)
```
alerickson: Are you using these constructors anywhere?  If not we can get rid of the constructors and just use the auto-implemented getters and setters, eg:  `public string Name { get; set; };`

src/code/RegisterPSResourceRepository.cs
```diff
@@ -0,0 +1,295 @@
...
+
+namespace Microsoft.PowerShell.PowerShellGet.Cmdlets
+{
+    /// <summary>
+    /// The Register-PSResourceRepository cmdlet replaces the Register-PSRepository from V2.
+    /// It registers a repository for PowerShell modules.
+    /// The repository is registered to the current user's scope and does not have a system-wide scope.
+    /// </summary>
+
+    using RepositorySettings = Microsoft.PowerShell.PowerShellGet.RepositorySettings.RepositorySettings;
```
alerickson: You can just add this using statement with the other using statements at the top of the file and just include `Microsoft.PowerShell.PowerShellGet.RepositorySetting`

src/code/RegisterPSResourceRepository.cs
```diff
@@ -0,0 +1,295 @@
...
+
+            set
+            {
+                try
+                {
+                    _name = value;
+                }
+                catch (Exception e)
+                {
+                    throw new Exception("perhaps you used the wrong type for the parameter" + e.GetType());
```
alerickson: We should also get @SydneyhSmith's input on language for all the error messages

src/code/RegisterPSResourceRepository.cs
```diff
@@ -0,0 +1,295 @@
...
+        }
+        private SwitchParameter _passThru;
+        #endregion
+
+        /// <summary>
+        /// Process the repository being registered
+        /// </summary>
+        protected override void ProcessRecord()
+        {
+            WriteDebug("new cmdlet");
```
alerickson: I would add something a little more specific for the debug messaging here, maybe something like "Entering Register-PSRepository"  or "Register-PSRepository::ProcessRecord"

src/code/RegisterPSResourceRepository.cs
```diff
@@ -0,0 +1,295 @@
...
+                NameParameterSetHelper(r, items);
+            }
+            else if(ParameterSetName.Equals("RepositoriesParameterSet"))
+            {
+                RepositoriesParameterSetHelper(r, items);
+            }
+
+            if(_passThru)
+            {
+                foreach (var item in items)
```
alerickson: Same comment as earlier about explicit typing here.

src/code/RepositorySettings.cs
```diff
@@ -0,0 +1,302 @@
...
+            }
+
+            // Create will make a new XML if one doesn't already exist
+            try
+            {
+                CreateNewRepositoryXML();
+            }
+            catch
+            {
+                throw new ArgumentException("Was not able to successfully create xml");
```
alerickson: With the exceptions, we should be consistent with how we're using them.  We can discuss with Paul (I know we had briefly discussed error handling in the past) to figure out what we want to set as our standard for error handling in the repo.

alerickson (commented):
I left some comments on this but I just wanted to make a meta comment since this pops up a lot throughout the entire repo:  kind of a nit, but as we're going through all the files we should change any implicit types to be explicit

src/code/ArgumentCompleter.cs
```diff
@@ -0,0 +1,43 @@
...
+
+        RepositorySettings repositorySettings = new RepositorySettings();
+        List<PSRepositoryItem> listOfRepositories = repositorySettings.Read(null);
+
+        foreach (PSRepositoryItem repo in listOfRepositories)
+        {
+            string repoName = repo.Name;
+            if (repoName.StartsWith(wordToComplete, StringComparison.OrdinalIgnoreCase))
+            {
+                res.Add(new CompletionResult(repoName));
```
PaulHigin: We don't need to use the `res` list, and instead can use the C# `yield` keyword here, since the return type is `IEnumerable`.  For an example see:
https://github.com/PowerShell/SecretManagement/blob/9de1193e327291b2ec1b0f94680a69b5011bcf3c/src/code/SecretManagement.cs#L106

src/code/ArgumentCompleter.cs
```diff
@@ -0,0 +1,43 @@
...
+internal class RepositoryNameCompleter : IArgumentCompleter
+{
+    public IEnumerable<CompletionResult> CompleteArgument(
+        string commandName,  // For cmdlets Get-PSResource, Set-PSResource, and Unregister-PSResource
+        string parameterName, // For -Name parameter
+        string wordToComplete,
+        CommandAst commandAst,
+        IDictionary fakeBoundParameters)
+    {
+        return CompleteRepositoryName(wordToComplete);
```
PaulHigin: Probably don't need to have a separate private method here, and just copy the method body here, with the `yield` keyword.

src/code/Get-PSResourceRepository.cs
```diff
@@ -0,0 +1,61 @@
...
+        [ValidateNotNullOrEmpty]
+        public string[] Name
+        {
+            get
+            { return _name; }
+
+            set
+            { _name = value; }
+        }
+        private string[] _name = new string[0];
```
PaulHigin: We don't need a private field and can just use C# properties:
```csharp
public string[] Name { get; set; } = new string[0];
```

src/code/Get-PSResourceRepository.cs
```diff
@@ -0,0 +1,61 @@
...
+        RemotingCapability = RemotingCapability.None)]
+    public sealed
+    class GetPSResourceRepository : PSCmdlet
+    {
+        #region Parameters
+        /// <summary>
+        /// Specifies the name(s) of a registered repository to find.
+        /// Does not support wild card characters.
+        /// </summary>
+        [Parameter(Position = 0, ValueFromPipeline = true,
```
PaulHigin: Since there seems to be only one parameter set, you don't need the `ParameterSetName` field here.  But when you do, we usually set the name string as a constant field, so it can be referred to later:
```csharp
private const string NameParameterSet = "NameParameterSet";
```

For an example please see:
https://github.com/PowerShell/SecretManagement/blob/9de1193e327291b2ec1b0f94680a69b5011bcf3c/src/code/SecretManagement.cs#L1371

src/code/Get-PSResourceRepository.cs
```diff
@@ -0,0 +1,61 @@
...
+    /// </summary>
+
+    using RepositorySettings = Microsoft.PowerShell.PowerShellGet.RepositorySettings.RepositorySettings;
+    using PSRepositoryItem = Microsoft.PowerShell.PowerShellGet.PSRepositoryItem.PSRepositoryItem;
+
+    [Cmdlet(VerbsCommon.Get,
+        "PSResourceRepository",
+        SupportsShouldProcess = true,
+        HelpUri = "<add>",
+        RemotingCapability = RemotingCapability.None)]
```
PaulHigin: I don't think `RemotingCapability` is needed anymore, and you can remove it here.

PaulHigin (changes requested):

src/code/Get-PSResourceRepository.cs
anamnavi: agreed!

src/code/ArgumentCompleter.cs
anamnavi: removed

src/code/RegisterPSResourceRepository.cs
anamnavi: left this in here by accident; removed. If we do add Debug messages I think we can have a chat with @SydneyhSmith on language and where/how often it's used, etc.

src/code/ArgumentCompleter.cs
anamnavi: these ones are used, but I'll remove any unused assemblies from all files I touch while refactoring.

src/code/PSRepositoryItem.cs
anamnavi: using the one with 4 params here in RepositorySettings [here](https://github.com/anamnavi/PowerShellGet/blob/refactoring/src/code/RepositorySettings.cs#L134) when I create a PSRepositoryItem instance that contains the Name, URL, Trusted, Priority fields filled in with the XML document attributes. Removed the other constructor not being used.

src/code/ArgumentCompleter.cs
anamnavi: Didn't know about this C# syntactic sugar, thanks for pointing it out and for the example!

src/code/PSRepositoryItem.cs
```diff
@@ -0,0 +1,74 @@
...
+using System;
+using System.Management.Automation;
+
+namespace Microsoft.PowerShell.PowerShellGet.PSRepositoryItem
+{
+    /// <summary>
+    /// This class contains information for a repository item.
+    public class PSRepositoryItem
+    {
+        private string _name;
```
PaulHigin: Again, there is no need for private fields.  We can just use properties:
```csharp
public string Name { get; set; }
...
```

src/code/RegisterPSResourceRepository.cs
```diff
@@ -0,0 +1,289 @@
...
+
+        #endregion
+
+        #region Parameters
+        /// <summary>
+        /// Specifies name for the repository to be registered.
+        /// </summary>
+        [Parameter(Mandatory = true, Position = 0, ParameterSetName = NameParameterSet)]
+        [ValidateNotNullOrEmpty]
+        public string Name
```
PaulHigin: Please remove private fields where not needed (I'll stop reminding and assume all classes will be so modified :)).

src/code/RegisterPSResourceRepository.cs
```diff
@@ -0,0 +1,289 @@
...
+            get
+            { return _url; }
+
+            set
+            {
+                Uri url;
+                if(!(Uri.TryCreate(value, string.Empty, out url)
+                    && (url.Scheme == Uri.UriSchemeHttp || url.Scheme == Uri.UriSchemeHttps || url.Scheme == Uri.UriSchemeFtp || url.Scheme == Uri.UriSchemeFile)))
+                    {
+                        var message = string.Format(CultureInfo.InvariantCulture, "The URL provided is not valid: {0}", value);
```
PaulHigin: It would be nice to provide information about what URLs are valid:
```
The provided URL, {0}, is not valid.  A valid repository URL has the form ...
```

src/code/RegisterPSResourceRepository.cs
PaulHigin: We should also throw a `System.Management.Automation.PSNotImplementedException` error until it is implemented.

src/code/RegisterPSResourceRepository.cs
```diff
@@ -0,0 +1,289 @@
...
+        }
+        private SwitchParameter _passThru;
+        #endregion
+
+        /// <summary>
+        /// Process the repository being registered
+        /// </summary>
+        protected override void ProcessRecord()
+        {
+            List<PSRepositoryItem> items = new List<PSRepositoryItem>();
```
PaulHigin: Instead of passing in `items` in each helper method, have each helper method return a repository list?

src/code/RegisterPSResourceRepository.cs
```diff
@@ -0,0 +1,289 @@
...
+        }
+
+        private void PSGalleryParameterSetHelper(List<PSRepositoryItem> items)
+        {
+            var psGalleryUri = new Uri(PSGalleryRepoURL);
+            try
+            {
+                items.Add(RepositorySettings.Add(PSGalleryRepoName, psGalleryUri, _priority, _trusted));
+            }
+            catch (Exception e)
```
PaulHigin: Re-throwing `Exception` here doesn't seem useful.  But using `this.ThrowTerminatingError()` may be worthwhile as we can provide an errorId and augment the error message.  What errors doe RepositorySettings throw?  I assume it does not allow duplicate entries, but does it perform other validation?  In any case we should ensure the user error message is comprehensible and should have error tests.

src/code/RegisterPSResourceRepository.cs
```diff
@@ -0,0 +1,289 @@
...
+                throw new ArgumentException(String.Format("Use 'Register-PSResourceRepository -PSGallery' to register the PSGallery repository."));
+            }
+            else{
+                try
+                {
+                    items.Add(RepositorySettings.Add(_name, _url, _priority, _trusted));
+                }
+                catch(Exception e)
+                {
+                    throw new Exception(e.Message);
```
PaulHigin: Same comment for exception rethrow as above.

src/code/RegisterPSResourceRepository.cs
```diff
@@ -0,0 +1,289 @@
...
+                }
+            }
+
+        }
+
+        private void RepositoriesParameterSetHelper(List<PSRepositoryItem> items)
+        {
+            foreach(Hashtable repo in _repositories)
+            {
+                if(repo.ContainsKey(PSGalleryRepoName))
```
PaulHigin: I think that hashtable parameter validation can be done in a helper method, and maybe it returns required values, either user provided or default if parameter is not required.

src/code/RegisterPSResourceRepository.cs
```diff
@@ -0,0 +1,289 @@
...
+                    _priority = repo.ContainsKey("Priority") ? (int)repo["Priority"] : defaultPriority;
+                    _trusted = repo.ContainsKey("Trusted") ? (bool)repo["Trusted"] : defaultTrusted;
+                    PSGalleryParameterSetHelper(items);
+                    continue;
+                }
+
+                // if not PSGallery, then check that sufficient keys exist for NameParameterSet style register option
+                if(!repo.ContainsKey("Name") || String.IsNullOrEmpty(repo["Name"].ToString()))
+                {
+                    throw new System.ArgumentException(string.Format(CultureInfo.InvariantCulture, "Repository name cannot be null"));
```
PaulHigin: We should not throw a terminating error here, because it ignores any hashtable entries afterwards.  Instead we should `WriteError` and continue the loop.  This can be done in a Hashtable validation helper method.

src/code/RegisterPSResourceRepository.cs
```diff
@@ -0,0 +1,289 @@
...
+                {
+                        throw new System.ArgumentException(string.Format(CultureInfo.InvariantCulture, "Invalid Url"));
+                }
+
+                _name = repo["Name"].ToString();
+                _url = _repoURL;
+                _priority = repo.ContainsKey("Priority") ? Convert.ToInt32(repo["Priority"].ToString()) : defaultPriority;
+                _trusted = repo.ContainsKey("Trusted") ? Convert.ToBoolean(repo["Trusted"].ToString()) : defaultTrusted;
+
+                NameParameterSetHelper(items);
```
PaulHigin: This helper method throws a terminating exception that terminates the loop, which we don't want.  Maybe `NameParameterSetHelper` just write errors or returns a success boolean and error.

src/code/RepositorySettings.cs
```diff
@@ -0,0 +1,300 @@
...
+
+    using PSRepositoryItem = Microsoft.PowerShell.PowerShellGet.PSRepositoryItem.PSRepositoryItem;
+
+    static class RepositorySettings
+    {
+        /// <summary>
+        /// Default file name for a settings file is 'psresourcerepository.config'
+        /// Also, the user level setting file at '%APPDATA%\NuGet' always uses this name
+        /// </summary>
+        public static readonly string DefaultRepositoryFileName = "PSResourceRepository.xml";
```
PaulHigin: Do we ever allow the repository file name and location to ever change?  If not then maybe we should remove the 'default' from the variable names and update the comment.

src/code/RepositorySettings.cs
```diff
@@ -0,0 +1,300 @@
...
+        public static readonly string DefaultRepositoryPath = Path.Combine(Environment.GetFolderPath(SpecialFolder.LocalApplicationData), "PowerShellGet"); //"%APPDATA%/PowerShellGet";  // c:\code\temp\repositorycache
+        public static readonly string DefaultFullRepositoryPath = Path.Combine(DefaultRepositoryPath, DefaultRepositoryFileName);
+
+        /// <summary>
+        /// Find a repository XML
+        /// Returns:
+        /// </summary>
+        /// <param name="sectionName"></param>
+        public static bool FindRepositoryXML()
```
PaulHigin: `CheckRepository()`?  `XML` doesn't seem necessary in the method name.

src/code/RepositorySettings.cs
```diff
@@ -0,0 +1,300 @@
...
+        {
+            // Search in the designated location for the repository XML
```
PaulHigin:
```csharp
return File.Exists(DefaultFullRepositoryPath);
```

src/code/RepositorySettings.cs
```diff
@@ -0,0 +1,300 @@
...
+
+            return false;
+        }
+
+        /// <summary>
+        /// Create a new repository XML
+        /// Returns: void
+        /// </summary>
+        /// <param name="sectionName"></param>
+        public static void CreateNewRepositoryXML()
```
PaulHigin: `CreateRepository()`.  This only creates a new repository if one doesn't already exists.

src/code/RepositorySettings.cs
```diff
@@ -0,0 +1,300 @@
...
+        {
+            // Check to see if the file already exists; if it does return
+            if (FindRepositoryXML())
```
PaulHigin: I wonder if this should also do a content validation check.  What if a user (malicious or otherwise) modifies or overwrites the file with something that does not provide valid registry information.  Should we just create a new one (after emitting a warning)?

src/code/RepositorySettings.cs
```diff
@@ -0,0 +1,300 @@
...
+            {
+                return;
+            }
+
+            // create directory if needed
```
PaulHigin: Comment here and on line 66 probably not needed.

src/code/RepositorySettings.cs
```diff
@@ -0,0 +1,300 @@
...
+                Directory.CreateDirectory(DefaultRepositoryPath);
+            }
+
+            // If the repository xml file doesn't exist yet, create one
+            XDocument newRepoXML = new XDocument(
+                    new XElement("configuration")
+            );
+
+            // Should be saved in:
+            newRepoXML.Save(DefaultFullRepositoryPath);
```
PaulHigin: What happens if path/file creation fails with access denied or out of memory?  Do we let the file/io exception propagate?  I think in this case it might be good to wrap this in a PSInvalidOperation exception, with a message that provides context:
```
"Repository store creation failed with error: {0}."
```

src/code/RepositorySettings.cs
```diff
@@ -0,0 +1,300 @@
...
+        }
+
+        /// <summary>
+        /// Add a repository to the XML
```
PaulHigin: Add a repository to the store?

src/code/RepositorySettings.cs
```diff
@@ -0,0 +1,300 @@
...
+namespace Microsoft.PowerShell.PowerShellGet.RepositorySettings
+{
+    /// <summary>
+    /// The class contains basic information of a repository path settings as well as methods to
+    /// perform CRUD operations on the repositories.
+    /// </summary>
+
+    using PSRepositoryItem = Microsoft.PowerShell.PowerShellGet.PSRepositoryItem.PSRepositoryItem;
+
+    static class RepositorySettings
```
PaulHigin: This should probably have `internal` access.

src/code/RepositorySettings.cs
PaulHigin: I don't think it make sense to wrap in `ArgumentException`, and it is better to remove try/catch and let cmdlet provide a wrapper exception with a message that can provide more context.

src/code/RepositorySettings.cs
```diff
@@ -0,0 +1,300 @@
...
+            if (string.IsNullOrEmpty(repoName))
+            {
+                // throw new ArgumentException(Resources.Argument_Cannot_Be_Null_Or_Empty, nameof(sectionName));
+                throw new ArgumentException("Repository name cannot be null or empty");
+            }
+
+            // We expect the xml to exist, if it doesn't user needs to register a repository
+            try
+            {
+                FindRepositoryXML();
```
PaulHigin: I think there should be a single `CheckRepositoryStore()` method that creates the file based repository store if it doesn't exist, (or if the file is corrupted?).  It can throw an exception if creation fails (access, IO error).  If Get, Update, Remove fail because the repository does not exist in the store, then that is the error to be returned.

src/code/RepositorySettings.cs
```diff
@@ -0,0 +1,300 @@
...
+        }
+
+        /// <summary>
+        /// Updates a repository name, URL, priority, or installation policy
+        /// Returns:  void
+        /// </summary>
+        public static void Update(string repoName, Uri repoURL, int repoPriority, bool? repoTrusted)
+        {
+            // Check to see if information we're trying to add to the repository is valid
+            if (string.IsNullOrEmpty(repoName))
```
PaulHigin: I think we should do name,uri validation at a higher level instead of for each low level helper method.

src/code/RepositorySettings.cs
```diff
@@ -0,0 +1,300 @@
...
+            catch
+            {
+                throw new ArgumentException("Was not able to successfully create xml");
+            }
+
+            // Open file
+            XDocument doc = XDocument.Load(DefaultFullRepositoryPath);
+
+            // Check if what's being added already exists, if it does throw an error
+            var node = doc.Descendants("Repository").Where(e => string.Equals(e.Attribute("Name").Value, repoName, StringComparison.InvariantCultureIgnoreCase)).FirstOrDefault();
```
PaulHigin: We should have a 'Get()' method that returns repository information from a name look up.  It can be used for Update and Remove operations as well.

src/code/RepositorySettings.cs
```diff
@@ -0,0 +1,300 @@
...
+            catch
+            {
+                throw new ArgumentException("Was not able to successfully find xml. Try running 'Register-PSResourceRepository -PSGallery'");
+            }
+
+            // Open file
+            XDocument doc = XDocument.Load(DefaultFullRepositoryPath);
+
+            // Check if what's being updated is actually there first
+            var node = doc.Descendants("Repository").Where(e => string.Equals(e.Attribute("Name").Value, repoName, StringComparison.InvariantCultureIgnoreCase)).FirstOrDefault();
```
PaulHigin: Using XmlDoc class for this simple repository information seems too heavy weight to me.  I wonder if we should use a simple json file format, which is consistent with other PowerShell modules.  Also, if two PowerShell instances modify the store file at the same time, information is lost.  We might want to think about synchronizing.  If so, we can look at SecretManagement for an example.  Let's discuss more offline.

src/code/RepositorySettings.cs
```diff
@@ -0,0 +1,300 @@
...
+
+                // Remove item from file
+                node.Remove();
+            }
+
+            // Close the file
+            root.Save(DefaultFullRepositoryPath);
+        }
+
+        public static List<PSRepositoryItem> Read(string[] repoNames)
```
PaulHigin: Oh, I see that there is this `Read` method (instead of `Get`).

PaulHigin (changes requested):
These are a lot of comments.  Let's get together and talk about them and what design/style changes should be made.

Overall this is a great start!

src/code/RegisterPSResourceRepository.cs
```diff
@@ -0,0 +1,287 @@
...
```
anamnavi: this is where non terminating error handling should happen. Also NameParameterSet helper should not throw terminating error bc rest of repositories will not be considered/evaluated

src/code/RegisterPSResourceRepository.cs
```diff
@@ -0,0 +1,287 @@
...
+                }
+
+                Uri _repoURL;
+                if(!(Uri.TryCreate(repo["URL"].ToString(), UriKind.Absolute, out _repoURL)
+                    && (_repoURL.Scheme == Uri.UriSchemeHttp || _repoURL.Scheme == Uri.UriSchemeHttps || _repoURL.Scheme == Uri.UriSchemeFtp || _repoURL.Scheme == Uri.UriSchemeFile)))
+                {
+                        throw new System.ArgumentException(string.Format(CultureInfo.InvariantCulture, "Invalid Url"));
+                }
+
+                _name = repo["Name"].ToString();
```
anamnavi: pass in values, instead of using fields (implicit values where you can't be sure 100% of their state persisting)

src/code/RegisterPSResourceRepository.cs
anamnavi: problem, if possible what is wrong and how can be remedied (can be of this type, etc) -> general template

src/code/RegisterPSResourceRepository.cs
```diff
@@ -0,0 +1,287 @@
...
+        #endregion
+
+        /// <summary>
+        /// Process the repository being registered
+        /// </summary>
+        protected override void ProcessRecord()
+        {
+            List<PSRepositoryItem> items = new List<PSRepositoryItem>();
+
+            if(ParameterSetName.Equals(PSGalleryParameterSet)){
```
anamnavi: use switch statements instead

src/code/RegisterPSResourceRepository.cs
anamnavi: The `RepositorySettings` and `PSRepositoryItem` classes are now part of the `Microsoft.PowerShell.PowerShellGet.UtilClasses` namespace to avoid this workaround for using a namespace and class of same name.

src/code/RepositorySettings.cs
anamnavi: do we want to make this file tamper proof? Possible side effects of this (based on how tests are currently done, etc). If we wish to look at .NET libraries being used to detect this in SecretManagement

anamnavi: also can change to a json format file for the repository store..

src/code/RepositorySettings.cs
anamnavi: CheckRepositoryStore() and FindRepositoryStore() should be private

src/code/RepositorySettings.cs
anamnavi: rename to show it's conditionally creating repo store. CheckAndCreateRepositoryStore()

src/code/RepositorySettings.cs
anamnavi: at minimum reset state if corrupted, emit error and let user reset it.

src/code/RepositorySettings.cs
anamnavi: discuss with @SydneyhSmith about intended change

src/code/RepositorySettings.cs
anamnavi: instead of get helper to parse through xml document, just use Read with that name.

src/code/RepositorySettings.cs
anamnavi: CheckRepositoryStore() is public per our discussion

src/code/RepositorySettings.cs
```diff
@@ -0,0 +1,266 @@
...
+    /// <summary>
+    /// The class contains basic information of a repository path settings as well as methods to
+    /// perform CRUD operations on the repositories.
+    /// </summary>
+
+    internal static class RepositorySettings
+    {
+        /// <summary>
+        /// Default file name for a settings file is 'psresourcerepository.config'
+        /// Also, the user level setting file at '%APPDATA%\NuGet' always uses this name
```
PaulHigin: I am not sure what this comment means, but it appears we use `%LOCALAPPDATA%` instead of `%APPDATA%`.

src/code/RepositorySettings.cs
```diff
@@ -0,0 +1,266 @@
...
+
+
+        /// <summary>
+        /// Check if repository store xml file exists, if not then create
+        /// </summary>
+        public static void CheckRepositoryStore()
+        {
+            if(!File.Exists(FullRepositoryPath))
+            {
+                try{
```
PaulHigin: Style nitpick:  
We should be consistent on bracket placement (For csharp, usually on next line rather than same line).
Also spacing before '(' on 'if' statement:
```csharp
if (...)
```

instead of 
```csharp
if(...)
```

src/code/RepositorySettings.cs
```diff
@@ -0,0 +1,266 @@
...
+            {
+                throw new PSInvalidOperationException(String.Format("Reading from repository store failed: {0}", e.Message));
+            }
+
+            // Sort by priority, then by repo name
+            var reposToReturn = foundRepos.OrderBy(x => x.Priority).ThenBy(x => x.Name);
+            return reposToReturn.ToList();
+        }
+
+        private static XElement FindExistingRepositoryHelper(XDocument doc, string name)
```
PaulHigin: A better name, `FindRepositoryElement` ?

src/code/RepositorySettings.cs
```diff
@@ -0,0 +1,266 @@
...
+        {
+            XElement node = doc.Descendants("Repository").Where(e => string.Equals(e.Attribute("Name").Value, name, StringComparison.InvariantCultureIgnoreCase)).FirstOrDefault();
+            if(node != null)
```
PaulHigin: Can just return here, but it would be good to indent to make this more readable.
```csharp
return doc.Descendants("Repository").Where(
    e => string.Equals(
        e.Attribute("Name").Value,
        name,
        StringComparison.InvariantCultureIgnoreCase)).FirstOrDefault();
```

src/code/RepositorySettings.cs
```diff
@@ -0,0 +1,266 @@
...
+
+                // Close the file
+                root.Save(FullRepositoryPath);
+            }
+            catch(Exception e)
+            {
+                throw new PSInvalidOperationException(String.Format("Adding to repository store failed: {0}", e.Message));
+            }
+
+            PSRepositoryItem repoItem = new PSRepositoryItem(repoName, repoURL, repoPriority, repoTrusted);
```
PaulHigin: We don't really need the repoItem variable.

```csharp
return new RepositoryItem(repoName, ...)
```

src/code/RepositorySettings.cs
```diff
@@ -0,0 +1,266 @@
...
+                if(node == null)
+                {
+                    throw new ArgumentException("Cannot find the repository because it does not exist. Try registering the repository using 'Register-PSResourceRepository'");
+                }
+
+                // Else, keep going
+                // Get root of XDocument (XElement)
+                var root = doc.Root;
+
+                if (repoURL != null)
```
PaulHigin, line 135: It is Ok to have a null repository url?  It seems like the repository entry is useless without one.  Except maybe for PSGallery that has a fixed Url?  If that is the only case we should verify.

src/code/RepositorySettings.cs
```diff
@@ -0,0 +1,266 @@
...
+            try{
+                // Open file
+                XDocument doc = XDocument.Load(FullRepositoryPath);
+                if(repoNames == null || !repoNames.Any() || string.Equals(repoNames[0], "*") || repoNames[0] == null)
+                {
+                    // Name array or single value is null so we will list all repositories registered
+                    // iterate through the doc
+                    foreach(XElement repo in doc.Descendants("Repository"))
+                    {
+                        Uri thisUrl;
```
PaulHigin: This declaration can be removed.  Also it seems like we should test the result of TryCreate and return appropriate error, or return a null Uri value?  Not sure what is expected.
```csharp
if (!Uri.TryCreate(repo.Attribute("Url").Value, UriKind.Absolute, out Uri thisUrl))
{
    ...
}
```

src/code/RepositorySettings.cs
```diff
@@ -0,0 +1,266 @@
...
+                                thisUrl,
+                                Int32.Parse(node.Attribute("Priority").Value),
+                                Boolean.Parse(node.Attribute("Trusted").Value));
+
+                            foundRepos.Add(currentRepoItem);
+                        }
+                    }
+                }
+            }
+            catch (Exception e)
```
PaulHigin: Since we are processing multiple repository names, I feel we shouldn't have a catch all - throw here and instead return whatever results we can.  There can be a separate try/catch around the XDocument.Load() since if it fails we cannot proceed.  But if an exception is thrown in a processing loop we should save the error and continue with other entries.

We should change the method signature to something like:
``` csharp
public static List<PSRepositoryItem> Read(string[] repoNames, out string[] errorMsgs)
```
where the calling cmdlet can provide individual error records as needed.

PaulHigin (changes requested):
Added comments for the RepositorySettings class, but still need to review the other cmdlet implementations.

src/code/RepositorySettings.cs
anamnavi: discussed

src/code/RepositorySettings.cs
anamnavi: 1) changed declaration
2) I thought we could assume that repositories already registered will have to had correct URLs (that we tested can be created as URIs). I can add a check in case the user modifies the repository.xml and somehow adds a URL not of valid Uri type...if TryCreate fails I'll write an error and skip past that repository

Done..pass in cmdlet and write debug statements for each method

anamnavi: Instead: write debug at cmdlet level, and have them write the error I think.

src/code/RepositorySettings.cs
anamnavi: Done.

src/code/ArgumentCompleter.cs
```diff
@@ -0,0 +1,39 @@
...
+        string parameterName,
+        string wordToComplete,
+        CommandAst commandAst,
+        IDictionary fakeBoundParameters)
+    {
+        List<PSRepositoryItem> listOfRepositories;
+        try{
+            listOfRepositories = RepositorySettings.Read(null, out string[] _);
+        }
+        catch(PSInvalidOperationException e)
```
PaulHigin: Is there a reason to catch and rethrow this error?

src/code/Get-PSResourceRepository.cs
```diff
@@ -0,0 +1,88 @@
...
+    /// <summary>
+    /// The Get-PSResourceRepository cmdlet replaces the Get-PSRepository cmdlet from V2.
+    /// It searches for the PowerShell module repositories that are registered for the current user.
+    /// By default it will return all registered repositories, or if the -Name parameter argument is specified then it wil return the repository with that name.
+    /// It returns PSRepositoryItemInfo objects which describe each resource item found.
+    /// </summary>
+
+    [Cmdlet(VerbsCommon.Get,
+        "PSResourceRepository",
+        SupportsShouldProcess = true,
```
PaulHigin: We don't seem to be implementing `SupportsShouldProcess` so we should remove this attribute.  This cmdlet doesn't change any state so it doesn't seem necessary.

src/code/Get-PSResourceRepository.cs
```diff
@@ -0,0 +1,88 @@
...
+        [Parameter(Position = 0, ValueFromPipeline = true, ValueFromPipelineByPropertyName = true)]
+        [ArgumentCompleter(typeof(RepositoryNameCompleter))]
+        [ValidateNotNullOrEmpty]
+        public string[] Name { get; set; } = new string[0];
+        #endregion
+
+        #region Methods
+        protected override void ProcessRecord()
+        {
+            try
```
PaulHigin: This only needs to be called once.  Since `Name` parameter takes pipeline input, this block will be called for each name passed in the pipeline.  I think this check should be added to the `BeginProcessing()` block and executed only once per cmdlet instance.

src/code/Get-PSResourceRepository.cs
```diff
@@ -0,0 +1,88 @@
...
+            string[] errorMsgs;
+            List<PSRepositoryItem> items;
+            try
+            {
+                WriteDebug(String.Format("reading repository {0}. Calling Read() API now", Name));
+                items = RepositorySettings.Read(Name, out errorMsgs);
+            }
+            catch (PSInvalidOperationException e)
+            {
+                throw new Exception(string.Format("Unable to successfully read repository store. {0}", e.Message));
```
PaulHigin: I am wondering if we should throw here (and stop the running pipeline).  Could other names in the pipeline succeed?  If not then this is correct as is.

src/code/PSRepositoryItem.cs
```diff
@@ -0,0 +1,74 @@
...
+// Licensed under the MIT License.
+
+using System;
+using System.Management.Automation;
+
+namespace Microsoft.PowerShell.PowerShellGet.UtilClasses
+{
+    /// <summary>
+    /// This class contains information for a repository item.
+    public class PSRepositoryItem
```
PaulHigin: public sealed?

```suggestion
    public sealed class PSRepositoryItem
```

src/code/PSRepositoryItem.cs
```diff
@@ -0,0 +1,74 @@
...
+            _priority = priority;
+            _trusted = trusted;
+        }
+
+        /// <summary>
+        /// the Name of the repository
+        /// </summary>
+        public string Name
+        {
+            get
```
PaulHigin: Do we need the getters?  Is this object meant to be mutable?

src/code/RegisterPSResourceRepository.cs
```diff
@@ -0,0 +1,354 @@
...
+    /// <summary>
+    /// The Register-PSResourceRepository cmdlet replaces the Register-PSRepository from V2.
+    /// It registers a repository for PowerShell modules.
+    /// The repository is registered to the current user's scope and does not have a system-wide scope.
+    /// </summary>
+
+    [Cmdlet(VerbsLifecycle.Register,
+        "PSResourceRepository",
+        DefaultParameterSetName = NameParameterSet,
+        SupportsShouldProcess = true,
```
PaulHigin, line 23: This attribute makes sense for a register operation, but I don't see the code implementation.

src/code/RegisterPSResourceRepository.cs
```diff
@@ -0,0 +1,354 @@
...
+        [Parameter(ParameterSetName = NameParameterSet)]
+        [Parameter(ParameterSetName = PSGalleryParameterSet)]
+        [ValidateNotNullOrEmpty]
+        [ValidateRange(0, 50)]
+        public int Priority { get; set; } = defaultPriority;
+
+        /// <summary>
+        /// Specifies a proxy server for the request, rather than a direct connection to the internet resource.
+        /// </summary>
+        [Parameter(ValueFromPipelineByPropertyName = true)]
```
PaulHigin: Just curious why Proxy, ProxyCredential is passed in through a pipeline object?

PaulHigin (changes requested):
Just a few more comments.  Will finish tomorrow.

src/code/RegisterPSResourceRepository.cs
```diff
@@ -0,0 +1,354 @@
...
+                        ThrowTerminatingError(new ErrorRecord(
+                            new PSInvalidOperationException(e.Message),
+                            "ErrorInRepositoriesParameterSet",
+                            ErrorCategory.InvalidArgument,
+                            this));
+                    }
+                    break;
+
+                default:
+                    WriteDebug("Invalid parameter set");
```
PaulHigin: I would use a `Debug.Assert()` here, since this should never happen.

src/code/RegisterPSResourceRepository.cs
```diff
@@ -0,0 +1,354 @@
...
+                foreach (PSRepositoryItem repo in items)
+                {
+                    WriteObject(repo);
+                }
+            }
+        }
+
+        private PSRepositoryItem AddToRepositoryStoreHelper(string repoName, Uri repoUrl, int repoPriority, bool repoTrusted)
+        {
+            if (String.IsNullOrEmpty(repoName) || repoName.Contains(" ") || repoName.Contains("*"))
```
PaulHigin: Can use `string.IsNullOrWhiteSpace()` here.

src/code/RegisterPSResourceRepository.cs
```diff
@@ -0,0 +1,354 @@
...
+            {
+                throw new ArgumentException("Invalid url, must be one of the following Uri schemes: HTTPS, HTTP, FTP, File Based");
+            }
+            WriteDebug("All required values to add to repository provided, calling internal Add() API now");
+            return RepositorySettings.Add(repoName, repoUrl, repoPriority, repoTrusted);
+        }
+
+        private PSRepositoryItem NameParameterSetHelper(string repoName, Uri repoUrl, int repoPriority, bool repoTrusted)
+        {
+            if (repoName.Equals("PSGallery"))
```
PaulHigin: This should be a case insensitive compare:
```csharp
if (repoName.Equals("PSGallery", StringComparison.OrdinalIgnoreCase))
```

src/code/RegisterPSResourceRepository.cs
```diff
@@ -0,0 +1,354 @@
...
+                        reposAddedFromHashTable.Add(parsedRepoAdded);
+                    }
+                }
+            }
+            return reposAddedFromHashTable;
+        }
+
+        private PSRepositoryItem RepoValidationHelper(Hashtable repo)
+        {
+            if (!repo.ContainsKey("Name") || String.IsNullOrEmpty(repo["Name"].ToString()))
```
PaulHigin, line 307: Can repo name contain white spaces?  If so, the argument completer above will need to insert single quotes.  Ping me for more info.

src/code/RegisterPSResourceRepository.cs
```diff
@@ -0,0 +1,354 @@
...
+            {
+                WriteError(new ErrorRecord(
+                        new PSInvalidOperationException("Repository name cannot ne null"),
+                        "NullNameForRepositoriesParameterSetRegistration",
+                        ErrorCategory.InvalidArgument,
+                        this));
+                return null;
+            }
+            if (repo["Name"].ToString().Equals("PSGallery"))
```
PaulHigin, line 317: Case insensitive compare?

src/code/RegisterPSResourceRepository.cs
```diff
@@ -0,0 +1,354 @@
...
+        {
+            if (!repo.ContainsKey("Name") || String.IsNullOrEmpty(repo["Name"].ToString()))
+            {
+                WriteError(new ErrorRecord(
+                        new PSInvalidOperationException("Repository name cannot ne null"),
+                        "NullNameForRepositoriesParameterSetRegistration",
+                        ErrorCategory.InvalidArgument,
+                        this));
+                return null;
+            }
```
PaulHigin, line 315: PowerShell style guidelines state there should always be a new line after a closing bracket.

src/code/RegisterPSResourceRepository.cs
```diff
@@ -0,0 +1,354 @@
...
+                        new PSInvalidOperationException("Cannot register PSGallery with -Name parameter. Try: Register-PSResourceRepository -PSGallery"),
+                        "PSGalleryProvidedAsNameRepoPSet",
+                        ErrorCategory.InvalidArgument,
+                        this));
+                return null;
+            }
+            if (!repo.ContainsKey("Url") || String.IsNullOrEmpty(repo["Url"].ToString()))
+            {
+                WriteError(new ErrorRecord(
+                        new PSInvalidOperationException("Repository url cannot ne null"),
```
PaulHigin:
```suggestion
                        new PSInvalidOperationException("Repository url cannot be null"),
```

src/code/ArgumentCompleter.cs
anamnavi: There isn't. As per our discussion yesterday I removed it.

src/code/RepositorySettings.cs
```diff
@@ -0,0 +1,290 @@
...
+
+                if (repoURL != null)
+                {
+                    node.Attribute("Url").Value = repoURL.AbsoluteUri;
+                }
+                if (repoPriority >= 0)
+                {
+                    node.Attribute("Priority").Value = repoPriority.ToString();
+                }
+                // false, setting to true
```
PaulHigin: Can probably remove these comments?

src/code/RepositorySettings.cs
```diff
@@ -0,0 +1,290 @@
...
+            }
+            // Close the file
+            root.Save(FullRepositoryPath);
+            errorMsgs = temp.ToArray();
+        }
+
+        public static List<PSRepositoryItem> Read(string[] repoNames, out string[] errorMsgs)
+        {
+            errorMsgs = null;
+            List<string> temp = new List<string>();
```
PaulHigin: Nitpick, var name: errorList?

src/code/Get-PSResourceRepository.cs
anamnavi: No, if this error occurs other names in the pipeline would be run into the same error. I removed the try/catch per above comment and our discussion...so if this error occurs it'll be thrown from RepositorySettings, propagate up the stack and stop the running pipeline (which is fine since the other names in the pipeline couldn't have succeeded anyways).

src/code/Get-PSResourceRepository.cs
anamnavi: I agree

src/code/RepositorySettings.cs
```diff
@@ -0,0 +1,290 @@
...
+                        PSRepositoryItem currentRepoItem = new PSRepositoryItem(node.Attribute("Name").Value,
+                            thisUrl,
+                            Int32.Parse(node.Attribute("Priority").Value),
+                            Boolean.Parse(node.Attribute("Trusted").Value));
+
+                        foundRepos.Add(currentRepoItem);
+                    }
+                    if (!repoMatch)
+                    {
+                        temp.Add(String.Format("Unable to find repository matching Name '{0}'.  Use Get-PSResourceRepository to see all available repositories.", repo));
```
PaulHigin: We (PowerShell) don't emit an error if a wildcard search returns no results.  An error is emitted only if a specific name (no wildcard) is provided.

src/code/Unregister-PSResourceRepository.cs
```diff
@@ -0,0 +1,75 @@
...
+                WriteDebug(String.Format("removing repository {0}. Calling Remove() API now", Name));
+                RepositorySettings.Remove(Name, out errorMsgs);
+            }
+            catch (PSInvalidOperationException e)
+            {
+                throw new Exception(string.Format("Unable to successfully unregister repository due to issue reading repository store. {0}", e.Message));
+            }
+
+            // handle non-terminating errors
+            foreach (string error in errorMsgs)
```
PaulHigin: It looks like RepositorySettings class can return a null value for `errorMsgs`.  I think we should change it so that it never returns null, but an empty array for no errors.  That we you can use this code here to just iterate without first having to check for null value.  Otherwise you will need to add a check for null here.

src/code/Unregister-PSResourceRepository.cs
```diff
@@ -0,0 +1,75 @@
...
+            {
+                if (!String.IsNullOrEmpty(error))
```
PaulHigin: Since we control the internal class that reports these errors, I think you should just never return a null or empty error message, so the high level code here does not have to check each returned item.

test/GetPSResourceRepository.Tests.ps1
```diff
@@ -0,0 +1,99 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+Import-Module "$psscriptroot\PSGetTestUtils.psm1" -Force
+
+Describe "Test Register-PSResourceRepository" {
+    BeforeEach {
+
+        Get-NewPSResourceRepositoryFile
+        $tmpDir1Path = Join-Path -Path (Get-TempPath) -ChildPath "tmpDir1"
```
PaulHigin: You can use `$testdrive` built-in variable for creating test directories and files, which will be cleaned up automatically so you don't need to in an After block.

```powershell
$tmpDirPath = Join-Path -Path $testdrive -ChildPath "tmpDir1"
...
```

test/GetPSResourceRepository.Tests.ps1
```diff
@@ -0,0 +1,99 @@
...
+
+        $res = Get-PSResourceRepository -Name "testRepository","MyGallery"
+        foreach ($entry in $res) {
+            $entry.Name | Should -BeIn "testRepository","MyGallery"
+        }
+    }
+
+    It "not get repository that hasn't been registered/invalid name" {
+        $nonRegisteredRepoName = "nonRegisteredRepository"
+        $errorMsg = "Unable to find repository matching Name '$nonRegisteredRepoName'.  Use Get-PSResourceRepository to see all available repositories."
```
PaulHigin: We normally don't use error messages for testing because a) they are (sometimes) localized to different languages, and b) are subject to change.  Instead we use the ErrorRecord `FullyQualifiedErrorId` string as that should never change.
E.g.,

```powershell
It "Verifies Set-SecretInfo fails with expected unsupported error" {
            Set-SecretInfo -Name TestDefaultMeta -Metadata @{ Fail = $true } -ErrorVariable err 2>$null
            $err | Should -HaveCount 2
            $err[1].FullyQualifiedErrorId | Should -BeExactly 'SetSecretMetadataCommandNotSupported,Microsoft.PowerShell.SecretManagement.SetSecretInfoCommand'
        }
```

The `FullyQualifiedErrorId` is for the user to be able to uniquely specify an error for any language.
I don't know if/when we would localize error messages for this module so it is not as important here, but it might be good to follow PowerShell testing conventions.
```

**Review by PaulHigin — CHANGES_REQUESTED** [review]

File: src/code/Get-PSResourceRepository.cs (hunk 55 above)
Comment (anamnavi) [reply src/code/Get-PSResourceRepository.cs]: added a `BeginProcessing()` block for Get and Unregister which both support values from pipeline.

File: src/code/Unregister-PSResourceRepository.cs (hunk 71 above)
Comment (anamnavi) [reply src/code/Unregister-PSResourceRepository.cs]: I actually don't need to set it to null or even an empty array at the start. (I think I added this here when I was debugging some other issue for it that was resolved.)

File: test/GetPSResourceRepository.Tests.ps1 (hunk 73 above)
Comment (anamnavi) [reply test/GetPSResourceRepository.Tests.ps1]: I replaced the TempPath with $TestDrive, thanks for pointing this out! From Pester $TestDrive documentation: "When the Describe block is finished all contents of the TestDrive are discarded", so I still had to include the AfterAll block.

File: src/code/PSRepositoryItem.cs (hunk 58 above)
Comment (anamnavi) [reply src/code/PSRepositoryItem.cs]: remove setter, use Property in the constructor too.

File: src/code/RegisterPSResourceRepository.cs (hunk 60 above)
Comment (anamnavi) [reply src/code/RegisterPSResourceRepository.cs]: it shouldn't be until we implement it. Removed ValueFromPipeline tag from these parameters

File: src/code/RegisterPSResourceRepository.cs (hunk 59 above)
Comment (anamnavi) [reply src/code/RegisterPSResourceRepository.cs:23]: added, with target/action formatting. Also added `ShouldProcess` to Unregister-PSResourceRepository.

File: src/code/RegisterPSResourceRepository.cs [hunk 75]
```diff
@@ -0,0 +1,377 @@
...
+
+        /// <summary>
+        /// Specifies the priority ranking of the repository, such that repositories with higher ranking priority are searched
+        /// before a lower ranking priority one, when searching for a repository item across multiple registered repositories.
+        /// Valid priority values range from 0 to 50, such that a lower numeric value (i.e 10) corresponds
+        /// to a higher priority ranking than a higher numeric value (i.e 40). Has default value of 50.
+        /// </summary>
+        [Parameter(ParameterSetName = NameParameterSet)]
+        [Parameter(ParameterSetName = PSGalleryParameterSet)]
+        [ValidateNotNullOrEmpty]
```
PaulHigin: Unnecessary attribute for int type.

src/code/RegisterPSResourceRepository.cs
```diff
@@ -0,0 +1,377 @@
...
+        [ValidateRange(0, 50)]
+        public int Priority { get; set; } = defaultPriority;
+
+        /// <summary>
+        /// Specifies a proxy server for the request, rather than a direct connection to the internet resource.
+        /// </summary>
+        [ValidateNotNullOrEmpty]
```
PaulHigin, line 109: Missing `[parameter]` attribute

src/code/RegisterPSResourceRepository.cs
```diff
@@ -0,0 +1,377 @@
...
+        public Uri Proxy { get; set; }
+
+        /// <summary>
+        /// Specifies a user account that has permission to use the proxy server that is specified by the Proxy parameter.
+        /// </summary>
+        public PSCredential ProxyCredential { get; set; }
```
PaulHigin, line 116: Missing `[parameter]` attribute

src/code/RegisterPSResourceRepository.cs
```diff
@@ -0,0 +1,377 @@
...
+        [Parameter]
+        public SwitchParameter PassThru { get; set; }
+
+        #endregion
+
+        #region Methods
+
+        protected override void ProcessRecord()
+        {
+            if (Proxy != null || ProxyCredential != null)
```
PaulHigin, line 130: Please move to `BeginProcessing()` block, since values are not passed in via pipeline.

src/code/RegisterPSResourceRepository.cs
```diff
@@ -0,0 +1,377 @@
...
+                    "ParametersNotImplementedYet",
+                    ErrorCategory.NotImplemented,
+                    this));
+            }
+
+            try
+            {
+                RepositorySettings.CheckRepositoryStore();
+            }
+
```
PaulHigin: Please remove space.

src/code/RegisterPSResourceRepository.cs
```diff
@@ -0,0 +1,377 @@
...
+                ThrowTerminatingError(new ErrorRecord(
+                    new PSNotImplementedException("Proxy and ProxyCredential are not yet implemented. Please rerun cmdlet with other parameters."),
+                    "ParametersNotImplementedYet",
+                    ErrorCategory.NotImplemented,
+                    this));
+            }
+
+            try
+            {
+                RepositorySettings.CheckRepositoryStore();
```
PaulHigin, line 141: Can also be moved to `BeginProcessing()` block.

src/code/RegisterPSResourceRepository.cs
```diff
@@ -0,0 +1,377 @@
...
+                            this));
+                    }
+                    break;
+
+                default:
+                    Dbg.Assert(false, "Invalid parameter set");
+                    break;
+
+            }
+            if (PassThru)
```
PaulHigin, line 209: Please add space after bracket.

src/code/RegisterPSResourceRepository.cs
```diff
@@ -0,0 +1,377 @@
...
+                    }
+
+                    try
+                    {
+                        WriteDebug("(RepositoriesParameterSet): on repo: PSGallery. Registers PSGallery repository");
+                        reposAddedFromHashTable.Add(PSGalleryParameterSetHelper(
+                            repo.ContainsKey("Priority") ? (int)repo["Priority"] : defaultPriority,
+                            repo.ContainsKey("Trusted") ? (bool)repo["Trusted"] : defaultTrusted));
+                    }
+
```
PaulHigin: Please remove space

src/code/RegisterPSResourceRepository.cs
```diff
@@ -0,0 +1,377 @@
...
+                    catch (Exception e)
+                    {
+                        WriteError(new ErrorRecord(
+                            new PSInvalidOperationException(e.Message),
+                            "ErrorParsingIndividualRepoPSGallery",
+                            ErrorCategory.InvalidArgument,
+                            this));
+                    }
+                }
+
```
PaulHigin: Please remove space

src/code/RegisterPSResourceRepository.cs
```diff
@@ -0,0 +1,377 @@
...
+                else
+                {
+                    PSRepositoryInfo parsedRepoAdded = RepoValidationHelper(repo);
+                    if (parsedRepoAdded != null)
+                    {
+                        reposAddedFromHashTable.Add(parsedRepoAdded);
+                    }
+                }
+            }
+            return reposAddedFromHashTable;
```
PaulHigin, line 302: Please add space after bracket.

src/code/RegisterPSResourceRepository.cs
```diff
@@ -0,0 +1,377 @@
...
+
+            try
+            {
+                WriteDebug(String.Format("(RepositoriesParameterSet): on repo: {0}. Registers Name based repository", repo["Name"]));
+                return NameParameterSetHelper(repo["Name"].ToString(),
+                    repoURL,
+                    repo.ContainsKey("Priority") ? Convert.ToInt32(repo["Priority"].ToString()) : defaultPriority,
+                    repo.ContainsKey("Trusted") ? Convert.ToBoolean(repo["Trusted"].ToString()) : defaultTrusted);
+            }
+
```
PaulHigin: Please remove white space.

src/code/RepositorySettings.cs
```diff
@@ -0,0 +1,291 @@
...
+    /// perform CRUD operations on the repository store file.
+    /// </summary>
+
+    internal static class RepositorySettings
+    {
+        /// <summary>
+        /// File name for a user's repository store file is 'PSResourceRepository.xml'
+        /// The repository store file's location is currently only at '%LOCALAPPDATA%\NuGet' for the user account.
+        /// </summary>
+        public static readonly string RepositoryFileName = "PSResourceRepository.xml";
```
PaulHigin: I believe these can be `private` access as they are only used within the class.

PaulHigin (changes requested):
Just a handful of minor changes.

PaulHigin (approved):
LGTM

PaulHigin:
@anamnavi  Can you look at the test failures?

src/code/ArgumentCompleter.cs
```diff
@@ -0,0 +1,37 @@
...
+internal class RepositoryNameCompleter : IArgumentCompleter
+{
+    public IEnumerable<CompletionResult> CompleteArgument(
+        string commandName,
+        string parameterName,
+        string wordToComplete,
+        CommandAst commandAst,
+        IDictionary fakeBoundParameters)
+    {
+        List<PSRepositoryInfo> listOfRepositories = RepositorySettings.Read(null, out string[] _);
```
alerickson, line 20: Is the out variable for error?  And does it get used or passed on anywhere?

src/code/ArgumentCompleter.cs
```diff
@@ -0,0 +1,37 @@
...
+
+        wordToComplete = Utils.TrimQuotes(wordToComplete);
+        var wordToCompletePattern = WildcardPattern.Get(
```
alerickson, line 23: I'm a little confused about the wildcard within the argument completer.
The scenario I typically think of for argument completer is:  get-psresourcerepository ps  <tab through>.  
It looks like this is trying to handle the wildcard scenario  get-psresourcerepository ps* <tab through>?
Is that correct?

src/code/ArgumentCompleter.cs
anamnavi: Yes, the out variable is a array of error strings. RepositorySettings.Read() takes a `out string[]` as a parameter so when we call Read() from ArgumentCompleter we need to pass it a variable that the called method must modify. But we don't have any errors that get returned to ArgumentCompleter, so we pass a `_` variable which means we wish to discard whatever results it gives us. However, the out variable gets returned errors in Get-PSResourceRepository or Unregister-PSResourceRepository when we call Read() so this is just a case where we  must provide Read() the out variable but it's results are never interesting.

src/code/ArgumentCompleter.cs
anamnavi: Yes that's correct. Before the refactoring Argument Completer would check for any repositories that started with (via `StartsWith()` ) the (partial) repository name provided. So if I typed `Get-PSResourceRepository ps` it would find any repositories starting with ps . This basically does the same thing because it'll just search for repository whose names match ps* so it effectively does the StartsWith() check but without the StartsWith() call because now it uses a wildcard pattern matching.

I think the added benefit of this way is that if we have a Name that has a space in the middle (allowed use case) then when that becomes a wildcard pattern the space is included and handled.

src/code/RegisterPSResourceRepository.cs
```diff
@@ -0,0 +1,377 @@
...
+                        ThrowTerminatingError(new ErrorRecord(
+                            new PSInvalidOperationException(e.Message),
+                            "ErrorInRepositoriesParameterSet",
+                            ErrorCategory.InvalidArgument,
+                            this));
+                    }
+                    break;
+
+                default:
+                    Dbg.Assert(false, "Invalid parameter set");
```
alerickson, line 204: I'm not sure if this is needed since we have a default parameter set, anything that's not part of the PSGalleryParameterSet or the RepositoriesParameterSet should be part of the default parameter set

src/code/RepositorySettings.cs
```diff
@@ -0,0 +1,291 @@
...
+    /// </summary>
+
+    internal static class RepositorySettings
+    {
+        /// <summary>
+        /// File name for a user's repository store file is 'PSResourceRepository.xml'
+        /// The repository store file's location is currently only at '%LOCALAPPDATA%\NuGet' for the user account.
+        /// </summary>
+        private static readonly string RepositoryFileName = "PSResourceRepository.xml";
+        private static readonly string RepositoryPath = Path.Combine(Environment.GetFolderPath(SpecialFolder.LocalApplicationData), "PowerShellGet"); //"%APPDATA%/PowerShellGet";  // c:\code\temp\repositorycache
```
alerickson: We should delete the comment on this line

src/code/RepositorySettings.cs
```diff
@@ -0,0 +1,291 @@
...
+        }
+
+        /// <summary>
+        /// Add a repository to the store
+        /// Returns: PSRepositoryInfo containing information about the repository just added to the repository store
+        /// </summary>
+        /// <param name="sectionName"></param>
+        public static PSRepositoryInfo Add(string repoName, Uri repoURL, int repoPriority, bool repoTrusted)
+        {
+            Dbg.Assert(!string.IsNullOrEmpty(repoName), "Repository name cannot be null or empty");
```
alerickson, line 75: I'm curious about System.Diagnostics.Debug vs WriteDebug.  Is there a particular reason why the former is chosen?

src/code/RepositorySettings.cs
```diff
@@ -0,0 +1,291 @@
...
+                    bool repoMatch = false;
+                    WildcardPattern nameWildCardPattern = new WildcardPattern(repo, WildcardOptions.IgnoreCase);
+
+                    foreach (var node in doc.Descendants("Repository").Where(e => nameWildCardPattern.IsMatch(e.Attribute("Name").Value)))
+                    {
+                        repoMatch = true;
+                        if (!Uri.TryCreate(node.Attribute("Url").Value, UriKind.Absolute, out Uri thisUrl))
+                        {
+                            //debug statement
+                            temp.Add(String.Format("Unable to read incorrectly formatted URL for repo {0}", node.Attribute("Name").Value));
```
alerickson: This is a nit that I don't think is necessary to change but I personally feel like we should stay away from using variable names that are generic (ie temp).  I think it's appropriate here, just wanted to mention it as a high level comment.

src/code/UnregisterPSResourceRepository.cs
```diff
@@ -0,0 +1,75 @@
...
+        {
+            try
+            {
+                WriteDebug("Calling API to check repository store exists in non-corrupted state");
+                RepositorySettings.CheckRepositoryStore();
+            }
+            catch (PSInvalidOperationException e)
+            {
+                ThrowTerminatingError(new ErrorRecord(
+                    new PSNotImplementedException(e.Message),
```
alerickson: I'm not sure if 'PSNotImplementedException' is the correct error to be throwing here.  I think something along the lines of does not exist or incorrect format would be more appropriate here.

src/code/Utils.cs
```diff
@@ -0,0 +1,45 @@
...
+            }
+
+            return "'" + CodeGeneration.EscapeSingleQuotedStringContent(name) + "'";
+        }
+
+        #endregion
+    }
+
+    #endregion
+}
```
alerickson: I'm not sure why it complains about there being no new line here, but do you need to add one here?

src/code/RegisterPSResourceRepository.cs
```diff
@@ -0,0 +1,377 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System;
+using System.Collections;
+using System.Management.Automation;
+using System.Globalization;
```
alerickson: Another nit to have these in alphabetical order.

alerickson (approved):
Everything looks good, just left a few nit comments and a couple of general questions

src/code/RepositorySettings.cs
anamnavi: agree. I also updated the location in the comment on L 26

src/code/RepositorySettings.cs
anamnavi: good question :) `WriteDebug()` is declared in the abstract Cmdlet class. Since `RepositorySettings` doesn't inherit from `PSCmdlet` (which in turn inherits from `Cmdlet`) like the cmdlet classes do, I can't use WriteDebug there like I do in my cmdlet classes (i.e Register-PSResourceRepository class).

System.Diagnostics.Debug doesn't have that restriction as it's provided from the NuGet library to any class.

PaulHigin: This is a correct use of a debug assert.  This class/method is internal access and we control all code paths, and the contract is that arguments should never be null.  The assert is to ensure this, and I feel WriteDebug() is not necessary here.

src/code/RegisterPSResourceRepository.cs
PaulHigin: This is a correct use of debug assert as this is something that should never happen.  If down the line, someone adds a new parameter set but forgets to adds processing, this assert will show it immediately.

src/code/ArgumentCompleter.cs
PaulHigin: Personally, I am fine with it either way.  The WildcardPattern is not crucial and the previous match code works just as well.

src/code/RepositorySettings.cs
anamnavi: modified this to be `tempErrorList`

src/code/Utils.cs
anamnavi: added

src/code/RegisterPSResourceRepository.cs
alerickson: Okay, that makes sense

src/code/ArgumentCompleter.cs
anamnavi: I prefer the wildcard pattern way because it'll just be consistent with using the wildcard pattern in RepositorySettings Read() API (for Get-PSResourceRepository.cs) and any other future code we may refactor (i.e Find-PSResource.cs)

src/code/UnregisterPSResourceRepository.cs
anamnavi: just throw it.

PR #351, opened 2021-03-12
Abandoned: Refactor Get-InstalledPSResource cmdlet

alerickson:
# PR Summary
Still a few small updates that need to be made (see below).  This PR primarily addresses refactoring and completing functionality for 'Get-InstalledPSResource', which primarily touch GetInstalledPSResource.cs and GetHelper.cs.  A new class PSResourceInfo was created to output a concrete dotnet type in order to address formatting returned output. 

Get-InstalledPSResoure should retrieve a resource (PS module or PS script) that was installed via any version of PSGet. 

TODOs:
- [x] Ensure proper folder structure (in re to srcOld and src directories)
- [ ] "Complete" tests      
           - waiting for 'Install-PSResource' to be merged before updating the tests to use this cmdlet in order to test 'Get-InstalledPSResource' functionality 
- [x] Add 'help' info

## PR Checklist

    - [x] None
    - **OR**
        - [ ] Issue filed: 
- **User-facing changes**
    - [x] Not Applicable
    - **OR**
        - [ ] Issue filed: 
    - **OR**
- **Tooling**
    - **OR**


src/code/PSResourceInfo.cs
```diff
@@ -0,0 +1,66 @@
+
```
anamnavi: add license info to file (at least I've done that so if we don't need to then I can remove for consistency sake)

src/code/PSResourceInfo.cs
```diff
@@ -0,0 +1,66 @@
+
+using System;
+using System.Collections;
+using System.Collections.Generic;
+
+namespace Microsoft.PowerShell.PowerShellGet
```
anamnavi: I added my concrete dotnet type output class (PSRepositoryInfo.cs) to the `Microsoft.PowerShell.PowerShellGet.UtilClasses` namespace...it'll be helpful to have them both be in the same namespace for consistency

srcOld/code/GetInstalledPSResource.cs
```diff
@@ -0,0 +1,51 @@
...
+using System.Management.Automation;
+using System.Threading;
+
+namespace Microsoft.PowerShell.PowerShellGet.Cmdlets
+{
+    /// <summary>
+    /// It retrieves a resource that was installEd with Install-PSResource
+    /// Returns a single resource or multiple resource.
+    /// </summary>
+    [Cmdlet(VerbsCommon.Get, "InstalledPSResource", HelpUri = "<add>", RemotingCapability = RemotingCapability.None)]
```
anamnavi, line 13: @PaulHigin said we could remove RemotingCapability as it's not used.

src/PSGet.Format.ps1xml
```diff
@@ -0,0 +1,27 @@
...
+                            <TableColumnItem><PropertyName>Name</PropertyName></TableColumnItem>
+                            <TableColumnItem><PropertyName>Version</PropertyName></TableColumnItem>
+                            <TableColumnItem><PropertyName>Description</PropertyName></TableColumnItem>
+      	                </TableColumnItems>
+      		    </TableRowEntry>
+                </TableRowEntries>
+            </TableControl>
+        </View>
+    </ViewDefinitions>
+</Configuration>
```
anamnavi: add new line here.

src/PowerShellGet.psd1
```diff
@@ -1,129 +1,130 @@
...
+### 3.0.0-beta1
+BREAKING CHANGE
+* Preview version of PowerShellGet. Many features are not fully implemented yet. Please see https://devblogs.microsoft.com/powershell/powershellget-3-0-preview1 for more details.
+
+'@
+        }
+    }
+
+    HelpInfoURI       = 'http://go.microsoft.com/fwlink/?linkid=855963'
+}
```
anamnavi: add new line here too

src/code/GetHelper.cs
```diff
@@ -0,0 +1,389 @@
...
+using static System.Environment;
+using System.Globalization;
+using System.IO;
+using System.Linq;
+using System.Management.Automation;
+using System.Net;
+using System.Runtime.InteropServices;
+using System.Text.RegularExpressions;
+using System.Threading;
+using Microsoft.PowerShell.PowerShellGet.RepositorySettings;
```
anamnavi: is this class being used? If so I can't see where in the file it is...but just a heads up that when my changes for RepositorySettings get merged in if you use it here you may need to update some of your code and update the using statement to be `using Microsoft.PowerShell.PowerShellGet.UtilClasses` and access the class from that namespace

src/code/GetHelper.cs
```diff
@@ -0,0 +1,389 @@
...
+using NuGet.Configuration;
+using NuGet.Packaging.Core;
+using NuGet.Protocol;
+using NuGet.Protocol.Core.Types;
+using NuGet.Versioning;
+
+namespace Microsoft.PowerShell.PowerShellGet.Cmdlets
+{
+    /// <summary>
+    /// Find helper class
```
anamnavi: Do you mean `Get helper class`?

src/code/GetHelper.cs
```diff
@@ -0,0 +1,389 @@
...
+        private string programFilesPath;
+        private string myDocumentsPath;
+
+        public GetHelper(CancellationToken cancellationToken, PSCmdlet cmdletPassedIn)
+        {
+            this.cancellationToken = cancellationToken;
+            this.cmdletPassedIn = cmdletPassedIn;
+        }
+
+        public List<PSObject> ProcessGetParams(string[] name, string version, bool prerelease, string path)
```
anamnavi: could we return `List<PSResourceInfo>` instead?

src/code/GetHelper.cs
```diff
@@ -0,0 +1,389 @@
...
+                    myDocumentsPath = System.IO.Path.Combine(Environment.GetFolderPath(SpecialFolder.LocalApplicationData), "powershell");
+                    programFilesPath = System.IO.Path.Combine("/usr", "local", "share", "powershell");
+                }
+#endif
+                cmdletPassedIn.WriteDebug(string.Format("Current user scope path: '{0}'", myDocumentsPath));
+                cmdletPassedIn.WriteDebug(string.Format("All users scope path: '{0}'", programFilesPath));
+
+                /*** Will search first in PSModulePath, then will search in default paths ***/
+
+                foreach (var modulePath in modulePaths)
```
anamnavi: have the explicit type of modulePath instead of var if possible.

src/code/GetHelper.cs
```diff
@@ -0,0 +1,389 @@
...
+                {
+                    dirsToSearch.AddRange(Directory.GetFiles(mdScriptsPath).ToList());
+                }
+
+
+                // uniqueify 
+                dirsToSearch = dirsToSearch.Distinct().ToList();
+            }
+
+            foreach (var dir in dirsToSearch)
```
anamnavi: same as above, use explicit type instead of var

src/code/GetHelper.cs
```diff
@@ -0,0 +1,389 @@
...
+                            installedPkgsToReturn.Add(versionPath);
+                        }
+                    }
+                }
+            }
+            else
+            {
+                cmdletPassedIn.WriteDebug("No version provided-- check each path for the requested package");
+                // if no version is specified, just get the latest version
+                foreach (var pkgPath in dirsToSearch)
```
anamnavi: same comment about explicit type of pkgPath instead of var. I'm going to stop commenting on these now and assume it gets switched in all future instances.

src/code/GetHelper.cs
```diff
@@ -0,0 +1,389 @@
...
+                        cmdletPassedIn.WriteDebug(string.Format("Searching through package version path: '{0}'", versionPath));
+                        NuGetVersion dirAsNugetVersion;
+                        var dirInfo = new DirectoryInfo(versionPath);
+                        NuGetVersion.TryParse(dirInfo.Name, out dirAsNugetVersion);
+                        cmdletPassedIn.WriteDebug(string.Format("Directory parsed as NuGet version: '{0}'", dirAsNugetVersion));
+
+                        if (versionRange.Satisfies(dirAsNugetVersion))
+                        {
+                            // just search scripts paths
+                            if (pkgPath.ToLower().Contains("Scripts"))
```
anamnavi: do you mean `pkgPath.ToLower().Contains("scripts")` since an uppercase S in scripts won't be present after ToLower()? Or maybe do IgnoreCase()?

src/code/GetHelper.cs
```diff
@@ -0,0 +1,389 @@
...
+
+                        // Check if the pkg path actually has version sub directories.
+                        if (versionsDirs.Length != 0)
+                        {
+                            Array.Sort(versionsDirs, StringComparer.OrdinalIgnoreCase);
+                            Array.Reverse(versionsDirs);
+
+                            var pkgXmlFilePath = System.IO.Path.Combine(versionsDirs.First(), "PSGetModuleInfo.xml");
+
+                            // TODO:  check if this xml file exists, if it doesn't check if it exists in a previous version
```
anamnavi: does this check still need to happen or it does and we can remove TODO comment

src/code/GetHelper.cs
```diff
@@ -0,0 +1,389 @@
...
+                        powerShellGetFormatVersionInfo = deserializedObj.Properties.Match("PowerShellGetFormatVersion");
+                        releaseNotesInfo = deserializedObj.Properties.Match("ReleaseNotes");
+                        dependenciesInfo = deserializedObj.Properties.Match("Dependencies");
+                        repositorySourceLocationInfo = deserializedObj.Properties.Match("RepositorySourceLocation");
+                        repositoryInfo = deserializedObj.Properties.Match("Repository");
+                        additionalMetadataInfo = deserializedObj.Properties.Match("AdditionalMetadata");
+                        installedLocationInfo = deserializedObj.Properties.Match("InstalledLocation");
+
+
+                        /* // testing adding prerelease parameter
```
anamnavi: will we be adding prerelease support...do we need to uncomment this?

src/code/GetHelper.cs
```diff
@@ -0,0 +1,389 @@
...
+                    // we still want all pkgs to be returned if -Prerelease is passed in as a param
+                    //if ((_prerelease == false && isPrelease == false) || _prerelease == true)
+                    //{
+
+                    //var foundPkgs = List<PSObject>();
+                    
+                    PSObject pkgAsPSObject = new PSObject();
+                    try
+                    {
+                        pkgAsPSObject.Members.Add(new PSNoteProperty("Name", nameInfo.Any()? nameInfo.FirstOrDefault().Value : string.Empty));
```
anamnavi: Instead of PSObject this should be saved as an instance of PSResourceInfo object

src/code/GetHelper.cs
```diff
@@ -0,0 +1,389 @@
...
+                        pkgAsPSObject.Members.Add(new PSNoteProperty("AdditionalMetadata", additionalMetadataInfo.Any()? additionalMetadataInfo.FirstOrDefault().Value : string.Empty));
+                        pkgAsPSObject.Members.Add(new PSNoteProperty("InstalledLocation", installedLocationInfo.Any()? installedLocationInfo.FirstOrDefault().Value : string.Empty));
+
+
+                        // verify that this works, then add the object to a list and return it
+                        //WriteObject(pkgAsPSObject);
+                        foundInstalledPkgs.Add(pkgAsPSObject);
+
+                    }
+                    catch { }
```
anamnavi: What's done in the catch? Should this throw a terminating error (`ThrowTerminatingError()`) or non-terminating error (`WriteError()`)?

src/code/GetHelper.cs
```diff
@@ -0,0 +1,389 @@
...
+                    
+                }
+            }
+
+            // return here 
+            return foundInstalledPkgs;
+        }
+    }
+}
```
anamnavi: add newline

src/code/GetInstalledResource.cs
```diff
@@ -0,0 +1,51 @@
...
+            WriteDebug("Entering GetInstalledPSResource");
+
+            GetHelper getHelper = new GetHelper(cancellationToken, this);
+            foreach (PSResourceInfo psObject in getHelper.ProcessGetParams(Name, Version, Path))
+            {
+                WriteObject(psObject);
+            }
+        }
+    }
+}
```
anamnavi: add new line

src/code/GetInstalledResource.cs
```diff
@@ -0,0 +1,51 @@
...
+        
+        protected override void ProcessRecord()
+        {
+            CancellationTokenSource source = new CancellationTokenSource();
+            CancellationToken cancellationToken = source.Token;
+
+            WriteDebug("Entering GetInstalledPSResource");
+
+            GetHelper getHelper = new GetHelper(cancellationToken, this);
+            foreach (PSResourceInfo psObject in getHelper.ProcessGetParams(Name, Version, Path))
```
anamnavi: are there any errors that could come from ProcessGetParams that we need to handle here (via a try/catch)? @PaulHigin may have a better idea about the error handling needed in `ProcessGetParams()` as he helped me recognize that for Register-PSResourceRepository's helper functions/classes.

src/code/PSResourceInfo.cs
```diff
@@ -0,0 +1,66 @@
...
+            public enum VersionType
+            {
+                Unknown,
+                MinimumVersion,
+                RequiredVersion,
+                MaximumVersion
+            }
+        }
+    }
+}
```
anamnavi: new line needed, but rest of this LGTM

srcOld/code/GetHelper.cs
```diff
@@ -1,389 +1,389 @@
-﻿// Copyright (c) Microsoft Corporation. All rights reserved.
```
anamnavi, line 1: were there any changes made to this file? I thought our intention was to not modify any of the `srcOld` files..if we don't need to make any changes to it perhaps we can remove this file off this PR?

srcOld/code/GetInstalledPSResource.cs
```diff
@@ -0,0 +1,51 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
```
anamnavi: I thought our intention was to not modify any of the srcOld files..if we don't need to make any changes to it perhaps we can remove this file off this PR?

srcOld/code/PSResourceInfo.cs
```diff
@@ -0,0 +1,66 @@
+
```
anamnavi: I thought our intention was to not modify any of the srcOld files (and not create any new ones there either)..if we don't need to make any changes to it perhaps we can remove this file off this PR since we already have a PSResourceInfo.cs class in `src`?

test/GetInstalledResource.Tests.ps1
```diff
@@ -0,0 +1,103 @@
+# Copyright (c) Microsoft Corporation.
```
anamnavi, line 1: tests look good to me so far! One comment Steve had last time was that we don't need the `Purpose, Action, Expected comments before each test...the It "<test description>" is probably clear enough. I know the Find-PSResource tests follow the old comment style though, but I believe you could remove them here. Overall, good use of the Pester templating!

anamnavi (changes requested):
Overall looks good to me. I left a couple of comments but a couple of meta comments:
- I think we should change any implicit types to explicit
- most files needed a new line added at the end
- do we need to modify the srcOld files? If not I think we can just remove them entirely from the PR
- do we need error handling in Get-Helper GetProcessParams()?

srcOld/code/GetInstalledPSResource.cs
anamnavi: deleting because this change shouldn't be made to srcOld file

src/code/GetHelper.cs
anamnavi: remove directive

srcOld/code/GetHelper.cs
```diff
@@ -1,389 +1,389 @@
...
-using static System.Environment;
-using System.Globalization;
-using System.IO;
-using System.Linq;
-using System.Management.Automation;
-using System.Net;
-using System.Runtime.InteropServices;
-using System.Text.RegularExpressions;
-using System.Threading;
-using Microsoft.PowerShell.PowerShellGet.RepositorySettings;
```
anamnavi, line 16: remove this directive

src/PSGet.Format.ps1xml
```diff
@@ -0,0 +1,27 @@
...
+<Configuration>
+    <ViewDefinitions>
+        <View>
+            <Name>PSResourceInfo</Name>
+            <ViewSelectedBy>
+                <TypeName>Microsoft.PowerShell.PowerShellGet.PSResourceInfo</TypeName>
+	    </ViewSelectedBy>
+    	    <TableControl>
+     		<TableHeaders>
+      		    <TableColumnHeader><Label>Name</Label></TableColumnHeader>
```
PaulHigin: Nitpick.  Fix indentation.

src/PSModule.psm1
```diff
@@ -9,11 +9,9 @@ $isCore = ($PSVersionTable.Keys -contains "PSEdition") -and ($PSVersionTable.PSE
 if ($isCore)
 {
     $script:Framework = 'netstandard2.0'
-    $script:FrameworkToRemove = 'net472'
```
PaulHigin: Weren't we going to move to a single framework dependency?  PSReadline and SecretManagement both use 'net461' and I wonder if that is enough here.

I think our refactor should include this change if we can make it.

src/PowerShellGet.psd1
```diff
@@ -1,129 +1,130 @@
-@{
-    RootModule        = 'PSModule.psm1'
-    ModuleVersion     = '3.0.0'
-    GUID              = '1d73a601-4a6c-43c5-ba3f-619b18bbb404'
-    Author            = 'Microsoft Corporation'
-    CompanyName       = 'Microsoft Corporation'
-    Copyright         = '(c) Microsoft Corporation. All rights reserved.'
-    Description       = 'PowerShell module with commands for discovering, installing, updating and publishing the PowerShell artifacts like Modules, DSC Resources, Role Capabilities and Scripts.'
```
PaulHigin, line 8: Are we still supporting 'Role Capabilities'?  I thought we were removing it for our first release.

src/code/GetHelper.cs
PaulHigin: I think var is correct here since there is no question about type.

src/code/GetHelper.cs
```diff
@@ -0,0 +1,389 @@
...
+using NuGet.Protocol;
+using NuGet.Protocol.Core.Types;
+using NuGet.Versioning;
+
+namespace Microsoft.PowerShell.PowerShellGet.Cmdlets
+{
+    /// <summary>
+    /// Find helper class
+    /// </summary>
+    class GetHelper : PSCmdlet
```
PaulHigin: Since a PSCmdlet object is passed in to use, do we still need to derive this class from PSCmdlet class?  Also we should explicitly define the accessibility (probably 'internal').

src/code/GetHelper.cs
```diff
@@ -0,0 +1,389 @@
...
+                dirsToSearch.AddRange(Directory.GetDirectories(path).ToList());
+            }
+            else
+            {
+                // PSModules path
+                var psModulePath = Environment.GetEnvironmentVariable("PSModulePath");
+                var modulePaths = psModulePath.Split(';');
+
+
+#if NET472
```
PaulHigin: Do we need this #ifdef?  I believe these APIs will work equally well on either framework.  If we just need to know what version of PowerShell is running, this can be a runtime check.

src/code/GetHelper.cs
```diff
@@ -0,0 +1,389 @@
...
+                    programFilesPath = System.IO.Path.Combine(Environment.GetFolderPath(SpecialFolder.ProgramFiles), "PowerShell");
+                }
+                else
+                {
+                    // Paths are the same for both Linux and MacOS
+                    myDocumentsPath = System.IO.Path.Combine(Environment.GetFolderPath(SpecialFolder.LocalApplicationData), "powershell");
+                    programFilesPath = System.IO.Path.Combine("/usr", "local", "share", "powershell");
+                }
+#endif
+                cmdletPassedIn.WriteDebug(string.Format("Current user scope path: '{0}'", myDocumentsPath));
```
PaulHigin: I think it is better to use 'WriteVerbose' instead of 'WriteDebug', as most users use the '-Verbose' switch to get more details on running operations.

src/code/GetHelper.cs
PaulHigin: Instead of having a separate foreach to write search directories, you can just add the 'WriteVerbose' above where each directory is added.

src/code/GetHelper.cs
```diff
@@ -0,0 +1,389 @@
...
+
+            if (path != null)
+            {
+                cmdletPassedIn.WriteDebug(string.Format("Provided path is: '{0}'", path));
+                dirsToSearch.AddRange(Directory.GetDirectories(path).ToList());
+            }
+            else
+            {
+                // PSModules path
+                var psModulePath = Environment.GetEnvironmentVariable("PSModulePath");
```
PaulHigin: These should be defined below, right next to where they are used.

src/code/GetHelper.cs
PaulHigin: Nitpick.  `name` should be `names`, then inside loops you can use a single `name` variable.

src/code/GetHelper.cs
```diff
@@ -0,0 +1,389 @@
...
+            {
+                cmdletPassedIn.WriteDebug(string.Format("All directories to search: '{0}'", dir));
+            }
+
+            // Or a list of the passed in names
+            if (name != null && !name[0].Equals("*"))
+            {
+                var nameLowerCased = new List<string>();
+                var scriptXMLnames = new List<string>();
+                Array.ForEach(name, n => nameLowerCased.Add(n.ToLower()));
```
PaulHigin: Both nameLowerCased and scriptXMLnames lists can both be filled in one foreach loop, with two statements in the lambda block.

src/code/GetHelper.cs
```diff
@@ -0,0 +1,389 @@
...
+                Array.ForEach(name, n => scriptXMLnames.Add((n + "_InstalledScriptInfo.xml").ToLower()));
+
+                dirsToSearch = dirsToSearch.FindAll(p => (nameLowerCased.Contains(new DirectoryInfo(p).Name.ToLower())
```
PaulHigin: I am not clear what this does, but it appears to be filtering the directory search list to provided directory names.  This should be made clear in the comment above.  So the method `name` parameter should probably be `dirName` or `modName` or `resourceName` and the comment should make clear how it is used.

Also converting to lower case and allocation is not good practice in a loop.  If we want to do case insensitive matches, we should use .Net string comparer with `IgnoreCase`.  

You could use a `HashSet<string>` for unique lists such as dirsToSearch, and construct it with a case insensitive string comparer.  It will guarantee uniqueness and you don't need to use `dirsToSearch.Disctinct()`.

src/code/GetHelper.cs
```diff
@@ -0,0 +1,389 @@
...
+                    || scriptXMLnames.Contains((System.IO.Path.GetFileName(p)).ToLower())));
+
+                cmdletPassedIn.WriteDebug(dirsToSearch.Any().ToString());
```
PaulHigin: I agree WriteDebug is better than WriteVerbose in this case, since it prints dirs multiple times after filtering/processing.  WriteVerbose should inform on progression steps and WriteDebug provides more details on how state varies.

src/code/GetHelper.cs
```diff
@@ -0,0 +1,389 @@
...
+            }
+
+            // try to parse into a specific NuGet version
+            VersionRange versionRange = null;
+            if (version != null)
+            {
+                NuGetVersion specificVersion;
```
PaulHigin: Don't need to declare specificVersion variable above.
Does Version.TryParse() exist?  If so we should use it.

```csharp
// Try parsing as version
if (NuGetVersion.TryParse(version, out NuGetVersion specificVersion)
{
    ...
}
else if (Version.TryParse(...))
{

}
```

src/code/GetHelper.cs
```diff
@@ -0,0 +1,389 @@
...
+                {
+                    // exact version
+                    versionRange = new VersionRange(specificVersion, true, specificVersion, true, null, null);
+                    cmdletPassedIn.WriteDebug(string.Format("A specific version, '{0}', is specified", versionRange.ToString()));
+
+                }
+                else
+                {
+                    // check if version range
+                    versionRange = VersionRange.Parse(version);
```
PaulHigin: Can this throw?  We should use TryParse if available.  Also what if version is non-null but doesn't parse to a valid version or version range?  Do we ignore, emit a warning, or terminate with error?

src/code/GetHelper.cs
```diff
@@ -0,0 +1,389 @@
...
+                }
+#endif
+                cmdletPassedIn.WriteDebug(string.Format("Current user scope path: '{0}'", myDocumentsPath));
+                cmdletPassedIn.WriteDebug(string.Format("All users scope path: '{0}'", programFilesPath));
+
+                /*** Will search first in PSModulePath, then will search in default paths ***/
+
+                foreach (var modulePath in modulePaths)
+                {
+                    dirsToSearch.AddRange(Directory.GetDirectories(modulePath).ToList());
```
PaulHigin: I think we should check to see if the path exists before adding to dirsToSearch, since they are used as-is in later processing and may cause a 'not found' error.

src/code/GetHelper.cs
PaulHigin: I think this is just if the path leaf node is 'scripts', correct?  What if other parts of the path contains the word 'scripts'?  We may have to parse the path here (e.g., use GetFileName() again).  Otherwise instead of ToLower and Contains, you can use `pkgPath.IndexOf('scripts', StringComparison.OrdinalIgnoreCase) > -1` here.

src/code/GetHelper.cs
```diff
@@ -0,0 +1,389 @@
...
+                }
+            }
+
+            List<string> installedPkgsToReturn = new List<string>();
+
+            //2) use above list to check 
+            // if the version specificed is a version range
+            if (versionRange != null)
+            {
+                foreach (var pkgPath in dirsToSearch)
```
PaulHigin: These are not really package paths yet are they?  They are directories in which packages (modules, scripts) may exist in their own sub directories, correct?  Or maybe I missed something above.  Variable names are important for complex code like this, and comments should make steps and expected (path) state clear.

src/code/GetHelper.cs
PaulHigin: I see now the `GetDirectories` is called so these are actually individual 'package' or 'resource' sub directories.  What if modulePath does not exist, or there is an access denied error?  How are the errors handled?

src/code/GetHelper.cs
```diff
@@ -0,0 +1,389 @@
...
+
+
+            // Flatten returned pkgs before displaying output returnedPkgsFound.Flatten().ToList()[0]
+            var flattenedPkgs = installedPkgsToReturn.Flatten();
+
+
+
+            List<PSObject> foundInstalledPkgs = new List<PSObject>();
+
+            foreach (string xmlFilePath in flattenedPkgs)
```
PaulHigin: I feel each of these sections should go into their own private methods.  This helps define and clarify the individual processing steps that are performed.

1. Gather search directories
2. Gather individual package or resource directories
3. Filter package or resource directories based on user input
4. Create package object for each found resource directory

src/code/GetHelper.cs
```diff
@@ -0,0 +1,389 @@
...
+                // Open xml and read metadata from it     
+                if (File.Exists(xmlFilePath))
+                {
+                    ReadOnlyPSMemberInfoCollection<PSPropertyInfo> nameInfo;
+                    ReadOnlyPSMemberInfoCollection<PSPropertyInfo> versionInfo;
+                    ReadOnlyPSMemberInfoCollection<PSPropertyInfo> typeInfo;
+                    ReadOnlyPSMemberInfoCollection<PSPropertyInfo> descriptionInfo;
+                    ReadOnlyPSMemberInfoCollection<PSPropertyInfo> authorInfo;
+                    ReadOnlyPSMemberInfoCollection<PSPropertyInfo> companyNameInfo;
+                    ReadOnlyPSMemberInfoCollection<PSPropertyInfo> copyrightInfo;
```
PaulHigin: Weren't we going to create a concrete class for this, instead of creating a PowerShell custom object?  It would make creating the package info object easier and easier to use down the line.

src/code/GetHelper.cs
```diff
@@ -0,0 +1,389 @@
...
+
+
+                    //var isPrelease = false;
+                    using (StreamReader sr = new StreamReader(xmlFilePath))
+                    {
+
+                        string text = sr.ReadToEnd();
+                        var deserializedObj = (PSObject)PSSerializer.Deserialize(text);
+
+                        nameInfo = deserializedObj.Properties.Match("Name");
```
PaulHigin: Will this throw if the match fails?  If not then we are good, but we don't want to terminate the loop processing for one bad package xml.  Same goes for the Deserialize() method, we may want to try/catch rather than let an obscure error surface.

PaulHigin (changes requested):

src/PSModule.psm1
alerickson: Note:  remove targetting net472 (or any .net framework target0

src/PowerShellGet.psd1
alerickson: Remove Role Capabilities

src/code/GetHelper.cs
alerickson: Yup, remove the #ifdefs

src/code/GetHelper.cs
alerickson: We'll think about this and where to use debug vs verbose

src/code/GetHelper.cs
alerickson: I'll add the type here

src/code/GetHelper.cs
alerickson: I need to look into error handling here

src/code/GetHelper.cs
alerickson: Make sure to check if the paths in modulePaths exist before adding it in

src/code/GetHelper.cs
alerickson: Add a comment on the functionality here (filtering)

src/code/GetHelper.cs
alerickson: > Also what if version is non-null but doesn't parse to a valid version or version range? Do we ignore, emit a warning, or terminate with error?

This needs to be addressed... I believe this should be throwing a non-terminating error

src/code/GetHelper.cs
alerickson: Create a Util method to figure out if path "is a script" path.

src/code/GetHelper.cs
alerickson: This still needs to be implemented

srcOld/code/GetHelper.cs
```diff
@@ -1,389 +1,389 @@
...
-                    
-                }
-            }
-
-            // return here 
-            return foundInstalledPkgs;
-        }
-    }
-}
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
alerickson, line 1: Please ignore this file.  There are no changes here and functioning GetHelper code is in src/code/GetHelper.cs

test/GetInstalledResource.Tests.ps1
```diff
@@ -0,0 +1,103 @@
...
+
+Import-Module "$psscriptroot\PSGetTestUtils.psm1" -Force
+
+Describe 'Test Get-InstalledPSResource for Module' {
+
+    BeforeAll{
+        $TestGalleryName = Get-PoshTestGalleryName
+        Get-NewPSResourceRepositoryFile
+
+        Register-PSRepository $TestGalleryName -SourceLocation "https://www.poshtestgallery.com/api/v2"
```
anamnavi, line 13: There's a helper method in PSGetTestUtils.psm1 that returns the PoshTestGallery URL so if you add this to the BeforeAll: `$PoshTestGalleryURL = Get-PoshTestGalleryLocation` then you can use the `$PoshTestGalleryURL` variable for the `-SourceLocation` value.

test/GetInstalledResource.Tests.ps1
```diff
@@ -0,0 +1,103 @@
...
+    # Purpose: Get resource when given Name, Version param not null
+    # Action: Get-InstalledPSResource -Name ContosoServer -Version
+    # Expected Result: Returns ContosoServer resource
+    It "find resource when given Name to <Reason> <Version>" -TestCases @(
+        @{Name="*ShellG*";    ExpectedName="PowerShellGet"; Reason="validate name, with wildcard at beginning and end of name: *ShellG*"},
+        @{Name="PowerShell*"; ExpectedName="PowerShellGet"; Reason="validate name, with wildcard at end of name: PowerShellG*"},
+        @{Name="*ShellGet";   ExpectedName="PowerShellGet"; Reason="validate name, with wildcard at beginning of name: *ShellGet"},
+        @{Name="Power*Get";   ExpectedName="PowerShellGet"; Reason="validate name, with wildcard in middle of name: Power*Get"}
+    ) {
+        param($Version, $ExpectedVersion)
```
anamnavi, line 42: should be `param($Name, $ExpectedName)` here

test/GetInstalledResource.Tests.ps1
```diff
@@ -0,0 +1,103 @@
...
+    # Expected Result: Should get ContosoServer resource
+    It "Get specific module resource by name" {
+        $pkg = Get-InstalledPSResource -Name ContosoServer
+        $pkg.Name | Should -Be "ContosoServer"
+    }
+
+    # Purpose: Get resource when given Name, Version param not null
+    # Action: Get-InstalledPSResource -Name ContosoServer -Version
+    # Expected Result: Returns ContosoServer resource
+    It "find resource when given Name to <Reason> <Version>" -TestCases @(
```
anamnavi: remove `<Version>` here as this variable doesn't exist in your hashtables

test/GetInstalledResource.Tests.ps1
```diff
@@ -0,0 +1,103 @@
...
+        @{Name="*ShellG*";    ExpectedName="PowerShellGet"; Reason="validate name, with wildcard at beginning and end of name: *ShellG*"},
+        @{Name="PowerShell*"; ExpectedName="PowerShellGet"; Reason="validate name, with wildcard at end of name: PowerShellG*"},
+        @{Name="*ShellGet";   ExpectedName="PowerShellGet"; Reason="validate name, with wildcard at beginning of name: *ShellGet"},
+        @{Name="Power*Get";   ExpectedName="PowerShellGet"; Reason="validate name, with wildcard in middle of name: Power*Get"}
+    ) {
+        param($Version, $ExpectedVersion)
+        $pkgs = Get-InstalledPSResource -Name $Name
+        $pkgs.Name | Should -Be "PowerShellGet"
```
anamnavi, line 44: use `$ExpectedName` instead of `"PowerShellGet"` here.

test/GetInstalledResource.Tests.ps1
```diff
@@ -0,0 +1,103 @@
...
+    }
+
+    # Purpose: Get installed resources when given Name, and Version is '*'
+    # Action: Get-InstalledPSResource -Name ContosoServer -Version "*"
+    # Expected Result: returns 4 ContosoServer resources (of all versions in descending order)
+    It "Get resources when given Name,  and Version is '*'" {
+        $pkgs = Get-InstalledPSResource -Name Carbon -Version "*"
+        $pkgs.Count | Should -BeGreaterOrEqual 2
+    }
+}
```
anamnavi: add newline

src/code/GetHelper.cs
```diff
@@ -0,0 +1,636 @@
...
+                Array.ForEach(name, n => scriptXMLnames.Add((n + "_InstalledScriptInfo.xml").ToLower()));
+                char[] fileNameDelimiter = new char[] { '_' };
+
+                List<string> wildCardDirsToSearch = new List<string>();
+                foreach (var n in name)
+                {
+                    if (n.Contains("*"))
+                    {
+                        char[] wildcardDelimiter = new char[] { '*' };
+                        var tokenizedName = n.Split(wildcardDelimiter, StringSplitOptions.RemoveEmptyEntries);
```
anamnavi: I'll reach out about this offline, but might be better to use wildcardpattern.IsMatch() on names

src/code/GetHelper.cs
```diff
@@ -0,0 +1,636 @@
...
+                            cmdletPassedIn.WriteError(ErrorParsingName);
+                        }
+
+                        // We need to check versions specifically for scripts...
+                        try
+                        {
+                            // Since we can only determine a script version from the metadata file or the script itself,
+                            // we now need to validate that if a version parameter was passed into the cmdlet, the version of this script falls within that version range
+                            // If version range is not null and the xmlFilePath is for a script
+                            if (versionRange != null && xmlFilePath.EndsWith("_InstalledScriptInfo.xml"))
```
anamnavi: based on L105 are the xml file paths that are used here been converted to lowercase? (Just wondering in case that means EndsWith() condition is never hit.

src/code/GetHelper.cs
```diff
@@ -0,0 +1,636 @@
...
+                        }
+                        
+                        try
+                        {
+                            LanguagePrimitives.TryConvertTo(
+                                (deserializedObj.Properties["Author"] != null ? deserializedObj.Properties["Author"].Value?.ToString() : String.Empty), out string authorProp);
+                            pkgAsPSObject.Author = authorProp;
+                        }
+                        catch (Exception e)
+                        {
```
anamnavi: I agree with WriteError() for not being able to be parse non-mandatory parameters which weren't provided (instead of throwing a terminating error). However, maybe we can have 1 try catch where you try to parse all these optional fields and in the catch it writes a general parsing error. Or have a helper where this deserialized object is being parsed. Just some ideas :)

src/code/GetInstalledResource.cs
```diff
@@ -0,0 +1,51 @@
...
+using System.Management.Automation;
+using System.Threading;
+
+namespace Microsoft.PowerShell.PowerShellGet.Cmdlets
+{
+    /// <summary>
+    /// It retrieves a resource that was installEd with Install-PSResource
+    /// Returns a single resource or multiple resource.
+    /// </summary>
+    [Cmdlet(VerbsCommon.Get, "InstalledPSResource", HelpUri = "<add>", RemotingCapability = RemotingCapability.None)]
```
anamnavi: can remove RemotingCapability

src/code/GetHelper.cs
alerickson: > Can this throw? We should use TryParse if available. 

No, but I've updated this to use TryParse now.

> Also what if version is non-null but doesn't parse to a valid version or version range? Do we ignore, emit a warning, or terminate with error?

Currently it's ignored, but I've updated it to throw a non-terminating error

src/code/GetHelper.cs
```diff
@@ -0,0 +1,697 @@
...
+                foreach (string modulePath in modulePaths)
+                {
+                    cmdletPassedIn.WriteDebug(string.Format("Retrieving directories in the '{0}' module path", modulePath));
+                    try
+                    {
+                        dirsToSearch.AddRange(Directory.GetDirectories(modulePath).ToList());
+                    }
+                    catch (Exception e)
+                    {
+                        cmdletPassedIn.WriteVerbose(string.Format("Error retrieving directories from '{0}': '{1)'", modulePath, e.Message));
```
anamnavi: typo in error string, it should be: `{1}`

src/code/GetHelper.cs
```diff
@@ -0,0 +1,697 @@
...
+                                (new DirectoryInfo(p).Name.ToLower().StartsWith(tokenizedName[0])
+                                && new DirectoryInfo(p).Name.ToLower().EndsWith(tokenizedName[1]))));
+                            wildCardDirsToSearch.AddRange(dirsToSearch.FindAll(p =>
+                                (System.IO.Path.GetFileName(p).Split(fileNameDelimiter, StringSplitOptions.RemoveEmptyEntries))[0].ToLower().StartsWith(tokenizedName[0])
+                                && (System.IO.Path.GetFileName(p).Split(fileNameDelimiter, StringSplitOptions.RemoveEmptyEntries))[0].ToLower().EndsWith(tokenizedName[0])));
+                        }
+                    }
+                    else
+                    {
+                        wildCardDirsToSearch.AddRange(dirsToSearch.FindAll(p => (new DirectoryInfo(p).Name.ToLower().Equals(n))));
```
anamnavi: let's try to use wildcardpattern to do <strike>this entire thing</strike> just the `if` condition code...the issue with tokenized names is that it doesn't account for all cases with more than 2 tokens: i.e `pow*rsh*llget`

src/code/GetHelper.cs
```diff
@@ -0,0 +1,697 @@
...
+            // try to parse into a specific NuGet version
+           versionRange = null;
+            if (version != null)
+            {
+                NuGetVersion.TryParse(version, out NuGetVersion specificVersion);
+
+                if (specificVersion != null)
+                {
+                    // check if exact version
+                    versionRange = new VersionRange(specificVersion, true, specificVersion, true, null, null);
```
anamnavi: do you mean to pass in the same argument (`specificVersion`) to the constructor twice?

src/code/GetHelper.cs
```diff
@@ -0,0 +1,697 @@
...
+                        }
+                        catch (Exception e)
+                        {
+                            try
+                            {
+                                LanguagePrimitives.TryConvertTo(
+                                    (deserializedObj.Properties["PublishedDate"] != null ? (DateTime?)deserializedObj.Properties["PublishedDate"].Value : null), out DateTime? publishedDateProp);
+                                pkgAsPSObject.PublishedDate = publishedDateProp;
+                            }
+                            catch (Exception e2)
```
anamnavi: just to have a better understanding, why are you doing 2 parsings (in 2 try blocks) here?

src/code/GetHelper.cs
```diff
@@ -0,0 +1,697 @@
...
+                string psModulePath = Environment.GetEnvironmentVariable("PSModulePath");
+                string[] modulePaths = psModulePath.Split(';');
+
+                var PSVersion6 = new Version(6, 0);
+                var isCorePS = cmdletPassedIn.Host.Version >= PSVersion6;
+
+                if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+                {
+                    // If PowerShell 6+
+                    if (isCorePS)
```
anamnavi: not necessary but just an idea: 

instead of the `if (isCorePS)` `else` you could do:
`string powershellType = isCorePS ? "PowerShell" : "WindowsPowerShell"`
and then have just one assignment of myDocumentsPath and programFilesPath here like so:
` myDocumentsPath = Path.Combine(Environment.GetFolderPath(SpecialFolder.MyDocuments), powerShellType);`
` programFilesPath = Path.Combine(Environment.GetFolderPath(SpecialFolder.ProgramFiles), powerShellType);`

instead of separate similar assignments in the if/else. If you feel like it's clearer to have the if/else feel free to keep it as it is :) just an idea.

anamnavi (changes requested):
Looks great! I think GetHelper.cs is looking more intuitive and easier to follow. I left some questions and comments. I also wanted to ask is there a reason ` srcOld/code/GetHelper.cs ` is included in the PR?

Requesting change because of the typo comment.

src/code/GetHelper.cs
anamnavi: let me write out the code for this suggestion one min

src/code/GetHelper.cs
```diff
@@ -0,0 +1,697 @@
...
+            if (name != null && !name[0].Equals("*"))
+            {
+
+                List<string> scriptXMLnames = new List<string>();
+                Array.ForEach(name, n => scriptXMLnames.Add((n + "_InstalledScriptInfo.xml").ToLower()));
+                char[] fileNameDelimiter = new char[] { '_' };
+
+                foreach (var n in name)
+                {
+                    if (n.Contains("*"))
```
anamnavi: replace the code starting on L 152 til the end of the foreach with:

`WildcardPattern nameWildCardPattern = new WildcardPattern(n, WildcardOptions.IgnoreCase);`

`wildCardDirsToSearch.AddRange(dirsToSearch.FindAll(p => nameWildCardPattern.IsMatch(new DirectoryInfo(p).Name.ToLower())));`

` wildCardDirsToSearch.AddRange(dirsToSearch.FindAll(p => nameWildCardPattern.IsMatch(System.IO.Path.GetFileName(p))));`

src/code/GetHelper.cs
anamnavi: I put the code above for the `if (name contains *)` condition. The same wildcard pattern code (where I check if each name has a match) may not work here in the `else` condition.

Another thing: I'm not sure why you're calling `ToLower()` on the names...if it was for the tokenized names code and you can use the wildcard pattern code instead remove ToLower() as the wildcard pattern will ignore case if we specify it to. Then over here on L 193 you could remove the `Name.ToLower()` and just do `Name.Equals()`

src/PSModule.psm1
```diff
@@ -5,24 +5,14 @@ Set-StrictMode -Version Latest
...
-if ($isCore)
-{
-    $script:Framework = 'netstandard2.0'
-    $script:FrameworkToRemove = 'net472'
-    
-} else {
-    $script:Framework = 'net472'
-    $script:FrameworkToRemove = 'netstandard2.0'
-}
+$script:Framework = 'netstandard2.0'
```
PaulHigin: I think we want to standardize on 'net461' framework.  But we can discuss.

src/code/GetHelper.cs
```diff
@@ -0,0 +1,177 @@
...
+using NuGet.Versioning;
+
+namespace Microsoft.PowerShell.PowerShellGet.Cmdlets
+{
+    /// <summary>
+    /// Get helper class provides the core functionality for Get-InstalledPSResource.
+    /// </summary>
+    internal class GetHelper
+    {
+        private CancellationToken cancellationToken;
```
PaulHigin: Since this is only assigned in the constructor, it can also be 'readonly'.
But this doesn't appear to be used anywhere in this class, but maybe will be?

src/code/GetHelper.cs
```diff
@@ -0,0 +1,177 @@
...
+using Microsoft.PowerShell.PowerShellGet.UtilClasses;
+using static Microsoft.PowerShell.PowerShellGet.UtilClasses.PSResourceInfo;
+using NuGet.Versioning;
+
+namespace Microsoft.PowerShell.PowerShellGet.Cmdlets
+{
+    /// <summary>
+    /// Get helper class provides the core functionality for Get-InstalledPSResource.
+    /// </summary>
+    internal class GetHelper
```
PaulHigin: Nitpick: Maybe named 'GetUtils'?

src/code/GetHelper.cs
```diff
@@ -0,0 +1,177 @@
...
+        private CancellationToken cancellationToken;
+        private readonly PSCmdlet cmdletPassedIn;
+
+        public GetHelper(CancellationToken cancellationToken, PSCmdlet cmdletPassedIn)
+        {
+            this.cancellationToken = cancellationToken;
+            this.cmdletPassedIn = cmdletPassedIn;
+        }
+
+        public IEnumerable<PSResourceInfo> ProcessGetParams(string[] name, VersionRange versionRange, List<string>pathsToSearch)
```
PaulHigin: Name: 'GetPackages'?

src/code/GetHelper.cs
```diff
@@ -0,0 +1,177 @@
...
+            filteredPathsToSearch = GetResourceMetadataFiles(versionRange, filteredPathsToSearch);
+
+            foreach (PSResourceInfo pkgObject in OutputPackageObject(filteredPathsToSearch, versionRange))
+            {
+                yield return pkgObject;
+            }
+        }
+
+        // Filter packages by user provided name
+        public List<string> FilterPkgsByName(string[] names, List<string> dirsToSearch)
```
PaulHigin: Should this be 'private' access or is it used outside the class?
Also maybe this should be called 'FilterPkgPathsByName' since it returns directory paths and not packages.

src/code/GetHelper.cs
```diff
@@ -0,0 +1,177 @@
...
+            if (names == null)
+            {
+                names = new string[] { "*" };
+            }
+
+            foreach (string n in names)
+            {
+                WildcardPattern nameWildCardPattern = new WildcardPattern(n, WildcardOptions.IgnoreCase);
+                
+                // ./Modules/Test-Module
```
PaulHigin: Are these comments still applicable?  It seems like we are not making a distinction in path type yet...

src/code/GetHelper.cs
```diff
@@ -0,0 +1,177 @@
...
+            }
+             cmdletPassedIn.WriteDebug(wildCardDirsToSearch.Any().ToString());
+
+            return wildCardDirsToSearch;
+        }
+
+        // Filter by user provided version
+        public List<string> GetResourceMetadataFiles(VersionRange versionRange, List<string> dirsToSearch)
+        {
+            List<string> installedPkgsToReturn = new List<string>();  // these are the xmls
```
PaulHigin: // these are the package xml file paths
Name: GetResourceMetadataFilePaths ?

src/code/GetHelper.cs
alerickson: I never replied to this comment, but this is where we create a version range from an exact version eg: [1.0.0,1.0.0]

PaulHigin (changes requested):
LGTM, as per our offline review.

However, I see some test failures so please resolve them.
Please resolve any valid Codacy errors.
Also, please close all open conversations.

Then I think we can merge this PR.

PR #353, opened 2021-03-23
Refactor Uninstall-PSResource

### LINKED ISSUE #366: Uninstall-Module -RequiredVersion doesn't recognize version-number variations that should be equivalent
It is possible to use shortened version numbers in a module manifest, such as `ModuleVersion = '1.4'` instead of `ModuleVersion = '1.4.0'` or `ModuleVersion = '1.4.0.0'`

While using variations works fine with `Install-Module`, it doesn't with `UninstallModule -RequiredVersion`

Other version-number parsing/matching code may be affected as well.

## Steps to reproduce

```powershell
# Note: v1.4 of the 'Native' module has `ModuleVersion = '1.4'` in its module manifest.

trap { $_ | Write-Error; continue }

# OK: Install-Module: Targeting 1.4 as '1.4.0'
{ Install-Module -ErrorAction Stop -Force Native -RequiredVersion 1.4.0  } | Should -Not -Throw

# !! BROKEN: Uninstall-Module: Targeting 1.4 as '1.4.0' does *not* work.
{ Uninstall-Module -ErrorAction Stop -Force Native -RequiredVersion 1.4.0  } | Should -Not -Throw

# Clean up; only using the version number exactly as specified in the module manifest works.
Uninstall-Module -Force Native -RequiredVersion 1.4
```

## Expected behavior

Both tests should succeed.

## Actual behavior

The second test fails.

```none
Expected no exception to be thrown, but an exception 
"No match was found for the specified search criteria and module names 'Native'." 
was thrown
```

## Environment data

```none
PowerShell Core 7.2.0-preview.5
```

alerickson:
# PR Summary
Refactoring the contents of Uninstall-PSResource

Fixes #366 #365 

TODOs:
- [x] help documentation
- [ ] completing tests (after install-psresource is complete)

## PR Checklist

    - [ ] None
    - **OR**
        - [ ] Issue filed: 
- **User-facing changes**
    - [ ] Not Applicable
    - **OR**
        - [ ] Issue filed: 
    - **OR**
- **Tooling**
    - **OR**


src/code/UninstallPSResource.cs
```diff
@@ -147,26 +106,22 @@ protected override void ProcessRecord()
...
             }
-#endif
-
-
-            foreach (var pkgName in _name)
+            foreach (var pkgName in Name)
             {
-                var successfullyUninstalledPkg = UninstallPkgHelper(pkgName, cancellationToken);
+                bool successfullyUninstalledPkg = UninstallPkgHelper(pkgName, cancellationToken);
                 if (successfullyUninstalledPkg)
```
anamnavi: you can just use the call to UninstallPkgHelper() in your if statement condition:

(excuse the weird indentations):

`if (UninstallPkgHelper(pkgName, cancellationToken)) {`
  `     WriteVerbose("Successfully uninstalled " + pkgName);`
`}`
`else {`
 `       ...`
`}`

src/code/UninstallPSResource.cs
```diff
@@ -240,74 +193,43 @@ private bool UninstallPkgHelper(string pkgName, CancellationToken cancellationTo
             // If we can't find the resource, just return
             if (!foundResourceObj)
```
anamnavi: was trying to leave a comment for L186 above where you set foundResourceObj: `var foundResourceObj = foundInMyDocuments || foundInProgramFiles || File.Exists(scriptPathMyDocuments) || File.Exists(scriptPathProgramFiles) ? true : false;`
you could do:
`bool foundResourceObj = (the condition);`

src/code/UninstallPSResource.cs
```diff
@@ -99,42 +54,46 @@ public SwitchParameter Force
...
+            if (Version != null)
+            {
+                NuGetVersion.TryParse(Version, out nugetVersion);
+                if (nugetVersion == null)
+                {
+                    VersionRange.TryParse(Version, out versionRange);
+                }
 
-            var consoleIsElevated = false;
+                if (nugetVersion == null && versionRange == null)
```
anamnavi: you could nest this under Line 69 above. So instead of checking again if `(nugetVersion == null)` alongside the second (`versionRange == null`) condition, after you TryParse on Line 69, versionRange should be set (or remain set to null). so right under that code (and inside the if statement of `nugetVersion == null`) you could have an if statement check for `versionRange == null`).

(excuse the weird indentations):

`if(nugetVersion == null)`
`{`
`     VersionRange.TryParse(Version, out versionRange);`
`     if(versionRange == null) {`
`          ThrowTerminatingError()`
`     }`      
`}`

src/code/UninstallPSResource.cs
```diff
@@ -240,74 +193,43 @@ private bool UninstallPkgHelper(string pkgName, CancellationToken cancellationTo
             // If we can't find the resource, just return
             if (!foundResourceObj)
             {
+                WriteDebug("Could not find the resource: ");
```
anamnavi: WriteDebug("Could not find the resource: " + pkgName);

src/code/UninstallPSResource.cs
```diff
@@ -357,59 +275,46 @@ private bool UninstallModuleHelper(string pkgName, string dirName, string[] vers
...
-                    if (pkgsWithRequiredModules.Any())
+                    if (pkgsWithRequiredModules.Any() && !Force)
                     {
                         var uniquePkgNames = pkgsWithRequiredModules.Select(p => p.Properties["Name"].Value).Distinct().ToArray();
-
                         var strUniquePkgNames = string.Join(",", uniquePkgNames);
 
-                        throw new System.ArgumentException(string.Format(CultureInfo.InvariantCulture, "Cannot uninstall {0}, the following package(s) take a dependency on this package: {1}", pkgName, strUniquePkgNames));
-
+                        throw new System.ArgumentException(string.Format(CultureInfo.InvariantCulture, "Cannot uninstall {0}, the following package(s) take a dependency on this package: {1}.  If you would still like to uninstall, rerun the command with -Force", pkgName, strUniquePkgNames));
```
anamnavi: is this exception wrapped and handled in the caller method? or is it supposed to terminate the cmdlet right here?

src/code/UninstallPSResource.cs
```diff
@@ -520,35 +401,23 @@ private bool UninstallScriptHelper(string pkgName, string scriptsPath, string fu
```
anamnavi: left a similar comment above, but is this a wrapped exception handled by the caller or do we expect the cmdlet to terminate over here? Perhaps in `ProcessRecord() we add a try/catch around when this method or (`UninstallModuleHelper()`) gets called.

test/UninstallPSResource.ps1
```diff
@@ -0,0 +1,188 @@
...
+    AfterAll {
+        Get-RevertPSResourceRepositoryFile
+    }
+
+    It "Uninstall a specific module by name" {
+        $pkg = Uninstall-PSResource -name Bicep 
+    }
+
+    It "Uninstall a list of modules by name" {
+        $pkg = Uninstall-PSResource -Name Benchpress, bogreader
```
anamnavi: Fine to use production modules for now, but at some point we may need to use our own test modules that we get from the Gallery, install and then uninstall.

test/UninstallPSResource.ps1
```diff
@@ -0,0 +1,188 @@
...
+
+    It "Uninstall script when given Name, Version  --> '*'" {
+        $res = Uninstall-PSResource -Name ADSQL -Version "*"
+        $res.Count | Should -BeGreaterOrEqual 4
+    }
+
+    It "Uninstall script with latest with -WhatIf" {
+        $res = Uninstall-PSResource -Name "ActiveDirectoryTools" -WhatIf
+    }
+}
```
anamnavi: nitpick, add new line

anamnavi (commented):
this PR just showed changes, so some of the pre-existing code was hard to comment on because Github just wouldn't let me add a comment there. So I'm going to make them here:
- add license info at start of file
- remove RemotingCapability tag on Cmdlet (Line 22)
- as for error handling, it appears that UninstallScriptHelper() and UninstallModuleHelper() throw errors so we could add try/catch around when they get called in ProcessRecord()

src/code/UninstallPSResource.cs
```diff
@@ -99,42 +54,46 @@ public SwitchParameter Force
...
-
-            if (nugetVersion == null)
-            {
-                VersionRange.TryParse(_version, out versionRange); 
-            }
+            cancellationToken = source.Token;
+
+            if (Version != null)
+            {
+                NuGetVersion.TryParse(Version, out nugetVersion);
```
PaulHigin: Just curious, does `NuGetVersion.TryParse` return false or do you always have to check the result to be null?  It seems like `TryParse` should return false if it is unable to parse the string as a NuGetVersion, and you can skip the null check.
Same question for `VersionRange.TryParse`.

Also I believe this is used in other cmdlets so there can be a static helper method in the Utils class to parse a version string.

src/code/UninstallPSResource.cs
```diff
@@ -99,42 +54,46 @@ public SwitchParameter Force
...
+                {
+                    var exMessage = String.Format("Argument for -Version parameter is not in the proper format.");
+                    var ex = new ArgumentException(exMessage);
+                    var IncorrectVersionFormat = new ErrorRecord(ex, "IncorrectVersionFormat", ErrorCategory.InvalidArgument, null);
+
+                    ThrowTerminatingError(IncorrectVersionFormat);
+                }
+            }
+            
+            var isXPlatformPS = this.Host.Version.Major.CompareTo("6") > 1 ? true : false;
```
PaulHigin: Don't need ternary operation here, and I feel this is more readable:
```csharp
private static psVersion6 = new Version(6, 0);
var isWindowsPS = Host.Version < psVersion6;
```

Note that we normally leave off `this.`
```

**Review by PaulHigin — CHANGES_REQUESTED** [review]

File: src/code/UninstallPSResource.cs [hunk 11]
```diff
@@ -1,554 +1,423 @@
...
+using System.Collections.ObjectModel;
+using static System.Environment;
+using System.Runtime.InteropServices;
+
+namespace Microsoft.PowerShell.PowerShellGet.Cmdlets
+{
+
+
+    /// <summary>
+    /// Uninstall
```
alerickson: Add summary

src/code/UninstallPSResource.cs
```diff
@@ -1,554 +1,423 @@
...
-            return successfullyUninstalledPkg;
-        }
-
-
-    }
-}
+using System;
+using System.Management.Automation;
+using System.Threading;
+using NuGet.Versioning;
```
alerickson, line 8: move this below 'System' assemblies

src/code/UninstallPSResource.cs
```diff
@@ -1,554 +1,423 @@
...
+                {
+                    var exMessage = String.Format("Argument for -Version parameter is not in the proper format.");
+                    var ex = new ArgumentException(exMessage);
+                    var IncorrectVersionFormat = new ErrorRecord(ex, "IncorrectVersionFormat", ErrorCategory.InvalidArgument, null);
+
+                    ThrowTerminatingError(IncorrectVersionFormat);
+                }
+            }
+            
+            var isXPlatformPS = this.Host.Version.Major.CompareTo("6") > 1 ? true : false;
```
PaulHigin: This can be simplified to

```csharp
private static PSVersion6 = new Version (6, 0);
var isCorePS = Host.Version >= PSVersion6;
```

src/code/UninstallPSResource.cs
```diff
@@ -1,554 +1,423 @@
...
+                    programFilesPath = Path.Combine(Environment.GetFolderPath(SpecialFolder.ProgramFiles), "WindowsPowerShell");
+                }
+            }
+            else
+            {
+                // Paths are the same for both Linux and MacOS
+                myDocumentsPath = Path.Combine(Environment.GetFolderPath(SpecialFolder.LocalApplicationData), "Powershell");
+                programFilesPath = Path.Combine("usr", "local", "share", "Powershell");
+
+                using (System.Management.Automation.PowerShell pwsh = System.Management.Automation.PowerShell.Create())
```
PaulHigin: It is not clear what this does, and `results` does not appear to be used.  Can you add a comment or remove if not needed?
Also, since this is a cmdlet, you can use `InvokeCommand.InvokeScript` to run script in the cmdlet context, instead of creating a new PowerShell/Runspace.

E.g.,
https://github.com/PowerShell/SecretManagement/blob/b1fbbd0e7f4c0e035e7a2d495472f8a2790fe04d/src/code/SecretManagement.cs#L195

src/code/UninstallPSResource.cs
```diff
@@ -1,554 +1,423 @@
...
+        public string Version { get; set; }
+        
+        /// <summary>
+        /// </summary>
+        [Parameter(ParameterSetName = "NameParameterSet")]
+        public SwitchParameter Force { get; set; }
+
+
+        public static readonly string OsPlatform = System.Runtime.InteropServices.RuntimeInformation.OSDescription;
+        private string programFilesPath;
```
PaulHigin: Non static class fields naming convention is to have a preceding underscore character.
```csharp
private string _programfilesPath;
private string _myDocumentsPath;
...
```

src/code/UninstallPSResource.cs
```diff
@@ -1,554 +1,423 @@
...
+                // Paths are the same for both Linux and MacOS
+                myDocumentsPath = Path.Combine(Environment.GetFolderPath(SpecialFolder.LocalApplicationData), "Powershell");
+                programFilesPath = Path.Combine("usr", "local", "share", "Powershell");
+
+                using (System.Management.Automation.PowerShell pwsh = System.Management.Automation.PowerShell.Create())
+                {
+                    var results = pwsh.AddCommand("id").AddParameter("u").Invoke();
+                }
+            }
+            foreach (var pkgName in Name)
```
PaulHigin: This is really the only code you need in the `ProcessRecord()` block since the `Name` property values can come from the pipeline.

src/code/UninstallPSResource.cs
```diff
@@ -1,554 +1,423 @@
...
+                else
+                {
+                    WriteVerbose("Did not successfully uninstall " + pkgName);
+                }
+            }
+        }
+
+        
+        /// Just uninstall module, not dependencies
+        private bool UninstallPkgHelper(string pkgName, CancellationToken cancellationToken)
```
PaulHigin: Since `cancellationToken` is a class field, it does not need to be passed into the method.

src/code/UninstallPSResource.cs
```diff
@@ -1,554 +1,423 @@
...
+                }
+                else
+                {
+                    WriteVerbose("Did not successfully uninstall " + pkgName);
+                }
+            }
+        }
+
+        
+        /// Just uninstall module, not dependencies
```
PaulHigin: 'Installs a single resource`

src/code/UninstallPSResource.cs
```diff
@@ -1,554 +1,423 @@
...
+        private bool UninstallPkgHelper(string pkgName, CancellationToken cancellationToken)
+        {
+            var successfullyUninstalled = false;
+
+            dirsToDelete = new List<string>();
```
PaulHigin: `dirsToDelete` should not be a class field or instantiated here.  It is only used in the `UninstallModuleHelper` method and so should be scoped to there only.

src/code/UninstallPSResource.cs
```diff
@@ -1,554 +1,423 @@
...
+            }
+            
+            return successfullyUninstalled;
+        }
+        
+
+        /* Uninstall Module */
+        private bool UninstallModuleHelper(string pkgName, string dirName, string[] versionDirs, string[] parentDirFiles, CancellationToken cancellationToken)
+        {
+            var successfullyUninstalledPkg = false;
```
PaulHigin, line 184:
```csharp
var dirsToDelete = new List<string>();
```

src/code/UninstallPSResource.cs
```diff
@@ -1,554 +1,423 @@
...
+            foreach (var pkgName in Name)
+            {
+                bool successfullyUninstalledPkg = UninstallPkgHelper(pkgName, cancellationToken);
+                if (successfullyUninstalledPkg)
+                {
+                    WriteVerbose("Successfully uninstalled " + pkgName);
+                }
+                else
+                {
+                    WriteVerbose("Did not successfully uninstall " + pkgName);
```
PaulHigin: We should write a non-terminating error for each package that was not successfully uninstalled.  WriteVerbose should continue to list each packages that was successfully installed.

Also we should include the package path (if available) along with the package name that was uninstalled.  It would probably be better to write the error/verbose message closer to where the uninstall operation occurs, or pass back error/success messages with more information.

src/code/UninstallPSResource.cs
```diff
@@ -1,554 +1,423 @@
...
+            var psModulesPathProgramFiles = Path.Combine(programFilesPath, "Modules");
+
+            var psScriptPathMyDocuments = Path.Combine(myDocumentsPath, "Scripts");
+            var psScriptsPathProgramFiles = Path.Combine(programFilesPath, "Scripts");
+
+
+            /* Modules */
+            // My Documents
+            var dirNameMyDocuments = Path.Combine(psModulesPathMyDocuments, pkgName);
+            var versionDirsMyDocuments = (Directory.Exists(dirNameMyDocuments)) ? Directory.GetDirectories(dirNameMyDocuments) : null;
```
PaulHigin: Existence of `dirNameMyDocuments` path should be checked only once.

src/code/UninstallPSResource.cs
```diff
@@ -1,554 +1,423 @@
...
+            var parentDirFilesMyDocuments = (Directory.Exists(dirNameMyDocuments)) ? Directory.GetFiles(dirNameMyDocuments) : null;
```
PaulHigin: `parentDirFilesMyDocuments` path appears to be identical to `versionDirsMyDocuments` path.  Please add comment with examples to help clarify the difference between the two paths.

src/code/UninstallPSResource.cs
```diff
@@ -1,554 +1,423 @@
...
+            
+            /* Scripts */
+            // My Documents
+            var scriptPathMyDocuments = Path.Combine(psScriptPathMyDocuments, pkgName + ".ps1");
+            // Program Files
+            var scriptPathProgramFiles = Path.Combine(psScriptsPathProgramFiles, pkgName + ".ps1");
+
+
+            var psModulesPathAllDirs = new List<string>();
+            if (Directory.Exists(psModulesPathMyDocuments))
```
PaulHigin: We should check for directory existence just once and do all relevant processing at the same time (including the versionDirsProgramFiles, etc. from above).

src/code/UninstallPSResource.cs
```diff
@@ -1,554 +1,423 @@
...
+            {
+                psModulesPathAllDirs.AddRange(Directory.GetDirectories(psModulesPathMyDocuments).ToList());
+            }
+            if (Directory.Exists(psModulesPathProgramFiles))
+            {
+                psModulesPathAllDirs.AddRange(Directory.GetDirectories(psModulesPathProgramFiles).ToList());
+            }
+
+            var psScriptsPathAllFiles = new List<string>();
+            if (Directory.Exists(psScriptPathMyDocuments))
```
PaulHigin: Does order matter?  If a module appears in both directories, do we uninstall both?

src/code/UninstallPSResource.cs
```diff
@@ -1,554 +1,423 @@
...
+                                    || File.Exists(scriptPathProgramFiles) ? true : false;
+            
+
+            var isScript = (File.Exists(scriptPathMyDocuments) || File.Exists(scriptPathProgramFiles)) ? true : false;
+
+
+            // If we can't find the resource, just return
+            if (!foundResourceObj)
+            {
+                WriteDebug("Could not find the resource: ");
```
PaulHigin: We should write a 'not found' non-terminating error.  We should differentiate between package not found and package not successfully uninstalled.

src/code/UninstallPSResource.cs
```diff
@@ -1,554 +1,423 @@
...
+            {
+                psScriptsPathAllFiles.AddRange(Directory.GetFiles(psScriptPathMyDocuments).ToList())
+            }
+            if (Directory.Exists(psModulesPathMyDocuments))
+            {
+                psScriptsPathAllFiles.AddRange(Directory.GetFiles(psModulesPathMyDocuments).ToList());
+            }
+
+
+            var foundInMyDocuments = (Directory.Exists(dirNameMyDocuments) && (versionDirsMyDocuments.Any() || parentDirFilesMyDocuments.Any())) || File.Exists(scriptPathMyDocuments); // check for scripts
```
PaulHigin: Don't need to check for dir existence again, can just check `versionDirsMyDocuments` and `parentDirFilesMyDocuments` for null and empty.

src/code/UninstallPSResource.cs
```diff
@@ -1,554 +1,423 @@
...
+            }
+
+            var psScriptsPathAllFiles = new List<string>();
+            if (Directory.Exists(psScriptPathMyDocuments))
+            {
+                psScriptsPathAllFiles.AddRange(Directory.GetFiles(psScriptPathMyDocuments).ToList())
+            }
+            if (Directory.Exists(psModulesPathMyDocuments))
+            {
+                psScriptsPathAllFiles.AddRange(Directory.GetFiles(psModulesPathMyDocuments).ToList());
```
PaulHigin: `psScriptsPathAllFiles` does not appear to be used anywhere.

src/code/UninstallPSResource.cs
```diff
@@ -1,554 +1,423 @@
...
+            }
+
+
+            var foundInMyDocuments = (Directory.Exists(dirNameMyDocuments) && (versionDirsMyDocuments.Any() || parentDirFilesMyDocuments.Any())) || File.Exists(scriptPathMyDocuments); // check for scripts
+            var foundInProgramFiles = (Directory.Exists(dirNameProgramFiles) && (versionDirsProgramFiles.Any() || parentDirFilesProgramFiles.Any())) || File.Exists(scriptPathProgramFiles);
+
+            // First check if module or script is installed by looking in the specified modules path and scripts path
+            var foundResourceObj = foundInMyDocuments || foundInProgramFiles || File.Exists(scriptPathMyDocuments)
```
PaulHigin: Redundant `scriptPathMyDocuments` existence check.

PaulHigin (changes requested):

src/code/UninstallPSResource.cs
alerickson: TODO:
- Create a method in util class.  
- Add comments
- Get rid of pwsh creation (consider using invoke script if use is necessary)

src/code/UninstallPSResource.cs
alerickson: Add previous code into BeginProcessing

src/code/UninstallPSResource.cs
alerickson: Consider adding pkg path to both success and failed paths.

src/code/UninstallPSResource.cs
alerickson: Put in some examples to clarify, what you expect to see and what you expect it to look like

src/code/UninstallPSResource.cs
alerickson: Add comments, particularly for sections

src/code/UninstallPSResource.cs
alerickson: Check this and add comments for how ordering occurs.

src/code/UninstallPSResource.cs
alerickson: Remove this section

src/code/UninstallPSResource.cs
alerickson: This where dirsToDelete should be moved to

src/code/UninstallPSResource.cs
alerickson: Results was being used to figure out if a user was running as a superuser, I'll remove this since it's a relic from when the code was checking for elevated privileges.

src/code/UninstallPSResource.cs
alerickson: It looks like `parentDirFilesMyDocuments` is not used anywhere anymore so I'm completely removing it

src/code/UninstallPSResource.cs
```diff
@@ -1,554 +1,413 @@
...
+
+            Utils.TryParseVersionOrVersionRange(Version, out NuGetVersion _nugetVersion, out VersionRange _versionRange, this);
+           
+            var PSVersion6 = new Version(6, 0);
+            var isCorePS = Host.Version >= PSVersion6;
+
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+            {
+                // If PowerShell 6+
+                if (isCorePS)
```
anamnavi: I left the same optional suggestion in Get-InstalledPSResource...

not necessary but just an idea:

instead of the if (isCorePS) else you could do:
string powershellType = isCorePS ? "PowerShell" : "WindowsPowerShell"
and then have just one assignment of myDocumentsPath and programFilesPath here like so:
myDocumentsPath = Path.Combine(Environment.GetFolderPath(SpecialFolder.MyDocuments), powerShellType);
programFilesPath = Path.Combine(Environment.GetFolderPath(SpecialFolder.ProgramFiles), powerShellType);

instead of separate similar assignments in the if/else. If you feel like it's clearer to have the if/else feel free to keep it as it is :) just an idea.

src/code/UninstallPSResource.cs
```diff
@@ -1,554 +1,413 @@
...
+                }
+            }
+        }
+
+
+
+        // Uninstalls a single resource
+        private bool UninstallPkgHelper(string pkgName)
+        {
+            var successfullyUninstalled = false;
```
anamnavi: could do `bool successfullyUninstalled;` for default initialization

src/code/UninstallPSResource.cs
```diff
@@ -1,554 +1,413 @@
...
+
+            List<string> dirsToDelete = new List<string>();
+
+            if (String.IsNullOrWhiteSpace(pkgName))
```
anamnavi: I thought `Name` parameter could not be null or empty. If there's an array of names provided for `Name` parameter and one of the elements in that array is null or whitespace I think we should write an error with a specific "could not uninstall pkg - null name not allowed"?

For this, you do one of two things:
1) perhaps UninstallPkgHelper() could take a `out string[] errorMsgs` parameter and then in ProcessRecord() where you check if UninstallPkgHelper() returned false you could print any additional error messages.

2) before you return from UninstallPkgHelper() on L111 you could write the specific error message there "Name provided is null or whitespace". So overall there would be 2 error messages printing (one in UninstallPkgHelper() and one in ProcessRecord()).

src/code/UninstallPSResource.cs
```diff
@@ -1,554 +1,413 @@
...
+            }
+
+            /* Scripts */
+            // My Documents
+            var scriptPathMyDocuments = Path.Combine(psScriptPathMyDocuments, pkgName + ".ps1");
+            // Program Files
+            var scriptPathProgramFiles = Path.Combine(psScriptsPathProgramFiles, pkgName + ".ps1");
+            
+            var psModulesPathAllDirs = new List<string>();
+            bool foundInMyDocuments = false;
```
anamnavi: for these variables of type bool you can just use the default initialization. So `bool foundInMyDocuments;`

src/code/UninstallPSResource.cs
```diff
@@ -1,554 +1,413 @@
...
+using System.Threading;
+using NuGet.Versioning;
+using Microsoft.PowerShell.PowerShellGet.UtilClasses;
+
+namespace Microsoft.PowerShell.PowerShellGet.Cmdlets
+{
+    /// <summary>   
+    /// Uninstall-PSResource uninstalls a package found in a module or script installation path.
+    /// </summary>
+
```
PaulHigin: Please remove unneeded whitespace and 'RemotingCapability'.

src/code/UninstallPSResource.cs
```diff
@@ -1,554 +1,413 @@
...
+        [Cmdlet(VerbsLifecycle.Uninstall, "PSResource", DefaultParameterSetName = "NameParameterSet", SupportsShouldProcess = true,
+    HelpUri = "<add>", RemotingCapability = RemotingCapability.None)]
+    public sealed
+    class UninstallPSResource : PSCmdlet
+    {
+        /// <summary>
+        /// Specifies the exact names of resources to uninstall.
+        /// A comma-separated list of module names is accepted. The resource name must match the resource name in the repository.
+        /// </summary>
+        [Parameter(Mandatory = true, Position = 0, ValueFromPipeline = true, ValueFromPipelineByPropertyName = true, ParameterSetName = "NameParameterSet")]
```
PaulHigin: Currently we only have one parameter set 'NameParameterSet', but eventually will have another that takes 'PSResource' objects in the pipeline, correct?  We should go ahead and add the other parameter set.

src/code/UninstallPSResource.cs
```diff
@@ -1,554 +1,413 @@
...
+        [Parameter(ParameterSetName = "NameParameterSet")]
+        [ValidateNotNullOrEmpty]
+        public string Version { get; set; }
+        
+        /// <summary>
+        /// </summary>
+        [Parameter(ParameterSetName = "NameParameterSet")]
+        public SwitchParameter Force { get; set; }
+
+
```
PaulHigin, line 48: Please add `#region` sections for fields and PSCmdlet method overrides.

src/code/UninstallPSResource.cs
```diff
@@ -1,554 +1,413 @@
...
+        NuGetVersion _nugetVersion;
+        VersionRange _versionRange;
+
+        protected override void BeginProcessing()
+        {
+            _source = new CancellationTokenSource();
+            _cancellationToken = _source.Token;
+
+            Utils.TryParseVersionOrVersionRange(Version, out NuGetVersion _nugetVersion, out VersionRange _versionRange, this);
```
PaulHigin: This utility method should not throw, since it returns a success boolean.  It is better to throw at this higher cmdlet level where it has more context to provide a good error message.  

```csharp
if (! Utils.TryParseVersionOrVersionRange(..., out string errorMsg))
{
    ThrowTerminatingError(new PSArgumentException('Unable to parse Version parameter with error: {0}, errorMsg));
}

File: src/code/UninstallPSResource.cs [hunk 38]
```diff
@@ -1,554 +1,413 @@
...
+
+
+
+        protected override void ProcessRecord()
+        {
+            foreach (var pkgName in Name)
+            {
+                if (!UninstallPkgHelper(pkgName))
+                { 
+                    WriteVerbose("Did not successfully uninstall " + pkgName);
```
PaulHigin: This is an error, since the user asked that a specific package be uninstalled, and that operation failed.  But it is not a command terminating error, so we just write the error and continue.  But the error message should also contain the reason that the operation failed (invalid filepath, access denied, corrupted package info, ...):
```csharp
if (!UninstallPackage(pkgName, out string errorMsg))
{
    WriteError(string.Format("Did not successfully uninstall package {0}, with error: {1}", pkgName, errorMsg);
}
```

src/code/UninstallPSResource.cs
PaulHigin: I think local variables have to be explicitly initialized (?)

src/code/UninstallPSResource.cs
PaulHigin: Agreed. At this level `pkgname` should be assumed to be valid (not null or whitespace), since it is private access method that we control.  However, this is a good example where assert can be used to validate the implied contract.
```csharp
System.Diagnostics.Debug.Assert(!String.IsNullOrWhiteSpace(pgkName), "pkgName cannot be null or empty.");
```

src/code/UninstallPSResource.cs
```diff
@@ -1,554 +1,413 @@
...
+            var successfullyUninstalled = false;
+
+            List<string> dirsToDelete = new List<string>();
+
+            if (String.IsNullOrWhiteSpace(pkgName))
+            {
+                return successfullyUninstalled;
+            }
+
+            var psModulesPathMyDocuments = Path.Combine(_myDocumentsPath, "Modules");
```
PaulHigin: It looks like these paths never change and so they can be created just once in 'BeginProcessing()' block, and these variables can be class fields.  The `_myDocumentsPath`, etc don't need to be fields.

src/code/UninstallPSResource.cs
```diff
@@ -1,554 +1,413 @@
...
+            var psModulesPathProgramFiles = Path.Combine(_programFilesPath, "Modules");
+
+            var psScriptPathMyDocuments = Path.Combine(_myDocumentsPath, "Scripts");
+            var psScriptsPathProgramFiles = Path.Combine(_programFilesPath, "Scripts");
+
+
```
PaulHigin, line 141: Please remove extra whitespace.

src/code/UninstallPSResource.cs
```diff
@@ -1,554 +1,413 @@
...
+            var dirNameMyDocuments = Path.Combine(psModulesPathMyDocuments, pkgName);
+            string[] versionDirsMyDocuments = null;
+            /// modules can either be TestModule/m
+            if (Directory.Exists(dirNameMyDocuments))
+            {
+                // returns all the version directories
+                // eg:  TestModule/0.0.1
+                versionDirsMyDocuments = Directory.GetDirectories(dirNameMyDocuments);
+            }
+            // Program Files
```
PaulHigin: Please add line here and make comments consistent.

src/code/UninstallPSResource.cs
```diff
@@ -1,554 +1,413 @@
...
+            var scriptPathProgramFiles = Path.Combine(psScriptsPathProgramFiles, pkgName + ".ps1");
+            
+            var psModulesPathAllDirs = new List<string>();
+            bool foundInMyDocuments = false;
+            bool foundInProgramFiles = false;
+            bool isScript = false;
+            if (Directory.Exists(psModulesPathMyDocuments))
+            {
+                psModulesPathAllDirs.AddRange(Directory.GetDirectories(psModulesPathMyDocuments).ToList());
```
PaulHigin: .ToList() is not needed here.

PaulHigin (changes requested):

src/code/UninstallPSResource.cs
alerickson: Yes, that should be added as well.

src/code/UninstallPSResource.cs
alerickson: I think the intent early on was to fail quietly because to my understanding that's the philosophy PS generally takes.  V2 does not throw an error if uninstall was unsuccessful so I'm apprehensive about adding this.  I think this would be a good thing to discuss though.

src/code/UninstallPSResource.cs
alerickson: I believe so too, but I'll double check.

src/code/UninstallPSResource.cs
```diff
@@ -1,554 +1,413 @@
...
+            {
+                psModulesPathAllDirs.AddRange(Directory.GetDirectories(psModulesPathProgramFiles).ToList());
+                foundInProgramFiles = versionDirsProgramFiles.Any();
+                if (File.Exists(scriptPathProgramFiles))
+                {
+                    isScript = true;
+                    foundInProgramFiles = true;
+                }
+            }
+            bool foundResourceObj = foundInMyDocuments || foundInProgramFiles;
```
PaulHigin: Variable not needed since it is just used below.

```csharp
if (!foundInMyDocuments && !foundInProgramFiles)
{
    WriteError ...
}
```

src/code/UninstallPSResource.cs
```diff
@@ -1,554 +1,413 @@
...
+            {
+                WriteError(new ErrorRecord(
+                    new PSInvalidOperationException(string.Format("Could not find the resource '{0}'", pkgName)),
+                    "ErrorRetrievingSpecifiedResource",
+                    ErrorCategory.ObjectNotFound,
+                    this));
+
+                return successfullyUninstalled;
+            }
+
```
PaulHigin: Please remove extra white space.

src/code/UninstallPSResource.cs
```diff
@@ -1,554 +1,413 @@
...
+            
+            if (!isScript)
```
PaulHigin: Why is it important to know if the resource is script or not here?  Same as to whether the resource was found in 'MyDocuments' or 'ProgramFiles'.  I think this section can be simplified to just collecting a list of paths to uninstall that match the package Name match.  The helper function can then examine each resource path and figure out how to uninstall it (whether it includes version path or is script).

src/code/UninstallPSResource.cs
```diff
@@ -1,554 +1,413 @@
...
+        }
+        
+
+        /* Uninstall Module */
+        private bool UninstallModuleHelper(string pkgName, string dirName, string[] versionDirs, List<string> dirsToDelete)
+        {
+            var successfullyUninstalledPkg = false;
+
+            // if the version specificed is a version range
+            if (_versionRange != null)
```
PaulHigin: I wonder if we can move the version range check up to the calling function.  The calling function determines which resource directories match the name and version range and creates a list of them, and this helper function then just needs to figure out how to uninstall each resource path in the list.

src/code/UninstallPSResource.cs
```diff
@@ -1,554 +1,413 @@
...
+            }
+            else
+            {
+                // if no version is specified, just delete the latest version
+                Array.Sort(versionDirs);
+
+                dirsToDelete.Add(versionDirs[versionDirs.Length - 1]);
+            }
+            
+            // if dirsToDelete is empty, meaning we didn't find any modules, it's possible it's a script
```
PaulHigin: Please add comments that illustrate what kinds of resource paths are expected and how they look and all variations that need to be accommodated.  Something like:
```csharp
// Expected resource paths to uninstall:
//    c:\users\user\documents\modules\MyModule1\1.0.0
//    c:\users\user\documents\modules\MyModule1\1.1.0
//    c:\users\user\documents\modules\MyModule1\1.2.0
//    c:\Program Files\PowerShell\Modules\MyModule2\2.0.0
//    c:\Program Files\PowerShell\Scripts\MyScript1
```

PaulHigin (changes requested):

src/code/UninstallPSResource.cs
alerickson: I also agree that this needs to be addressed some how.   I think addressing validation of the package name earlier on makes much more sense and that creating writing a debug statement would be a clear way of relaying that information while still continuing to uninstall the rest of the packages.

src/code/UninstallPSResource.cs
alerickson: That makes sense to me, I think that's feasible.

src/code/UninstallPSResource.cs
alerickson: Just took a look at that and I think it's possible to refactor and get rid of those variables.  The latter variables I believe are relics from when there was a check for elevated privileges.  I think what you're suggesting would work and certainly makes sense.

src/code/UninstallPSResource.cs
alerickson: Consider moving to begin processing

src/code/Utils.cs
```diff
@@ -39,6 +40,34 @@ public static string QuoteName(string name)
...
+                        versionRange = new VersionRange(nugetVersion, true, nugetVersion, true, null, null);
+
+                    }
+                    else
+                    {
+                        successfullyParsed = VersionRange.TryParse(Version, out versionRange);
+                    }
+                }
+            }
+            return successfullyParsed;
```
anamnavi: I think you need a closing bracket `}` after this line? You may also need to assign the out variable `allVersions` before exiting the method

PaulHigin (approved):
LGTM

Please resolve all open conversations.

Also please look at Codacy issues (some may not apply):
https://app.codacy.com/gh/PowerShell/PowerShellGet/pullRequest?prid=7144195

After that I feel we can merge these changes.

anamnavi (approved):
LGTM. Just one question..

What's the difference between `test/UninstallPSResource.ps1 ` and `test/UninstallPSResource.Tests.ps1 `? (The latter follows our test file naming convention, should we remove the first from this PR?)

Issue #281, opened 2020-10-08
Find-PSResource Type parameter is a string not enum


ThomasNieto:
This parameter is currently a string parameter with a `ValidateSet` attribute. Can this be changed to an enum for easier discovery and maintainability?

https://github.com/PowerShell/PowerShellGet/blob/f1ddcb3bc602c61f9b98f79b1ab02b00331da8f5/src/code/FindPSResource.cs#L64-L65

alerickson:
This will be resolved in Preview 11.

Issue #284, opened 2020-10-13
Find-PSResource -Prerelease does not return non-prerelease


SydneyhSmith:
![image](https://user-images.githubusercontent.com/43417619/95897758-e96a6a00-0d42-11eb-8070-2750853146e5.png)
I would expect this to turn modules with this tag that are also not prerelease

alerickson:
This will be resolved in Preview 11.

Issue #264, opened 2020-09-22
Find-PSResource should include PreRelease property just like Get-Module does


JustinGrote:
Right now if you do `find-psresource -prerelease`, there's no way to tell what Prerelease is available.
![image](https://user-images.githubusercontent.com/15258962/93914783-e0dec080-fcbb-11ea-92c7-49a004450e1a.png)

SydneyhSmith:
Thanks @JustinGrote we will address this when we add formatting in the upcoming release:

https://github.com/PowerShell/PowerShellGet/issues/212

ThomasNieto:
The problem with that solution is that its not a property that can be used in automation but added in the formatting system. To properly solve this the `Version` type should be changed to `[NuGetVersion]` that supports both a `[System.Version]` and `[SemanticVersion]` version schemes as described in #269.

alerickson:
This will be resolved with the release of Preview 11.

Issue #259, opened 2020-09-17
Find-PSResource Always Prints Warning for Nuget v3 Feeds


bormm:
Tried version: 3.0.0-beta10

Hi,

in case I registered a nuget v3 feed from Azure DevOps, Find-PSResource always prints
`"WARNING: This functionality is not yet implemented"`
even the remaining functionality is given and the available resources are returned.

SydneyhSmith:
Thanks for calling this out @bormm , looks like the issue is occurring here: 
https://github.com/PowerShell/PowerShellGet/blob/e562f86317bef8f9ae2d1e7547fb890525f7e799/src/FindPSResource.cs#L211

bormm:
I do not think its a good idea to write "This functionality is not yet implemented" to output from somewhere in the code. What this "this" means is completely unknown for the user.

alerickson:
This will be resolved with the release of Preview 11.

PR #363, opened 2021-04-24
Refactor Find-PSResource cmdlet

anamnavi:
This PR includes refactoring for the Find-PSResource cmdlet. This cmdlet searches for requested packages.

**Note**: 
- At the moment this PR just includes core Find-PSResource functionality (ResourceNameParameterSet).
- DscResourceNameParameterSet, CommandNameParameterSet, TagParameterSet, TypeParameterSet soon to be completed.
- The code is refactored into a FindHelper.cs class so that Publish-PSResource may also utilize it.
- Resolves issues #281 #284 #259 #247 #201 #109 #173 #212 #264 

# PR Summary

Summary for this PR:
- [x] refactored Find-PSResource.cs
- [x] added tests with production module in this PR
- [x] updated help and design docs for this cmdlet in this PR.
- [x] add formatting for PSResourceInfo objects
## PR Checklist

    - [ ] None
    - **OR**
        - [ ] Issue filed: 
- **User-facing changes**
    - [ ] Not Applicable
    - **OR**
        - [ ] Issue filed: 
    - **OR**
- **Tooling**
    - **OR**


src/code/Utils.cs
```diff
@@ -52,6 +54,36 @@ public static string[] GetStringArray(ArrayList list)
             return strArray;
         }
 
+        public static bool TryParseVersionOrVersionRange(string Version, out VersionRange versionRange, out bool allVersions, PSCmdlet cmdletPassedIn)
+        {
+            var successfullyParsed = false;
+            NuGetVersion nugetVersion = null;
+            versionRange = null;
+            allVersions = false;
+            if (Version != null)
```
PaulHigin: You could have an early out here, and there would be less indentation afterward (easier to read).
```csharp
if (Version == null) { return successfullyParsed; }
```

src/code/Utils.cs
```diff
@@ -52,6 +54,36 @@ public static string[] GetStringArray(ArrayList list)
...
+            {
+                if (Version.Trim().Equals("*"))
+                {
+                    allVersions = true;
+                    successfullyParsed = true;
```
PaulHigin: Same here, could early out.

src/code/Utils.cs
```diff
@@ -3,7 +3,9 @@
 
 using System;
 using System.Collections;
+using System.Management.Automation;
 using System.Management.Automation.Language;
+using NuGet.Versioning;
```
PaulHigin, line 13: using statements should be ordered alphabetically.

src/code/Utils.cs
```diff
@@ -52,6 +54,36 @@ public static string[] GetStringArray(ArrayList list)
             return strArray;
         }
 
+        public static bool TryParseVersionOrVersionRange(string Version, out VersionRange versionRange, out bool allVersions, PSCmdlet cmdletPassedIn)
```
PaulHigin: `string Version` parameter name should be lower case.  Also `cmdletPassedIn` is not used.

src/code/Utils.cs
```diff
@@ -52,6 +54,36 @@ public static string[] GetStringArray(ArrayList list)
...
+                {
+                    allVersions = true;
+                    successfullyParsed = true;
+                }
+                else
+                {
+                    successfullyParsed = NuGetVersion.TryParse(Version, out nugetVersion);
+                    if (successfullyParsed)
+                    {
+                        versionRange = new VersionRange(nugetVersion, true, nugetVersion, true, null, null);
```
PaulHigin: For long complex function parameters, it is good to use named parameters for readability.

src/code/Utils.cs
```diff
@@ -52,6 +54,36 @@ public static string[] GetStringArray(ArrayList list)
...
+                    }
+                    else
+                    {
+                        successfullyParsed = VersionRange.TryParse(Version, out versionRange);
+                    }
+                }
+            }
+            return successfullyParsed;
+        }
+
```
PaulHigin, line 221: Please consider something of this form:
```csharp
        public static bool TryParseVersionRange(
            string version,
            out VersionRange versionRange,
            out bool allVersions)
        {
            versionRange = null;
            allVersions = false;

            if (version == null) { return false; }

            if (version.Trim().Equals("*"))
            {
                allVersions = true;
                return true;
            }

            // Try to parse as NuGetVersion
            if (NugetVersion.TryParse(version, out NuGetVersion nugetVersion))
            {
                versionRange = new VersionRange(
                    minVersion: nugetVersion,
                    includeMinVersion: true,
                    maxVersion: nugetVersion,
                    includeMaxVersion: true,
                    floatRange: null,
                    originalString: null);
                return true;
            }

            // Try to parse as a Version range
            return VersionRange.TryParse(version, out versionRange);
        }
```

PaulHigin (changes requested):

src/code/PSResourceInfo.cs
```diff
@@ -279,6 +283,114 @@ public sealed class PSResourceInfo
             }
         }
 
+        public static bool TryParse(
```
PaulHigin: We normally don't use 'TryParse' here since metadata should always be readable, and 'ConvertFromMetadata' is probably better, or maybe 'TryConvert'.  We may even want a class constructor that takes a metadata argument.

I feel this method should simply convert metadata into a PSResourceInfo object.  If some properties are missing then they are just set to null or empty string.  We can let higher level layers decide if required information is missing or not.

src/code/PSResourceInfo.cs
```diff
@@ -279,6 +283,114 @@ public sealed class PSResourceInfo
...
+            IPackageSearchMetadata metadataToParse,
+            out PSResourceInfo psGetInfo,
+            out string errorMsg)
+        {
+            psGetInfo = null;
+            errorMsg = String.Empty;
+
+            if (metadataToParse == null)
```
PaulHigin, line 398: In this case, we may want to consider throwing an exception here instead of returning a boolean.  We just want to convert from one form to another and we expect the operation to succeed except for null argument.

src/code/PSResourceInfo.cs
```diff
@@ -279,6 +283,114 @@ public sealed class PSResourceInfo
...
+            errorMsg = String.Empty;
+
+            if (objToParse == null)
+            {
+                errorMsg = "TryParseCatalogEntry: Invalid CatalogEntry object. Object cannot be null.";
+                return false;
+            }
+            try
+            {
+                NuspecReader pkgNuspec = objToParse.GetNuspecAsync().GetAwaiter().GetResult();
```
PaulHigin: I guess I don't understand how the package metadata is structured.  Some properties are directly from the catalog entry, and others seem to require a second retrieval of a nuspec part.  Please include comments to illustrate and clarify.

src/code/PSResourceInfo.cs
```diff
@@ -335,6 +447,154 @@ private static T ConvertToType<T>(PSObject psObject)
...
+        }
+
+        private static Uri ParseMetadataLicenseUri(IPackageSearchMetadata pkg)
+        {
+            return pkg.LicenseUrl;
+        }
+
+        private static string ParseMetadataName(IPackageSearchMetadata pkg)
+        {
+            if (pkg.Identity != null)
```
PaulHigin:
```csharp
return pkg.Identity?.Id ?? string.Empty;
```

src/code/PSResourceInfo.cs
```diff
@@ -335,6 +447,154 @@ private static T ConvertToType<T>(PSObject psObject)
...
+            if (pkgPublishedDate.HasValue)
+            {
+                publishDate = pkgPublishedDate.Value.DateTime;
+            }
+            return publishDate;
+        }
+
+        private static string[] ParseMetadataTags(IPackageSearchMetadata pkg)
+        {
+            char[] delimeter = new char[]{' ', ','};
```
PaulHigin: This should be static, so an allocation isn't performed each method call.

src/code/PSResourceInfo.cs
```diff
@@ -335,6 +447,154 @@ private static T ConvertToType<T>(PSObject psObject)
...
+            return pkg.Tags.Split(delimeter, StringSplitOptions.RemoveEmptyEntries);
+        }
+
+        private static string ParseMetadataType(IPackageSearchMetadata pkg)
+        {
+            string[] tags = ParseMetadataTags(pkg);
+            foreach(string tag in tags)
+            {
+                if(String.Equals(tag, "PSScript", StringComparison.OrdinalIgnoreCase))
```
PaulHigin: Maybe have an internal enum instead of string?

PaulHigin (changes requested):

src/code/PSResourceInfo.cs
anamnavi: has changed since, but I renamed the TryParse method to TryConvert

src/code/PSResourceInfo.cs
anamnavi: this is no longer happening, I was using this to get information on the package for a CatalogEntry object (which we are no longer using).

PaulHigin (approved):
LGTM

alerickson (approved):

PR #408, opened 2021-07-08
Install refactor

alerickson:
# PR Summary
This is a refactor of the Install-PSResource cmdlet, used to install PSResources (modules and scripts) from an online or local repository to a machine.

## PR Checklist

    - [x] None
    - **OR**
        - [ ] Issue filed: 
- **User-facing changes**
    - [x] Not Applicable
    - **OR**
        - [ ] Issue filed: 
    - **OR**
- **Tooling**
    - **OR**


src/code/InstallPSResource.cs
```diff
@@ -0,0 +1,266 @@
+using Microsoft.PowerShell.PowerShellGet.UtilClasses;
```
PaulHigin: Please add copyright info here.

src/code/PSResourceInfo.cs
```diff
@@ -12,7 +12,8 @@
 using NuGet.Packaging;
 using NuGet.Protocol.Core.Types;
 using NuGet.Versioning;
-
+using System.Reflection;
```
PaulHigin: Please ensure the copyright comment info is at the top of the file.

azure-pipelines-2.yml
```diff
@@ -0,0 +1,167 @@
+# The name of the build that will be seen in mscodehub
```
PaulHigin: Please remove this file from the PR.

azure-pipelines-1.yml
```diff
@@ -0,0 +1,167 @@
+# The name of the build that will be seen in mscodehub
```
PaulHigin: Please remove this file from the PR.

azure-pipelines.yml
```diff
@@ -0,0 +1,167 @@
+# The name of the build that will be seen in mscodehub
```
PaulHigin: Please remove this file from the PR.

src/code/InstallHelper.cs
```diff
@@ -0,0 +1,683 @@
...
+        string _specifiedPath;
+        bool _asNupkg;
+        bool _includeXML;
+        List<string> _pathsToSearch;
+
+        public InstallHelper(bool update, bool save, CancellationToken cancellationToken, PSCmdlet cmdletPassedIn)
+        {
+            this.update = update;
+            this.save = save;
+            this.cancellationToken = cancellationToken;
```
PaulHigin: It looks like this field is not used and can be removed.

src/code/InstallHelper.cs
```diff
@@ -0,0 +1,683 @@
...
+            this.cmdletPassedIn = cmdletPassedIn;
```
PaulHigin: Field names should begin with underscore character.

src/code/InstallHelper.cs
```diff
@@ -0,0 +1,683 @@
...
+
+namespace Microsoft.PowerShell.PowerShellGet.Cmdlets
+{
+    /// <summary>
+    /// Install helper class
+    /// </summary>
+    internal class InstallHelper : PSCmdlet
+    {
+        private CancellationToken cancellationToken;
+        private readonly bool update;
```
PaulHigin: Recommend name: `_updatePackage`.
Also this field is not currently used, but I assume will be?

src/code/InstallHelper.cs
```diff
@@ -0,0 +1,683 @@
...
+        private readonly bool save;
```
PaulHigin: Recommend name: `_savePackage`.

src/code/InstallHelper.cs
```diff
@@ -0,0 +1,683 @@
...
+
+        public InstallHelper(bool update, bool save, CancellationToken cancellationToken, PSCmdlet cmdletPassedIn)
+        {
+            this.update = update;
+            this.save = save;
+            this.cancellationToken = cancellationToken;
+            this.cmdletPassedIn = cmdletPassedIn;
+        }
+
+        public void ProcessInstallParams(
```
PaulHigin: Recommend name: `InstallPackages`

PaulHigin (changes requested):

src/code/InstallHelper.cs
```diff
@@ -0,0 +1,690 @@
...
+            _force = force;
+            _trustRepository = trustRepository;
+            _noClobber = noClobber;
+            _credential = credential;
+            _specifiedPath = specifiedPath;
+            _asNupkg = asNupkg;
+            _includeXML = includeXML;
+            _pathsToInstallPkg = pathsToInstallPkg;
+
+            IEnumerable<PSResourceInfo> pkgsAlreadyInstalled = new List<PSResourceInfo>();
```
PaulHigin: This unused variable can be removed.

src/code/InstallHelper.cs
```diff
@@ -0,0 +1,690 @@
...
+                        // Prompt for installation of package from untrusted repository
+                        var message = string.Format(CultureInfo.InvariantCulture, queryInstallUntrustedPackage, repoName);
+                        sourceTrusted = cmdletPassedIn.ShouldContinue(message, repositoryIsNotTrusted, true, ref yesToAll, ref noToAll);
+                    }
+                }
+                else {
+                    sourceTrusted = true;
+                }
+
+                if (sourceTrusted || yesToAll)
```
PaulHigin, line 144: Could to early out here:
```c#
if (!sourceTrusted && !yesToAll)
{
    Write-Verbose("Skipping untrusted repository 'A'");
    return;
}
```
```

File: src/code/InstallHelper.cs [hunk 13]
```diff
@@ -0,0 +1,690 @@
...
+                {
+                    cmdletPassedIn.WriteDebug("Untrusted repository accepted as trusted source.");
+
+                    // If it can't find the pkg in one repository, it'll look for it in the next repo in the list
+                    var isLocalRepo = repo.Url.AbsoluteUri.StartsWith(Uri.UriSchemeFile + Uri.SchemeDelimiter);
```
```suggestion
                    var isLocalRepo = repo.Url.AbsoluteUri.StartsWith(Uri.UriSchemeFile + Uri.SchemeDelimiter, StringComparison.OrdinalIgnoreCase);

```

src/code/InstallHelper.cs
```diff
@@ -0,0 +1,690 @@
...
+                        name: packageNames,
+                        type: ResourceType.None,
+                        version: _versionRange != null ? _versionRange.OriginalString : null, 
+                        prerelease: _prerelease,
+                        tag: null,
+                        repository: new string[] { repoName },
+                        credential: credential,
+                        includeDependencies: true);
+
+                    // Deduplicate any packages
```
PaulHigin: I feel that `findHelper.FindByResourceName()` should never return duplicate packages.  Unless there is a scenario where we would want duplicate packages returned (?), in which case I think `FindByResourceName` should take a parameter argument to optionally return only unique results.

This also sorts in name/version order, but shouldn't we expect `FindByResourceName` to already do this?

Please add comments to indicate the expected form of the package list:
```c#
// PacakgeA Version 1.5
// PackageA Version 1.2
// PackageA Version 1.0
// PackageB Version 1.0
```

src/code/InstallHelper.cs
```diff
@@ -0,0 +1,690 @@
...
+                    pkgsFromRepoToInstall = pkgsFromRepoToInstall.GroupBy(
+                        m => new { m.Name, m.Version }).Select(
+                            group => group.First()).ToList();
+
+                    // Make sure only the latest version of a module gets installed
```
PaulHigin: Comment should indicate expected order of returned packages (as shown in my comment above), and this comment should clarify that this code is selecting the first of each package name group, which is guaranteed to be the latest version.

src/code/InstallHelper.cs
```diff
@@ -0,0 +1,690 @@
...
+                    }
+
+                    // Check to see if the pkgs (including dependencies) are already installed (ie the pkg is installed and the version satisfies the version range provided via param)
+                    if (!_reinstall)
+                    {
+                        // Removes all of the names that are already installed from the list of names to search for
+                        pkgsFromRepoToInstall = FilterByInstalledPkgs(pkgsFromRepoToInstall);
+                    }
+
+                    if (!pkgsFromRepoToInstall.Any()) continue;
```
PaulHigin: Style guidelines, we should always use statement brackets.

src/code/InstallHelper.cs
```diff
@@ -0,0 +1,690 @@
...
+
+                    List<string> pkgsInstalled = InstallPackage(pkgsFromRepoToInstall, repoName, repo.Url.AbsoluteUri, credential, isLocalRepo);
+
+                    foreach (string name in pkgsInstalled)
+                    {
+                        if (packagesToInstall.Contains(name))
```
PaulHigin: We don't need this check to see if the list contains an item here.  `packagesToInstall.Remove()` just returns a boolean and doesn't throw if item to remove does not exist.

src/code/InstallHelper.cs
```diff
@@ -0,0 +1,690 @@
...
+        // Check if any of the pkg versions are already installed, if they are we'll remove them from the list of packages to install
+        public IEnumerable<PSResourceInfo> FilterByInstalledPkgs(IEnumerable<PSResourceInfo> packagesToInstall)
+        {
+            List<string> pkgNames = new List<string>();
+            foreach (var pkg in packagesToInstall)
+            {
+                pkgNames.Add(pkg.Name);
+            }
+
+            _pathsToSearch = new List<string>();
```
PaulHigin: `_pathsToSearch` should just be a local variable and not a field variable, since it is only used in this local scope.

src/code/InstallHelper.cs
```diff
@@ -0,0 +1,690 @@
...
+            GetHelper getHelper = new GetHelper(cmdletPassedIn);
+            // _pathsToInstallPkg will only contain the paths specified within the -Scope param (if applicable)
+            foreach (var path in _pathsToInstallPkg)
```
PaulHigin, line 227:
```c#
// _pathsToSearch will contain all resource package subdirectories within _pathsToInstallPkg path locations
// e.g.
// ./InstallPackagePath1/PackageA
// ./InstallPackagePath1/PackageB
// ./InstallPackagePath2/PackageC
// ./InstallPackagePath3/PackageD
```

src/code/InstallHelper.cs
```diff
@@ -0,0 +1,690 @@
...
+                    // Create PackageIdentity in order to download
+                    string createFullVersion = p.Version.ToString();
+                    if (p.IsPrerelease)
+                    {
+                        createFullVersion = p.Version.ToString() + "-" + p.PrereleaseLabel;
+                    }
+
+                    if (!NuGetVersion.TryParse(createFullVersion, out NuGetVersion pkgVersion))
+                    {
+                        cmdletPassedIn.WriteDebug("Error parsing version into a NuGetVersion");
```
PaulHigin: I assume we cannot continue processing the package and need to go to the next in the list.

src/code/InstallHelper.cs
```diff
@@ -0,0 +1,690 @@
...
+                             token: CancellationToken.None).GetAwaiter().GetResult();
+                        
+                        if (_asNupkg) // this is Save functionality
+                        {
+                            DirectoryInfo nupkgPath = new DirectoryInfo(((System.IO.FileStream)result.PackageStream).Name);
+                            File.Copy(nupkgPath.FullName, Path.Combine(tempInstallPath, pkgIdentity.Id + pkgIdentity.Version + ".nupkg"));
+
+                            continue;
+                        }
+                        else
```
PaulHigin: Don't need else block here since if block uses continue.

src/code/InstallHelper.cs
```diff
@@ -0,0 +1,690 @@
...
+                                packageFiles: result.PackageReader.GetFiles(),
+                                extractFile: (new PackageFileExtractor(result.PackageReader.GetFiles(), packageExtractionContext.XmlDocFileSaveMode)).ExtractPackageFile,
+                                logger: NullLogger.Instance,
+                                token: CancellationToken.None);
+                            result.Dispose();
+                        }
+                    }
+                    else
+                    {
+                        // Set up NuGet API resource for download
```
PaulHigin, line 326: Please add comment here that makes it clear this section is for downloading from non-local repository.

src/code/InstallHelper.cs
```diff
@@ -0,0 +1,690 @@
...
+                        }
+                        finally
+                        {
+                            // Need to close the .nupkg
+                            if (result != null) result.Dispose();
+                        }
+
+                        if (_asNupkg)  // Save functionality
+                        {
+                            /*
```
PaulHigin: Why is this commented out?  We should try to not check in any code with commented out sections.

src/code/InstallHelper.cs
```diff
@@ -0,0 +1,690 @@
...
+                    DeleteExtraneousFiles(tempInstallPath, pkgIdentity, tempDirNameVersion);
+
+                    // PSModules: 
+                    /// ./Modules
+                    /// ./Scripts
+                    /// _pathsToInstallPkg is sorted by desirability, Find will pick the pick the first Script or Modules path found in the list
+                    var installPath = isScript ? _pathsToInstallPkg.Find(path => path.EndsWith("Scripts", StringComparison.InvariantCultureIgnoreCase))
+                            : _pathsToInstallPkg.Find(path => path.EndsWith("Modules", StringComparison.InvariantCultureIgnoreCase));
+
+                    if (_includeXML) CreateMetadataXMLFile(tempDirNameVersion, installPath, repoName, p, isScript);
```
PaulHigin: Style guidelines, we should always use statement blocks with brackets.

src/code/InstallHelper.cs
```diff
@@ -0,0 +1,690 @@
...
+                    
+                    if (save)
```
PaulHigin: How is this save different from above _asnupkg save?  If it is just the form in which it is saved, then the save should always be done using this field (should be named `_save`) and `_asnupkg` is used to save in the correct form.  A separate helper function may be good for this.

src/code/InstallHelper.cs
```diff
@@ -0,0 +1,690 @@
...
+                    {
+                        //TODO:  SavePackage
```
PaulHigin: It would be great to include this functionality in this PR submission, for completeness.

src/code/InstallHelper.cs
```diff
@@ -0,0 +1,690 @@
...
+            int activityId = 0;
+            string activity = "";
+            string statusDescription = "";
+
+            // If the pkg exists in one of the names passed in, then we wont include it as a dependent package
+            activityId = 0;
+            activity = string.Format("Installing {0}...", p.Name);
+            statusDescription = string.Format("{0}% Complete:", i++);
+
+           // j = 1;
```
PaulHigin: Please remove commented out code.  If progress bar support is not ready, we can just add a TODO: comment, and implement later.

src/code/InstallPSResource.cs
```diff
@@ -0,0 +1,266 @@
...
+                                var exMessage = "Argument for version parameter is not in the proper format.";
+                                var ex = new ArgumentException(exMessage);
+                                var InputObjIncorrectVersionFormat = new ErrorRecord(ex, "InputObjIncorrectVersionFormat", ErrorCategory.InvalidArgument, null);
+                                ThrowTerminatingError(InputObjIncorrectVersionFormat);
+                            }
+
+                            installHelper.ProcessInstallParams(new[] { pkg.Name }, inputObjVersionRange, prerelease, Repository, AcceptLicense, Quiet, Reinstall, force: false, TrustRepository, NoClobber, Credential, RequiredResourceFile, _requiredResourceJson, _requiredResourceHash, specifiedPath: null, asNupkg: false, includeXML: true, _pathsToInstallPkg);
+                        }
+                    }
+                    else if (InputObject[0].GetType().Name.Equals("PSModuleInfo"))
```
anamnavi, line 222: how does this condition differ from that of the `if` statement above?

PaulHigin (changes requested):

src/code/InstallPSResource.cs
```diff
@@ -0,0 +1,266 @@
...
+        /// </summary>
+        [Parameter(ParameterSetName = NameParameterSet)]
+        //[Parameter(ParameterSetName = RequiredResourceFileParameterSet)]
+        public SwitchParameter Prerelease { get; set; }
+
+        /// <summary>
+        /// Specifies a user account that has rights to find a resource from a specific repository.
+        /// </summary>
+        [Parameter(ParameterSetName = NameParameterSet)]
+        // todo: add tab completion (look at get-psresourcerepository at the name parameter)
```
anamnavi, line 56: `[ArgumentCompleter(typeof(RepositoryNameCompleter))]` this would be the line of code to do so

anamnavi: [ArgumentCompleter(typeof(RepositoryNameCompleter))]

src/code/InstallPSResource.cs
anamnavi: how does this condition differ from that of the if statement above? The code in the if block and else if block look different so perhaps you meant to have a different condition for one?

src/code/PSResourceInfo.cs
```diff
@@ -524,85 +533,98 @@ private static Dependency[] GetDependencies(ArrayList dependencyInfos)
...
-
-                    dependenciesFound.Add(new Dependency(dependencyName, dependencyVersionRange));
-                    continue;
-                }
-
-                // neither Required, Minimum or Maximum Version provided
-                VersionRange dependencyVersionRangeAll = VersionRange.All;
-                dependenciesFound.Add(new Dependency(dependencyName, dependencyVersionRangeAll));
+            {
+                // can be an array or hashtable
```
PaulHigin: In what cases is the dependency info object an array?  Please clarify comment about what kind of array and when this occurs, and what form the information is organized.

src/code/PSResourceInfo.cs
```diff
@@ -671,6 +693,11 @@ private static Uri ParseMetadataIconUri(IPackageSearchMetadata pkg)
             return pkg.IconUrl;
         }
 
+        private static bool ParseMetadataIsPrerelease(IPackageSearchMetadata pkg)
+        {
+            return pkg.Identity.Version.IsPrerelease;
```
PaulHigin:
```suggestion
            return pkg.Identity?.Version?.IsPrerelease ?? false;
```

src/code/PSResourceInfo.cs
```diff
@@ -774,11 +801,16 @@ private static Version ParseMetadataVersion(IPackageSearchMetadata pkg)
         private PSObject ConvertToCustomObject()
         {
             var additionalMetadata = new PSObject();
-            foreach (var item in AdditionalMetadata)
+
+            // Need to add a null check here due to null ref exception getting thrown
+            if (AdditionalMetadata != null)
```
PaulHigin: In what cases is this null?  Please clarify.

PaulHigin (changes requested):

src/code/InstallPSResource.cs
```diff
@@ -0,0 +1,266 @@
...
+                default:
+                    // TODO: throw some kind of terminating error (unrecognized parameter set)
+                    // TODO: change to debug assert
+                    WriteDebug("Invalid parameter set");
+                    break;
+            }
+        }
+        #endregion
+    }
+}
```
anamnavi: add newline (at end of all files). I'll just make the comment here though :)

src/code/PowerShellGet.csproj
```diff
@@ -9,13 +9,14 @@
...
     <TargetFrameworks>netstandard2.0</TargetFrameworks>
+    <LangVersion>7.2</LangVersion>
   </PropertyGroup>
 
   <ItemGroup>
     <PackageReference Include="LinqKit.Core" Version="1.1.17" />
     <PackageReference Include="Microsoft.Extensions.Logging" Version="5.0.0-preview.7.20364.11" />
     <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="5.0.0-preview.7.20364.11" />
-    <PackageReference Include="Microsoft.Windows.Compatibility" Version="3.1.0"  Condition="'$(TargetFramework)' == 'netstandard2.0'" />
+    <PackageReference Include="Microsoft.Windows.Compatibility" Version="3.1.0" Condition="'$(TargetFramework)' == 'netstandard2.0'" />
```
PaulHigin: Why is this needed?

src/code/PowerShellGet.csproj
```diff
@@ -9,13 +9,14 @@
     <FileVersion>3.0.0</FileVersion>
     <InformationalVersion>3.0.0</InformationalVersion>
     <TargetFrameworks>netstandard2.0</TargetFrameworks>
+    <LangVersion>7.2</LangVersion>
```
PaulHigin: Language == 8.0 ?

src/code/InstallPkgParams.cs
```diff
@@ -0,0 +1,17 @@
+using System.Management.Automation;
+
+public class PkgParams
```
PaulHigin: What is this for?  It doesn't seem to be used anywhere.

src/code/Utils.cs
```diff
@@ -1,240 +1,347 @@
...
+using System.Management.Automation;
+using System.Management.Automation.Language;
+using System.Runtime.InteropServices;
+using NuGet.Versioning;
+
+namespace Microsoft.PowerShell.PowerShellGet.UtilClasses
+{
+    internal static class Utils
+    {
+                public static void WriteVerboseOnCmdlet(
```
anamnavi: nitpick, but the indentation got weird here..

PaulHigin (changes requested):

src/code/Utils.cs
```diff
@@ -1,240 +1,347 @@
...
+            }
+            else
+            {
+                // expecting the full version module path
+                // ex:  ./PowerShell/Modules/TestModule/1.0.0
+                return new DirectoryInfo(pkgPath).Parent.Name;
+            }
+        }
+
+        public static List<string> GetAllResourcePaths(PSCmdlet psCmdlet)
```
anamnavi, line 201: if you could add a comment here saying `this method returns paths to all installed resources, for example: 'c:\users\johndoe\Documents\PowerShell\Modules\TestModule'` that would be great :) That comment helped me understand the difference between this helper and GetInstallationPaths.

src/code/Utils.cs
```diff
@@ -1,240 +1,347 @@
...
+            // ./PowerShell/Scripts/Test-Script.ps1
+            // ./PowerShell/Modules/TestModule
+            // need to use .ToList() to cast the IEnumerable<string> to type List<string>
+            pathsToSearch = pathsToSearch.Distinct(StringComparer.InvariantCultureIgnoreCase).ToList();
+            pathsToSearch.ForEach(dir => psCmdlet.WriteDebug(string.Format("All paths to search: '{0}'", dir)));
+
+            return pathsToSearch;
+        }
+
+        // Find all potential installation paths given a scope
```
anamnavi, line 274: `only return the parent module/script path, for example 'c:\users\americks\Documents\PowerShell\Modules'` that was how you explained this helper method in one of our chats and that really helped if that can be added here.

anamnavi (changes requested):
looks good, just some nitpicks and questions :)

src/code/InstallHelper.cs
anamnavi: We could do this in FindHelper.FindResourceName() the only reason I hadn't was because we'd have to get all packages first, sort for unique and then yield from that list.

Currently it yields each package as it "finds" it. I thought collecting all packages and sorting would be better done as far up as possible to retain some of the efficiency we get by yielding

src/code/InstallPkgParams.cs
alerickson: Yup, this is old and got added in.  Will remove this file as well.

src/code/PowerShellGet.csproj
alerickson: Don't believe it is anymore, I've removed it.

src/code/InstallHelper.cs
alerickson: Yes that's correct

src/code/InstallHelper.cs
alerickson: I've updated the class so that where FindHelper was previously passing in a new cancellation token, it's now passing in the Install cancellation token.  This token is now also getting passed into the NuGet API calls (GetDownloadResourceResultAsync and PackageReader.CopyFiles).

src/code/InstallHelper.cs
alerickson: I don't think the early out would actually do anything because if that condition is not satisfied it just moves onto the next iteration of the loop.

src/code/InstallHelper.cs
PaulHigin: This sounds reasonable to me.  Let's keep the code as is then.

src/code/InstallHelper.cs
PaulHigin: Oh yes, thanks.  We would need a `continue;` here.

src/code/InstallPSResource.cs
```diff
@@ -0,0 +1,212 @@
...
+        /// </summary>
+        [Parameter(ValueFromPipelineByPropertyName = true, ParameterSetName = NameParameterSet)]
+        [ValidateNotNullOrEmpty]
+        public string Version { get; set; }
+        
+        /// <summary>
+        /// Specifies to allow installation of prerelease versions
+        /// </summary>
+        [Parameter(ParameterSetName = NameParameterSet)]
+        //[Parameter(ParameterSetName = RequiredResourceFileParameterSet)]
```
PaulHigin: I think we should not have commented out parameter sets.  We should either not include the code or include the code and throw a `PSNotImplementedExpection`.

src/code/InstallPSResource.cs
```diff
@@ -0,0 +1,212 @@
...
+        List<string> _pathsToInstallPkg;
+        VersionRange _versionRange;
+        #endregion
+
+        #region Methods
+        protected override void BeginProcessing()
+        {
+            // validate that if a -Version param is passed in that it can be parsed into a NuGet version range. 
+            // An exact version will be formatted into a version range.
+            if (ParameterSetName.Equals("NameParameterSet") && Version != null && !Utils.TryParseVersionOrVersionRange(Version, out _versionRange))
```
```suggestion
            if (ParameterSetName.Equals(NameParameterSet) && Version != null && !Utils.TryParseVersionOrVersionRange(Version, out _versionRange))

```

src/code/InstallPSResource.cs
```diff
@@ -0,0 +1,212 @@
...
+                        requiredResourceJson: null, 
+                        requiredResourceHash: null, 
+                        specifiedPath: null, 
+                        asNupkg: false, 
+                        includeXML: true, 
+                        pathsToInstallPkg: _pathsToInstallPkg);
+                    break;
+                    
+                case RequiredResourceFileParameterSet:
+                    WriteDebug("Not yet implemented");
```
PaulHigin: We should throw a `PSNotImplementedException`.

PaulHigin (changes requested):

src/code/InstallPSResource.cs
alerickson: I'll take this out right now

src/code/InstallHelper.cs
alerickson: Added now :)

src/code/InstallPSResource.cs
alerickson: Removing all of this code for now, and will add in later when we discuss whether we want to pursue "ValueFromPipelineByPropertyName" vs "InputObject"

src/code/InstallHelper.cs
alerickson: Yes that's correct, I added a continue statement here

src/code/InstallHelper.cs
alerickson: I've removed this condition because it's not needed anymore

src/code/PSResourceInfo.cs
PaulHigin: I see now.  Please disregard.

PR #432, opened 2021-08-02
Wildcard processing

anamnavi:
# PR Summary

Each cmdlet which takes a Name parameter should handle wildcards based on how it supports it for that cmdlet. I removed the original Utils.FilterOutWildcardNames() helper method (which didn't handle `name '*'` for different cmdlet cases (it just treated it as unsupported). I used Utils.ProcessNameWildcards() instead, which handles `name "*"` case in each cmdlet. I updated Get-InstalledPSResource, Find-PSResource, Install-PSResource (which doesn't support wildcards but needed error handling against it) and their tests accordingly.

Cmdlets left to add support/error handling for wildcards to:
- Save-PSResource
- Uninstall-PSResource
- Register-PSResourceRepository and Set-PSResourceRepository (only checks for `*` right now)

## PR Checklist

    - [ ] None
    - **OR**
        - [ ] Issue filed: 
- **User-facing changes**
    - [ ] Not Applicable
    - **OR**
        - [ ] Issue filed: 
    - **OR**
- **Tooling**
    - **OR**

### DIFF
--- a/src/code/GetInstalledPSResource.cs
+++ b/src/code/GetInstalledPSResource.cs
@@ -112,7 +112,7 @@ namespace Microsoft.PowerShell.PowerShellGet.Cmdlets
         {
             WriteDebug("Entering GetInstalledPSResource");
 
-            Name = Utils.ProcessNameWildcards(Name, out string[] errorMsgs, out bool _);
+            var namesToSearch = Utils.ProcessNameWildcards(Name, out string[] errorMsgs, out bool _);
             foreach (string error in errorMsgs)
             {
                 WriteError(new ErrorRecord(
@@ -124,13 +124,13 @@ namespace Microsoft.PowerShell.PowerShellGet.Cmdlets
 
             // this catches the case where Name wasn't passed in as null or empty,
             // but after filtering out unsupported wildcard names in BeginProcessing() there are no elements left in Name
-            if (Name.Length == 0)
+            if (namesToSearch.Length == 0)
             {
                  return;
             }
 
             GetHelper getHelper = new GetHelper(this);
-            foreach (PSResourceInfo pkg in getHelper.FilterPkgPaths(Name, _versionRange, _pathsToSearch))
+            foreach (PSResourceInfo pkg in getHelper.FilterPkgPaths(namesToSearch, _versionRange, _pathsToSearch))
             {
                 WriteObject(pkg);
             }
--- a/src/code/GetInstalledPSResource.cs
+++ b/src/code/GetInstalledPSResource.cs
@@ -106,6 +106,11 @@ namespace Microsoft.PowerShell.PowerShellGet.Cmdlets
                 // retrieve all possible paths
                 _pathsToSearch = Utils.GetAllResourcePaths(this);
             }
+        }
+
+        protected override void ProcessRecord()
+        {
+            WriteDebug("Entering GetInstalledPSResource");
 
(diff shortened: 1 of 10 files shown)


src/code/GetInstalledPSResource.cs
```diff
@@ -106,31 +106,29 @@ protected override void BeginProcessing()
...
-                    }
-                }
-            }
         }
 
         protected override void ProcessRecord()
         {
             WriteDebug("Entering GetInstalledPSResource");
 
+            Name = Utils.ProcessNameWildcards(Name, out string[] errorMsgs, out bool _);
```
PaulHigin: Please make this a local variable, since it is not used class wide.

src/code/InstallPSResource.cs
```diff
@@ -136,6 +136,33 @@ protected override void ProcessRecord()
             switch (ParameterSetName)
             {
                 case NameParameterSet:
+                    Name = Utils.ProcessNameWildcards(Name, out string[] errorMsgs, out bool nameContainsWildcard);
```
PaulHigin: Please make Name local variable is appropriate.

test/InstallPSResource.Tests.ps1
```diff
@@ -22,6 +22,17 @@ Describe 'Test Install-PSResource for Module' {
         Get-RevertPSResourceRepositoryFile
     }
 
+    $testCases = @{Name="*";                          ErrorId="NameContainsWildcard"}
+                 @{Name=$null;                        ErrorId="NameContainsWildcard"},
```
PaulHigin: Please remove null case, which is not needed.

PaulHigin (changes requested):
Also please add tests for Save-PSResource, Uninstall-PSResource.

PaulHigin (approved):
LGTM

src/code/UninstallPSResource.cs
```diff
@@ -83,6 +84,31 @@ protected override void ProcessRecord()
             switch (ParameterSetName)
             {
                 case NameParameterSet:
+                    Name = Utils.ProcessNameWildcards(Name, out string[] errorMsgs, out bool nameContainsWildcard);
```
PaulHigin: Please user local variable here instead of 'Name' property.

src/code/UninstallPSResource.cs
```diff
@@ -16,7 +17,7 @@ namespace Microsoft.PowerShell.PowerShellGet.Cmdlets
     /// <summary>
     /// Uninstall-PSResource uninstalls a package found in a module or script installation path.
     /// </summary>
-    [Cmdlet(VerbsLifecycle.Uninstall, "PSResource", DefaultParameterSetName = NameParameterSet, SupportsShouldProcess = true, HelpUri = "<add>")]
+    [Cmdlet(VerbsLifecycle.Uninstall, "PSResource", DefaultParameterSetName = NameParameterSet, SupportsShouldProcess = true, ConfirmImpact = ConfirmImpact.High)]
```
PaulHigin: Let's hold off on specifying high impact for now.  Not sure how it should work, and also the '-Force' parameter should override it for non-interactive use (or possibly -Confirm:$false).  But I agree that uninstalling all installed resources is a high impact operation.

PaulHigin (changes requested):

PaulHigin (approved):

PR #449, opened 2021-08-06
Support Azure Artifacts for PowerShell modules

### LINKED ISSUE #48: Support Azure Artifacts as a core scenario
JavaScript, C#, Java, and Python all have well-defined processes for
* publishing and hosting private modules in Azure Artifacts
* using Azure Artifacts as a secure mirror for public repositories

While there was a recently published [guide](https://devblogs.microsoft.com/powershell/using-powershellget-with-azure-artifacts/) on the subject, I have been unable to get it working (presumably due to PowerShell/PowerShellGetv2#433 ), and even still, there is no information on added PowerShell Gallery as an upstream package source, so there is no way of applying Component Governance to PowerShell modules or avoiding downloading from a public repository, which is forbidden in many security contexts.

Using Azure Artifacts as a private repository and mirror should be a core priority for PowerShellGet, as it enables PowerShell developers to operate at the same pedigree as application developers.  Please use this scenario when prioritizing features and writing documentation.

anamnavi:
# PR Summary

If Azure Artifacts (or other repository) contains packages that are not PowerShell packages they will not contain a `.psd1` file. Add error handling for this case.

## PR Context

Resolves #48 

## PR Checklist

    - [ ] None
    - **OR**
        - [ ] Issue filed: 
- **User-facing changes**
    - [ ] Not Applicable
    - **OR**
        - [ ] Issue filed: 
    - **OR**
- **Tooling**
    - **OR**

### DIFF
--- a/src/code/InstallHelper.cs
+++ b/src/code/InstallHelper.cs
@@ -1,3 +1,4 @@
+using System.Text;
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 using System;
@@ -385,10 +386,22 @@ namespace Microsoft.PowerShell.PowerShellGet.Cmdlets
                     if (!isScript)
                     {
                         var moduleManifest = Path.Combine(tempDirNameVersion, pkgIdentity.Id + ".psd1");
+                        if (!File.Exists(moduleManifest))
+                        {
+                            // NuGet resources may not contain a PowerShell resource, so the manifest file may not exist.
+                            // Note: Every PowerShell resource is wrapped into a NuGet resource and published, but
+                            // not every NuGet resource contains a PowerShell resource wrapped within
+                            var message = String.Format("Module manifest file: {0} does not exist. This is not a valid PowerShell resource.", moduleManifest);
+                            var ex = new ArgumentException(message);
+                            var psdataFileDoesNotExistError = new ErrorRecord(ex, "psdataFileNotExistError", ErrorCategory.ReadError, null);
+                            _cmdletPassedIn.WriteError(psdataFileDoesNotExistError);
+                            continue;
+                        }
 
                         var parsedMetadataHashtable = Utils.ParseModuleManifest(moduleManifest, _cmdletPassedIn);
                         if (parsedMetadataHashtable.Count == 0)
                         {
+                            // we ran into errors parsing the module manifest file which was found. Utils.ParseModuleManifest() wrote the errors
                             continue;
                         }
 
@@ -437,13 +450,17 @@ namespace Microsoft.PowerShell.PowerShellGet.Cmdlets
                 {
                     // Delete the temp directory and all its contents
                     _cmdletPassedIn.WriteVerbose(string.Format("Attempting to delete '{0}'", tempInstallPath));
+                    
                     if (Directory.Exists(tempInstallPath))
                     {
-                        Directory.Delete(tempInstallPath, true);
-                    }
-                    if (!Directory.Exists(tempInstallPath))
-                    {
-                        _cmdletPassedIn.WriteVerbose(string.Format("Sucessfully deleted '{0}'", tempInstallPath));
+                        if (!TryDeleteDirectory(tempInstallPath, out ErrorRecord errorMsg))
+                        {
+                            _cmdletPassedIn.WriteError(errorMsg);
+                        }
+                        else
+                        {
+                            _cmdletPassedIn.WriteVerbose(String.Format("Successfully deleted '{0}'", tempInstallPath));
+                        }
                     }
                 }
             }
@@ -590,6 +607,25 @@ namespace Microsoft.PowerShell.PowerShellGet.Cmdlets
             }
         }
 
+        private bool TryDeleteDirectory(
+            string tempInstallPath,
+            out ErrorRecord errorMsg)
+        {
+            errorMsg = null;
+
+            try
+            {
+                Utils.DeleteDirectory(tempInstallPath);
+            }
+            catch (Exception e)
+            {
+                var TempDirCouldNotBeDeletedError = new ErrorRecord(e, "errorDeletingTempInstallPath", ErrorCategory.InvalidResult, null);
+                errorMsg = TempDirCouldNotBeDeletedError;
+                return false;
+            }
+
+            return true;
+        }
         private void MoveFilesIntoInstallPath(
             PSResourceInfo p, 
             bool isScript, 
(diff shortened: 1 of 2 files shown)


src/code/Utils.cs
```diff
@@ -394,6 +394,15 @@ public static Hashtable ParseModuleManifest(string moduleFileInfo, PSCmdlet cmdl
             // a module will still need the module manifest to be parsed.
             if (moduleFileInfo.EndsWith(".psd1", StringComparison.OrdinalIgnoreCase))
             {
+                if (!File.Exists(moduleFileInfo))
+                {
+                    var message = String.Format("File {0} does not exist {0}. This is not a valid PowerShell package.", moduleFileInfo);
```
alerickson: Everything looks good, just a comment to consider changing 'package' to 'resource'

alerickson (approved):

src/code/InstallHelper.cs
```diff
@@ -386,7 +386,12 @@ private List<string> InstallPackage(IEnumerable<PSResourceInfo> pkgsToInstall, s
                     {
                         var moduleManifest = Path.Combine(tempDirNameVersion, pkgIdentity.Id + ".psd1");
 
-                        var parsedMetadataHashtable = Utils.ParseModuleManifest(moduleManifest, this);
+                        var parsedMetadataHashtable = Utils.ParseModuleManifest(moduleManifest, _cmdletPassedIn);
+                        if (parsedMetadataHashtable.Count == 0)
+                        {
+                            continue;
```
PaulHigin, line 401: Is this an error condition (invalid or non-existing module manifest?).  Should we emit an error message or at least a verbose message?

src/code/InstallHelper.cs
```diff
@@ -436,6 +441,10 @@ private List<string> InstallPackage(IEnumerable<PSResourceInfo> pkgsToInstall, s
                     {
                         Directory.Delete(tempInstallPath, true);
                     }
+                    if (!Directory.Exists(tempInstallPath))
```
PaulHigin: Directory.Delete will throw an exception if it fails, so this check is not needed and you can add the verbose message after line 442.  But if this fails (access denied, File.IO exceptions), it will break the loop.  It seems like we want to catch a failure, emit a message, and continue processing packages.

I suggest creating a utility method 'TryDeleteDirectory'.  Also we should be using Utils.DeleteDirectory().

PaulHigin (changes requested):

src/code/InstallHelper.cs
PaulHigin: If this is expected because we assume it is an artifact resource, please add a comment here to indicate that.  Is there any other way to determine an 'Az artifact' other than it does not have a manifest file?  Utils.ParseModuleManifest should have a way to indicate that no file exists, as opposed to the file exists but is empty (invalid manifest), or we should directly check for its existence here to make the code clear.

alerickson: This is an error condition, but there is error handling that happens within ParseModuleManifest that specifies that the manifest is invalid, for example.  However, because that error isn't terminating the method would return and continue on with the rest of this method.  Anam added the check to see if the hashtable is empty so that we know the parsing failed (at this point an error has already been displayed), and can exit out of this iteration.  

Another thing we could do is change 'ParseModuleManifest' to 'TryParseModuleManifest'  that way if false gets returned we just continue on to the next loop iteration instead of checking if the hashtable's empty.

src/code/InstallHelper.cs
anamnavi: I moved this into a helper method named TryDeleteDirectory()

src/code/InstallHelper.cs
PaulHigin: So parsedMetadataHashtable.Count == 0 means that there is no manifest file?  If so, please add a comment to make that clear.

PaulHigin: And also that it means the resource is a Az artifact and not a powershell module...

anamnavi: I agree that the error handling (to check if the module manifest file exists- which would only be false if the resource was a NuGet resource, not a PowerShell resource wrapped in a NuGet resource) should be done earlier.

So I modified this code so that we check if the module manifest file exists in InstallHelper. Utils.ParseModuleManifest() should only be called if the module manifest file exists. I think it makes sense for the caller to handle this.

I do the `parsedMetadataHashtable.Count == 0)` because by default the returned hashtable there is set to a new empty one. And if it's not filled then it'll have a count of 0 and no keys and cause exceptions to be thrown later.

anamnavi: @PaulHigin no, now if parsedMetadataHashtable.Count == 0, it would mean that we had errors while parsing the .psd1 file (which we wrote errors for in Utils.ParseModuleManifest(). But before calling this method we verified that the module manifest file existed- thus it was a PowerShell resource, not a NuGet resource)

src/code/InstallHelper.cs
```diff
@@ -385,8 +386,25 @@ private List<string> InstallPackage(IEnumerable<PSResourceInfo> pkgsToInstall, s
                     if (!isScript)
                     {
                         var moduleManifest = Path.Combine(tempDirNameVersion, pkgIdentity.Id + ".psd1");
+                        if (!File.Exists(moduleManifest))
+                        {
+                            // NuGet resources may not contain a PowerShell resource, so the manifest file may not exist.
+                            // Note: Every PowerShell resource is wrapped into a NuGet resource and published, but
+                            // not every NuGet resource contains a PowerShell resource wrapped within
+                            var message = String.Format("Module manifest file: {0} does not exist. This is not a valid PowerShell resource.", moduleManifest);
```
alerickson:
```suggestion
                            var message = String.Format("Module manifest file: {0} does not exist. This is not a valid PowerShell module.", moduleManifest);

```

src/code/InstallHelper.cs
```diff
@@ -385,8 +385,25 @@ private List<string> InstallPackage(IEnumerable<PSResourceInfo> pkgsToInstall, s
...
+
+                            var ex = new ArgumentException(message);
+                            var psdataFileDoesNotExistError = new ErrorRecord(ex, "psdataFileNotExistError", ErrorCategory.ReadError, null);
+                            _cmdletPassedIn.WriteError(psdataFileDoesNotExistError);
+                            continue;
+                        }
+
+                        if (!Utils.TryParseModuleManifest(moduleManifest, _cmdletPassedIn, out Hashtable parsedMetadataHashtable))
+                        {
+                            if (parsedMetadataHashtable.Count == 0)
```
alerickson: I think we should be able to get rid of this condition since the only time the hashtable will be empty is if the TryParse failed

PaulHigin (approved):
This looks good, after addressing @alerickson comment.

PR #450, opened 2021-08-06
Auto-create PSResourceRepository.xml file if the file does not already exist. 

### LINKED ISSUE #258: First run experience is... lacking. Cannot find XML

alerickson:
# PR Summary
Automatically create the PSGet repository store (PSResourceRepository.xml file) if the file does not already exist. 
This functionality already existed in the *-PSResourceRepository cmdlets, but did not exist for the *-PSResource cmdlets.  All of the latter cmdlets that read from the PSResourceRepository.xml file (ie, Find, Install, Save, and Publish) now check to makes sure that the file already exists before processing.  If the file does not exist, it gets created.

Another change is that upon each new PSResourceRepository.xml file creation, the PowerShellGallery gets registered by default with the following settings:
```
Name:  PSGallery
URL:  https://powershellgallery.com/api/v2
Priority: 50
Trusted: false
```

## PR Context

Resolves #258 

## PR Checklist

    - [x] None
    - **OR**
        - [ ] Issue filed: 
- **User-facing changes**
    - [x] Not Applicable
    - **OR**
        - [ ] Issue filed: 
    - **OR**
- **Tooling**
    - **OR**

### DIFF
--- a/src/code/FindPSResource.cs
+++ b/src/code/FindPSResource.cs
@@ -138,18 +138,7 @@ namespace Microsoft.PowerShell.PowerShellGet.Cmdlets
 
             // Create a respository story (the PSResourceRepository.xml file) if it does not already exist
             // This is to create a better experience for those who have just installed v3 and want to get up and running quickly
-            try
-            {
-                RepositorySettings.CheckRepositoryStore();
-            }
-            catch (PSInvalidOperationException e)
-            {
-                ThrowTerminatingError(new ErrorRecord(
-                    new PSInvalidOperationException(e.Message),
-                    "RepositoryStoreException",
-                    ErrorCategory.ReadError,
-                    this));
-            }
+            RepositorySettings.CheckRepositoryStore();
         }
 
         protected override void StopProcessing()
--- a/src/code/FindPSResource.cs
+++ b/src/code/FindPSResource.cs
@@ -135,6 +135,21 @@ namespace Microsoft.PowerShell.PowerShellGet.Cmdlets
         {
             _source = new CancellationTokenSource();
             _cancellationToken = _source.Token;
+
+            // Create a respository story (the PSResourceRepository.xml file) if it does not already exist
+            // This is to create a better experience for those who have just installed v3 and want to get up and running quickly
+            try
+            {
+                RepositorySettings.CheckRepositoryStore();
+            }
+            catch (PSInvalidOperationException e)
+            {
+                ThrowTerminatingError(new ErrorRecord(
+                    new PSInvalidOperationException(e.Message),
+                    "RepositoryStoreException",
+                    ErrorCategory.ReadError,
+                    this));
+            }
         }
 
         protected override void StopProcessing()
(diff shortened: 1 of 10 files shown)


src/code/RepositorySettings.cs
```diff
@@ -27,7 +27,10 @@ internal static class RepositorySettings
         private static readonly string RepositoryFileName = "PSResourceRepository.xml";
         private static readonly string RepositoryPath = Path.Combine(Environment.GetFolderPath(SpecialFolder.LocalApplicationData), "PowerShellGet");
         private static readonly string FullRepositoryPath = Path.Combine(RepositoryPath, RepositoryFileName);
-
+        private static readonly string PSGalleryRepoName = "PSGallery";
```
PaulHigin: These should be 'private const string' types.

src/code/RepositorySettings.cs
```diff
@@ -27,7 +27,10 @@ internal static class RepositorySettings
         private static readonly string RepositoryFileName = "PSResourceRepository.xml";
```
PaulHigin: This should also be 'private const string'

src/code/FindPSResource.cs
```diff
@@ -134,7 +134,22 @@ public sealed class FindPSResource : PSCmdlet
...
-            _cancellationToken = _source.Token;
+            _cancellationToken = _source.Token;
+
+            // Create a respository story (the PSResourceRepository.xml file) if it does not already exist
+            // This is to create a better experience for those who have just installed v3 and want to get up and running quickly
+            try
+            {
+                RepositorySettings.CheckRepositoryStore();
+            }
+            catch (PSInvalidOperationException e)
```
PaulHigin: It feels unnecessary to catch and rethrow the PSInvalidOperationException error.  I think we can simplify this by removing the try/catch and just have 'RepositorySettings.CheckRepositoryStore();' check, and let the original thrown exception terminate the cmdlet.  The error message is clear, and the error stack is preserved.

PaulHigin (changes requested):

PaulHigin (approved):

src/code/FindPSResource.cs
anamnavi: I agree.

anamnavi (approved):
LGTM

Issue #383, opened 2021-06-22
Install-PSResource thinks Pester is a script not a module


hbuckle:
Installing the latest version of Pester (5.2.2) does not install the module, it just installs a `Pester.ps1` script into `Documents\PowerShell\Scripts`

```
Install-PSResource -Name Pester -Version 5.2.2 -Scope CurrentUser -Verbose -Debug -Repository PSGallery
DEBUG: Parameters passed in >>> Name: 'Pester'; Version: '5.2.2'; Prerelease: 'False'; Repository: 'PSGallery'; Scope: 'CurrentUser'; AcceptLicense: 'False'; Quiet: 'False'; Reinstall: 'False'; TrustRepository: 'False'; NoClobber: 'False';
DEBUG: Entering InstallHelper::ProcessInstallParams
DEBUG: Console is elevated: 'False'
DEBUG: Console is Windows PowerShell: 'False'
DEBUG: Current user scope installation path: 'C:\Users\henry.buckle\Documents\PowerShell'
DEBUG: All users scope installation path: 'C:\Program Files\PowerShell'
VERBOSE: Scope is: CurrentUser
DEBUG: Checking to see if paths exist
DEBUG: Path: 'C:\Users\henry.buckle\Documents\PowerShell\Modules'  >>> exists? 'True'
DEBUG: Path: 'C:\Users\henry.buckle\Documents\PowerShell\Scripts'  >>> exists? 'True'
DEBUG: Path: 'C:\Users\henry.buckle\Documents\PowerShell\Scripts\InstalledScriptInfos'  >>> exists? 'True'
DEBUG: Attempting to search for packages in 'PSGallery'
DEBUG: Checking if untrusted repository should be used

Untrusted repository
You are installing the modules from an untrusted repository. If you trust this repository, change its Trusted value by running the Set-PSResourceRepository cmdlet. Are you sure you want to install the PSresource from 'PSGallery' ?
[Y] Yes  [A] Yes to All  [N] No  [L] No to All  [S] Suspend  [?] Help (default is "N"): y
DEBUG: Untrusted repository accepted as trusted source.
DEBUG: Parsed version is: '[5.2.2, 5.2.2]'
VERBOSE: Begin installing package: 'Pester'
DEBUG: Successfully able to download package from source to: 'C:\Users\henry.buckle\AppData\Local\Temp\e34066dc-c3a0-4da8-bad8-615d7ff351b8'
[...]

PR #504, opened 2021-10-06
Fix bug with some modules installing as scripts

alerickson:
# PR Summary

Any module that contained both a .ps1 file and a .psd1 file was being read as a script instead of a module.  This was causing issues with Pester in particular installing in the incorrect location (under the /scripts path instead of /modules).  

I've changed the condition to determine whether a package is a module or a script to check for a .psd1 instead of checking for a .ps1.  This logic makes more sense because a module can contain a .ps1 but a script cannot contain a .psd1.
 
## PR Context

Resolves issue #383 

## PR Checklist

    - [x] None
    - **OR**
        - [ ] Issue filed: 
- **User-facing changes**
    - [x] Not Applicable
    - **OR**
        - [ ] Issue filed: 
    - **OR**
- **Tooling**
    - **OR**

### DIFF
--- a/src/code/InstallHelper.cs
+++ b/src/code/InstallHelper.cs
@@ -376,9 +376,11 @@ namespace Microsoft.PowerShell.PowerShellGet.Cmdlets
                     var version4digitNoPrerelease = pkgIdentity.Version.Version.ToString();
                     string moduleManifestVersion = string.Empty;
                     var scriptPath = Path.Combine(tempDirNameVersion, (p.Name + ".ps1"));
-                    var isScript = File.Exists(scriptPath) ? true : false;
+                    var modulePath = Path.Combine(tempDirNameVersion, (p.Name + ".psd1"));
+                    // Check if the package is a module or a script
+                    var isModule = File.Exists(modulePath) ? true : false;
 
-                    if (!isScript)
+                    if (isModule)
                     {
                         var moduleManifest = Path.Combine(tempDirNameVersion, pkgIdentity.Id + ".psd1");
                         if (!File.Exists(moduleManifest))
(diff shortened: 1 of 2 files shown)


test/InstallPSResource.Tests.ps1
```diff
@@ -243,6 +243,13 @@ Describe 'Test Install-PSResource for Module' {
         $res = Get-Module "TestModule" -ListAvailable
         $res | Should -BeNullOrEmpty
     }
+
+    It "Validate that Pester is installing under Modules path" {
```
PaulHigin:
```suggestion
    It "Validates that a module with module-name script files (like Pester) installs under Modules path" {

```

PaulHigin (approved):

src/code/InstallHelper.cs
```diff
@@ -376,9 +376,11 @@ private List<string> InstallPackage(IEnumerable<PSResourceInfo> pkgsToInstall, s
                     var version4digitNoPrerelease = pkgIdentity.Version.Version.ToString();
                     string moduleManifestVersion = string.Empty;
                     var scriptPath = Path.Combine(tempDirNameVersion, (p.Name + ".ps1"));
-                    var isScript = File.Exists(scriptPath) ? true : false;
+                    var modulePath = Path.Combine(tempDirNameVersion, (p.Name + ".psd1"));
+                    // Check if the package is a module or a script
+                    var isModule = File.Exists(modulePath) ? true : false;
```
anamnavi: just a nitpick, but logic simplification suggestion here:
```suggestion
                    var isModule = File.Exists(modulePath);

```

anamnavi (approved):
LGTM, just minor logic simplification suggestion but is not a blocker

PR #506, opened 2021-10-08
Implement -DestinationPath parameter for Publish-PSResource

### LINKED ISSUE #430: Add -DestinationPath parameter to the Publish-PSResource cmdlet.
This is a reminder to add this new parameter set as an alternative location (other than a Repository selection) to publish a PSResource package.

```powershell
Publish-PSResource -Path c:\Modules\MyModule -DestinationPath c:\Packages
```

alerickson:
# PR Summary
Implements the -DestinationPath parameter for Publish-PSResource.  This allows the user to save the generated .nupkg that gets pushed to a repository to to a specified path on the local file system as well.

## PR Context
Resolves #430 

## PR Checklist

    - [x] None
    - **OR**
        - [ ] Issue filed: 
- **User-facing changes**
    - [x] Not Applicable
    - **OR**
        - [ ] Issue filed: 
    - [] N/A or can only be tested interactively
    - **OR**
- **Tooling**
    - **OR**

### DIFF
--- a/src/code/PublishPSResource.cs
+++ b/src/code/PublishPSResource.cs
@@ -48,7 +48,7 @@ namespace Microsoft.PowerShell.PowerShellGet.Cmdlets
         /// Specifies the path to the resource that you want to publish. This parameter accepts the path to the folder that contains the resource.
         /// Specifies a path to one or more locations. Wildcards are permitted. The default location is the current directory (.).
         /// </summary>
-        [Parameter(Mandatory = true, Position = 0, ParameterSetName = "PathParameterSet")]
+        [Parameter()]
         [ValidateNotNullOrEmpty]
         public string Path
         {
@@ -74,37 +74,10 @@ namespace Microsoft.PowerShell.PowerShellGet.Cmdlets
             }
         }
         private string _path;
-        
-        /// <summary>
-        /// Specifies a path to one or more locations. Unlike the Path parameter, the value of the LiteralPath parameter is used exactly as entered.
-        /// No characters are interpreted as wildcards. If the path includes escape characters, enclose them in single quotation marks.
-        /// Single quotation marks tell PowerShell not to interpret any characters as escape sequences.
-        /// </summary>
-        [Parameter(Mandatory = true, ParameterSetName = "PathLiteralParameterSet")]
-        [ValidateNotNullOrEmpty]
-        public string LiteralPath
-        {
-            get
-            { return _literalPath; }
-
-            set
-            {
-                if (Directory.Exists(value))
-                {
-                    _literalPath = value;
-                }
-                else if (File.Exists(value) && value.EndsWith(".ps1", StringComparison.OrdinalIgnoreCase))
-                {
-                    _literalPath = value;
-                }
-            }
-        }
-        private string _literalPath;
 
         /// <summary>
         /// Specifies the path to where the resource (as a nupkg) should be saved to. This parameter can be used in conjunction with the
         /// -Repository parameter to publish to a repository and also save the exact same package to the local file system.
-        /// Specifies a path to one or more locations. Wildcards are permitted. The default location is the current directory (.).
         /// </summary>
         [Parameter()]
         [ValidateNotNullOrEmpty]
@@ -125,11 +98,19 @@ namespace Microsoft.PowerShell.PowerShellGet.Cmdlets
                 {
                     _destinationPath = resolvedPath;
                 }
-                // consider throwing error if path does not exist
                 else {
-                    var ex = new ArgumentException("Destination path does not exist.");
-                    var DestinationPathDoesNotExist = new ErrorRecord(ex, "DestinationPathDoesNotExist", ErrorCategory.InvalidArgument, null);
-                    WriteError(DestinationPathDoesNotExist);
+                    // try to create the path 
+                    try
+                    {
+                        Directory.CreateDirectory(value);
+                    }
+                    catch (Exception e)
+                    {
+                        var exMessage = string.Format("Destination path does not exist and cannot be created: {0}", e.Message);
+                        var ex = new ArgumentException(exMessage);
+                        var InvalidDestinationPath = new ErrorRecord(ex, "InvalidDestinationPath", ErrorCategory.InvalidArgument, null);
+                        ThrowTerminatingError(InvalidDestinationPath);
+                    }
                 }
             }
         }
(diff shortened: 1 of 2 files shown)


src/code/PublishPSResource.cs
```diff
@@ -101,6 +101,40 @@ public string LiteralPath
...
+
+                if (Directory.Exists(resolvedPath))
+                {
+                    _destinationPath = resolvedPath;
+                }
+                // consider throwing error if path does not exist
+                else {
+                    var ex = new ArgumentException("Destination path does not exist.");
+                    var DestinationPathDoesNotExist = new ErrorRecord(ex, "DestinationPathDoesNotExist", ErrorCategory.InvalidArgument, null);
+                    WriteError(DestinationPathDoesNotExist);
```
anamnavi: so in this case, we'd just write the error and `_destinationPath` wouldn't be set to anything, i.e be an empty string?

src/code/PublishPSResource.cs
```diff
@@ -374,7 +408,27 @@ protected override void ProcessRecord()
...
+                // If -DestinationPath is specified then also publish the .nupkg there
+                if (!string.IsNullOrWhiteSpace(_destinationPath))
+                {
+                    try
+                    {
+                        var nupkgName = _pkgName + "." + _pkgVersion.ToNormalizedString() + ".nupkg";
+                        File.Move(System.IO.Path.Combine(outputNupkgDir, nupkgName), System.IO.Path.Combine(_destinationPath, nupkgName));
+                    }
+                    catch (Exception e) {
+                        var message = string.Format("Error moving .nupkg into destination path: '{0}'.", e.Message);
```
anamnavi: I read this expecting the `{0}` value to be the destinationPath, not the reason why the nupkg couldn't be moved to the destinationPath, so perhaps:

```suggestion
                        var message = string.Format("Error moving .nupkg into destination path '{0}' due to: '{1}'.", _destinationPath, e.Message);

```

src/code/PublishPSResource.cs
```diff
@@ -374,7 +408,27 @@ protected override void ProcessRecord()
                     return;
                 }
 
-                 PushNupkg(outputNupkgDir, repositoryUrl);
+                // This call does not throw any exceptions, but it will return unsuccessful response status codes
+                PushNupkg(outputNupkgDir, repositoryUrl);
```
anamnavi: do we need to handle the cases where we get unsuccessful response status codes?

test/PublishPSResource.Tests.ps1
```diff
@@ -271,4 +271,20 @@ Describe "Test Publish-PSResource" {
...
+    It "Publish a module with -Path -Repository and -DestinationPath" {
+        $version = "1.0.0"
+        New-ModuleManifest -Path (Join-Path -Path $script:PublishModuleBase -ChildPath "$script:PublishModuleName.psd1") -ModuleVersion $version -Description "$script:PublishModuleName module"
+
+        $tmpPath = Join-Path -Path $TestDrive -ChildPath "testtmppath"
+        New-Item $tmpPath -Itemtype directory -Force
+
+        Publish-PSResource -Path $script:PublishModuleBase -Repository $testRepository2 -DestinationPath $tmpPath
+
+        $expectedPath = Join-Path -Path $script:repositoryPath2  -ChildPath "$script:PublishModuleName.$version.nupkg"
```
anamnavi:
```suggestion
        $expectedPath = Join-Path -Path $script:repositoryPath2 -ChildPath "$script:PublishModuleName.$version.nupkg"

```

anamnavi (approved):
LGTM, just some minor nitpicks. Also perhaps we can update the help doc for the `-DestinationPath` parameter that would be great. I can definitely also help with this if needed :)

src/code/PublishPSResource.cs
alerickson: Ideally yes, but this method does not return anything and does not throw exceptions, it just writes to the console.  @PaulHigin do you have any suggestions on how to properly address this situation?

src/code/PublishPSResource.cs
alerickson: Yes, so essentially the -DestinationPath param would no longer be applicable.  Although now I'm wondering if we should throw a terminating error instead and ensure the user passes in a valid argument.

src/code/PublishPSResource.cs
```diff
@@ -101,6 +101,40 @@ public string LiteralPath
         }
         private string _literalPath;
 
+        /// <summary>
+        /// Specifies the path to where the resource (as a nupkg) should be saved to. This parameter can be used in conjunction with the
+        /// -Repository parameter to publish to a repository and also save the exact same package to the local file system.
+        /// Specifies a path to one or more locations. Wildcards are permitted. The default location is the current directory (.).
```
PaulHigin: I don't understand this comment description.  How can a single path string specify more than one location?  How can and why would a package be saved to multiple locations?  Is there a limit on the number of locations?  Why is there a default location, if this is an optional parameter ... a valid destination location is specified or it isn't.

src/code/PublishPSResource.cs
```diff
@@ -101,6 +101,40 @@ public string LiteralPath
...
+        [Parameter()]
+        [ValidateNotNullOrEmpty]
+        public string DestinationPath
+        {
+            get
+            { return _destinationPath; }
+
+            set
+            {
+                string resolvedPath = string.Empty;
```
PaulHigin, line 91: It should be Ok to set the path to null or empty string as this just means the package is not published to a local file path.
This should be the first check and return immediately.  If the path doesn't exist we should either create it (-Force) or throw terminating error.

However, it feels strange to have this option in `Publish-PSResource`, since it is doing two different things.  One is to publish to a repository.  The other is to save the package to a file path.  At the very lease I feel this should be a separate parameter set, and maybe should be a separate cmdlet (`Publish-PSResourceToPath`).

src/code/PublishPSResource.cs
PaulHigin: Sounds like a poor API, we should try to get the owners to fix.  In the mean time, is there a way to verify the publishing?  Maybe add an optional switch parameter to do the verification.

src/code/PublishPSResource.cs
```diff
@@ -374,7 +408,28 @@ protected override void ProcessRecord()
...
+                // This call does not throw any exceptions, but it will return unsuccessful response status codes
+                PushNupkg(outputNupkgDir, repositoryUrl);
+
+                // If -DestinationPath is specified then also publish the .nupkg there
+                if (!string.IsNullOrWhiteSpace(_destinationPath))
+                {
+                    try
+                    {
+                        var nupkgName = _pkgName + "." + _pkgVersion.ToNormalizedString() + ".nupkg";
+                        File.Move(System.IO.Path.Combine(outputNupkgDir, nupkgName), System.IO.Path.Combine(_destinationPath, nupkgName));
```
PaulHigin: We should consistently use the Utils.FileMove method that works over different volumes.

PaulHigin (changes requested):

src/code/PublishPSResource.cs
alerickson: Check that's a valid path and that the path actually exists.  Write a terminating error.

alerickson:
- Consider creating a new parameter set for -DestinationPath
     - Publish-PSResource -Path -DestinationPath
     - Publish-PSResource -LiteralPath -DestinationPath

- Consider getting rid of the -LiteralPath parameter

Issue #498, opened 2021-09-30
Install-PSResource should restore if reinstall fails


PaulHigin:
### Prerequisites

- [X] Write a descriptive title.
- [X] Make sure you are able to repro it on the [latest released version](https://www.powershellgallery.com/packages/PowerShellGet)
- [X] Search the existing issues.

### Steps to reproduce

When I install the `Microsoft.PowerShell.SecretStore` extension vault module, Install-PSResource also tries to reinstall the `Microsoft.PowerShell.SecretManagement` module that it depends on, even though the `-Reinstall` parameter switch is not used.

However, the `Microsoft.PowerShell.SecretManagement` module is currently loaded and so the reinstall fails because it is a binary cmdlet and the `Microsoft.PowerShell.SecretManagement.dll` cannot be deleted.

The result is the `Microsoft.PowerShell.SecretManagement` module is corrupted since all files are deleted except for the one binary.

I feel a reinstall operation should not occur unless specifically specified by the user.  Also, and if a reinstall fails we should have a fallback that restores the original installation.

The corruption occurs without any user error, so it is not apparent that SecretManagement is broken until the user tries running one of its commands.

### Expected behavior

```console
PS C:\> Get-SecretInfo

Name                 Type         VaultName
----                 ----         ---------
Secret1              SecureString AzKeyVault

PS C:\> Install-PSResource -Name Microsoft.PowerShell.SecretManagement -Repository PSGallery -Scope AllUsers

PS C:\> Register-SecretVault -Name SecretStore -ModuleName Microsoft.PowerShell.SecretStore -AllowClobber
```

### Actual behavior

```console
PS C:\> Register-SecretVault -Name SecretStore -ModuleName Microsoft.PowerShell.SecretStore -AllowClobber
Register-SecretVault: Could not load and retrieve module information for module: Microsoft.PowerShell.SecretStore with error : The required module 'Microsoft.PowerShell.SecretManagement' is not loaded. Load the module or remove the module from 'RequiredModules' in the file 'C:\Program Files\PowerShell\Modules\Microsoft.PowerShell.SecretStore\1.0.5\Microsoft.PowerShell.SecretStore.psd1'..
```

### Error details

```console
Error occurs because all `Microsoft.PowerShell.SecretManagement` module files are deleted except for the binary file.
```

### Environment data

```powershell
ModuleType Version    PreRelease Name                                ExportedCommands
---------- -------    ---------- ----                                ----------------
[...]

PR #525, opened 2021-10-29
Fix for Install-PSResource always reinstall issue

PaulHigin:
# PR Summary

This PR fixes the 'always reinstall and restore on fail issue' of Issue #498, and also updates the build version to the next release (beta12).  This PR also includes some general code clean up.

## PR Context

`Install-PSResource` was always reinstalling modules even though `-Reinstall` switch was not selected.  This is because the code was not correctly removing 'already installed' packages from the 'toinstall' list.

Also, if a reinstall fails because one of the files is locked (as can happen with loaded .NET binaries) then an error message is emitted and the original resource module files are restored.

Unused variables and parameter arguments were removed.
Some minor code refactoring and method renaming for better clarity.
Added some code regions where they were missing.
Alphabetized using statements per code style guidelines.
Upped module version to 3.0.12 beta.

## PR Checklist

    - [x] None
    - **OR**
        - [ ] Issue filed: 
- **User-facing changes**
    - [x] Not Applicable
    - **OR**
        - [ ] Issue filed: 
    - **OR**
- **Tooling**
    - **OR**

### DIFF
--- a/doBuild.ps1
+++ b/doBuild.ps1
@@ -36,6 +36,14 @@ function DoBuild
     Write-Verbose -Verbose -Message "Copying help files to '$BuildOutPath'"
     Copy-Item -Path "${HelpPath}/${Culture}" -Dest "$BuildOutPath" -Recurse -Force
 
+    # Copy license
+    Write-Verbose -Verbose -Message "Copying LICENSE file to '$BuildOutPath'"
+    Copy-Item -Path "./LICENSE" -Dest "$BuildOutPath"
+
+    # Copy notice
+    # Write-Verbose -Verbose -Message "Copying ThirdPartyNotices.txt to '$BuildOutPath'"
+    # Copy-Item -Path "./ThirdPartyNotices.txt" -Dest "$BuildOutPath"
+
     #
     # Copy DSC resources
     # TODO: This should not be part of PowerShellGet build/publish and should be moved to its own project
@@ -65,40 +73,21 @@ function DoBuild
             }
 
             # Place build results
-            if ($BuildFramework -eq "netstandard2.0") {
-                $assemblyNames = @(
-                    'PowerShellGet'
-                    'Microsoft.Extensions.Logging.Abstractions'
-                    'MoreLinq'
-                    'NuGet.Commands'
-                    'NuGet.Common'
-                    'NuGet.Configuration'
-                    'NuGet.Frameworks'
-                    'NuGet.Packaging'
-                    'NuGet.ProjectModel'
-                    'NuGet.Protocol'
-                    'NuGet.Repositories'
-                    'NuGet.Versioning'
-                    'Newtonsoft.Json'
-                )
-            } elseif ($BuildFramework -eq 'net472') {
-                $assemblyNames = @(
-                    'PowerShellGet'
-                    'Microsoft.Extensions.Logging.Abstractions'
-                    'MoreLinq'
-                    'NuGet.Commands'
-                    'NuGet.Common'
-                    'NuGet.Configuration'
-                    'NuGet.Frameworks'
-                    'NuGet.Packaging'
-                    'NuGet.ProjectModel'
-                    'NuGet.Protocol'
-                    'NuGet.Repositories'
-                    'NuGet.Versioning'
-                    'Newtonsoft.Json'
-                    'System.Security.Principal.Windows'
-                )
-            }
+            $assemblyNames = @(
+                'PowerShellGet'
+                'Microsoft.Extensions.Logging.Abstractions'
+                'MoreLinq'
+                'NuGet.Commands'
+                'NuGet.Common'
+                'NuGet.Configuration'
+                'NuGet.Frameworks'
+                'NuGet.Packaging'
+                'NuGet.ProjectModel'
+                'NuGet.Protocol'
+                'NuGet.Repositories'
+                'NuGet.Versioning'
+                'Newtonsoft.Json'
+            )
 
             $buildSuccess = $true
             foreach ($fileName in $assemblyNames)
--- a/src/code/InstallPSResource.cs
+++ b/src/code/InstallPSResource.cs
@@ -125,7 +125,7 @@ namespace Microsoft.PowerShell.PowerShellGet.Cmdlets
 
         protected override void ProcessRecord()
         {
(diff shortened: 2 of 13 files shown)


doBuild.ps1
```diff
@@ -36,6 +36,14 @@ function DoBuild
     Write-Verbose -Verbose -Message "Copying help files to '$BuildOutPath'"
     Copy-Item -Path "${HelpPath}/${Culture}" -Dest "$BuildOutPath" -Recurse -Force
 
+    # Copy license
+    Write-Verbose -Verbose -Message "Copying LICENSE file to '$BuildOutPath'"
+    Copy-Item -Path "./LICENSE" -Dest "$BuildOutPath"
+
+    # Copy notice
+    # Write-Verbose -Verbose -Message "Copying ThirdPartyNotices.txt to '$BuildOutPath'"
```
alerickson, line 44: Is this something we need to follow up on and potentially add in later?

src/code/InstallPSResource.cs
```diff
@@ -125,7 +125,7 @@ protected override void BeginProcessing()
 
         protected override void ProcessRecord()
         {
-            var installHelper = new InstallHelper(updatePkg: false, savePkg: false, cmdletPassedIn: this);
+            var installHelper = new InstallHelper(savePkg: false, cmdletPassedIn: this);
```
anamnavi, line 128: why do we remove 'updatePkg: false'?

src/code/Utils.cs
```diff
@@ -535,6 +580,31 @@ public static void DeleteDirectory(string dirPath)
...
+                string destFilePath = Path.Combine(destDirPath, Path.GetFileName(filePath));
+                if (!File.Exists(destFilePath))
+                {
+                    File.Copy(filePath, destFilePath);
+                }
+            }
+
+            foreach (string srcSubDirPath in Directory.GetDirectories(sourceDirPath))
+            {
+                string destSubDirPath = Path.Combine(destDirPath, Path.GetFileName(srcSubDirPath));
```
anamnavi, line 603: didn't know that `Path.GetFileName()` could be used for directories (not just files). Apparently if path looks like this: `/User/Folder` it'll return folder name. If it looks like: `/User/Folder**/**` it'll return ' ' (empty string).

src/code/Utils.cs
```diff
@@ -535,6 +580,31 @@ public static void DeleteDirectory(string dirPath)
...
+            foreach (string filePath in Directory.GetFiles(sourceDirPath))
+            {
+                string destFilePath = Path.Combine(destDirPath, Path.GetFileName(filePath));
+                if (!File.Exists(destFilePath))
+                {
+                    File.Copy(filePath, destFilePath);
+                }
+            }
+
+            foreach (string srcSubDirPath in Directory.GetDirectories(sourceDirPath))
```
anamnavi, line 601: if sourceDirPath ends with a file (i.e /User/Folder/test_script.ps1) this will just return an empty array? It won't throw an exception for this not being a directory path?

anamnavi (approved):
LGTM, left a few questions/comments for my own understanding.

src/code/Utils.cs
PaulHigin: Yeah, it really returns the 'leaf node' of the path.  I am assuming the `Directory.GetDirectories()` returns well formed directory paths.  Are you suggesting we need an additional check?

src/code/Utils.cs
PaulHigin: Again, I am assuming GetDirectories() is returning well formed directory paths.  `/User/Folder/test_script.ps1` is a valid directory path.  We are manipulating install directories that we create, so I didn't think we needed extra checks.  But if you know of a pathological case that we should check for, let me know.

src/code/Utils.cs
anamnavi: This was comment was my understanding after reading the docs for Directory.GetDirectories(). I left this comment in case anyone reviewing the PR wonders this. 

Since you only get subdirectories in this for loop (from `Directory.GetDirectories(sourceDirPath)`) I think this is fine. If this wasn't the case I'd suggest we have an additional check for Path.GetFilename(srcSubDirPath) returning an empty string but here I believe it's ok.

src/code/Utils.cs
anamnavi: "/User/Folder/test_script.ps1 is a valid directory path."

ok, this makes sense. I think we don't need any additional checks here then.

src/code/UpdatePSResource.cs
```diff
@@ -155,10 +150,9 @@ protected override void ProcessRecord()
                 repository: Repository,
                 acceptLicense: AcceptLicense,
                 quiet: Quiet,
-                reinstall: false,
+                reinstall: true,
```
alerickson, line 153: If reinstall is set to true here, won't that mean that packages that don't need to be updated will get updated? 
eg:  3.0.0 is installed, but because it's the latest version of the module it gets reinstalled even though no action should occur since it is already the latest version. 

In an `update-psresource -name *` we could potentially be installing dozens of modules that don't need to be installed.

doBuild.ps1
PaulHigin: Yes.  I am looking at implementing the notice generation in the build pipeline, but if that doesn't work out we can manually generate the notice and use this to copy it.  I don't think our dependencies will change very much if at all, so manually generate/copy may be good enough.

src/code/UpdatePSResource.cs
PaulHigin: I am assuming that update-psresource creates a list of resources that need to be updated (reinstalled with newer version).  From that list reinstall should always happen, i.e., we always replace the existing module with the newer one.

The update-psresource tests were failing after the fix because reinstall always occurred before the fix, so I made this change with my above assumptions.  I'll double check the code to ensure my assumptions hold.

PaulHigin: Hmm, Ok update-psresource does not check to see if a package version requires an update or not ... it just reinstalls it with the latest version.  That is why the tests began failing when I fixed the 'always reinstall' bug.  We need to separate 'reinstall' from 'update' actions.

I am going check this change in now, and we can fix the update next.

PR #554, opened 2021-12-01
Install should not silently fail

### LINKED ISSUE #154: Install-PSResource did not install module and is silent about it
I tried to install PSReadLine 2.1.0-beta2 like so:
```
06-06 13:42:09 6> Install-PSResource PSReadLine -Prerelease -TrustRepository
```
And got no output so I assumed it installed.  But no, it apparently didn't.  When I run this with `-Verbose` I see this:
```
06-06 13:42:40 7> Install-PSResource PSReadLine -Prerelease -TrustRepository -Verbose
VERBOSE: Current user scope installation path: C:\Users\Keith\Documents\PowerShell
VERBOSE: All users scope installation path: C:\Program Files\PowerShell
VERBOSE: Scope is: CurrentUser
```
If Install-PSResource didn't actually install the module I'd like to at least see a message (or warning) along those lines.

This is on PSGet 3.0.0-beta4 running on PS 7.0.1 on Windows 10 2004.  Also, I had PSRL 2.1.0-beta1 installed with PSGet 2.2.x in `~\Documents\PowerShell\Modules\PSReadLine\2.1.0`.  It appears something also installed PSRL directly in `~\Documents\PowerShell\Modules\PSReadLine`.  Anyway, I removed the PSRL folder and tried again:
```
06-06 14:13:32 1> Install-PSResource PSReadLine -Prerelease -TrustRepository -Verbose
VERBOSE: Current user scope installation path: C:\Users\Keith\Documents\PowerShell
VERBOSE: All users scope installation path: C:\Program Files\PowerShell
VERBOSE: Scope is: CurrentUser
VERBOSE: Verbose statement
```
This time it installed into `~\Documents\PowerShell\Modules\PSReadLine\2.1.0-betat2`.  Yay.  So I'm not sure exactly what this command did earlier when PSRL 2.1.0-beta1 was installed but it did not install anything i.e. PS still loaded 2.1.0-beta1.

Argh!  Now neither PS 7.0.1 nor PS 7.1.0-preview3 will load from that `2.1.0-beta2` directory. Off to rename it to remove the `-beta2`.  Dang! I liked that naming convention because it allows installing new prerelease versions without running into "file in use" conflicts.

anamnavi:
# PR Summary

Install-PSResource should not fail silently. Most of these cases were resolved by PR #525. However for the cases where a package name requested to be installed doesn't exist we want it to not fail silently. 

The new logic is that in InstallHelper we begin with a list of package names (passed in) for packages we wish to install. As installation attempts happen if the installation is successful we remove the name of the installed package from the original list of package names. If the attempt is unsuccessful we write an error and then also remove the name of that package from the original list of package names. At the end, the original list of names is either empty or left with names of packages we couldn't install because they couldn't be found (either name or version for that package). We then write errors for those packages not being found.

InstallHelper handles writing the errors because this way Install-PSResource, Update-PSResource, and Save-PSResource can all leverage the error handling.
## PR Context

Resolves #154

## PR Checklist

    - [ ] None
    - **OR**
        - [ ] Issue filed: 
- **User-facing changes**
    - [ ] Not Applicable
    - **OR**
        - [ ] Issue filed: 
    - **OR**
- **Tooling**
    - **OR**


src/code/InstallPSResource.cs
```diff
@@ -266,6 +267,15 @@ private void ProcessInstallHelper(string[] pkgNames, bool pkgPrerelease, string[
                 savePkg: false,
                 pathsToInstallPkg: _pathsToInstallPkg);
```
anamnavi: Given that we have a for loop for the `- PassThru` condition next, the runtime of this check for whether packages we expected to be installed were actually installed plus the -PassThru check ends up being O(n) + O(n) + O(n) = 3 O(n).

An alternative solution is listed below. With this the runtime would be O(n) + O(n) = 2 O(n). It also doesn't use LINQ. However I feel like the original solution is more clear and the different operations are separated out better for clarity. Also, `-PassThru` won't always be used so runtime may be 2 O(n) most of the time.
```suggestion
            List<string> actualPkgNames = new List<string>();
            foreach (PSResourceInfo pkg in installedPkgs)
            {
                actualPkgNames.Add(pkg.Name);
                if (PassThru)
                {
                    WriteObject(pkg);
                }
            }

            foreach (string expectedName in pkgNames)
            {
                if (!actualPkgNames.Contains(expectedName))
                {
                    WriteWarning(String.Format("Package '{0}' was not installed. Please run the cmdlet with -Verbose to see more information", expectedName));
                }
            }
```

alerickson: I prefer the implementation that's currently in the code.  I think it's a little cleaner and easier to understand and to me the efficiency benefits vs the implementation above seems minimal.

src/code/InstallPSResource.cs
```diff
@@ -266,6 +267,15 @@ private void ProcessInstallHelper(string[] pkgNames, bool pkgPrerelease, string[
...
 
+            List<string> actualPkgNames = installedPkgs.Select(x => x.Name).ToList();
+            foreach(string expectedPkgName in pkgNames)
+            {
+                if (!actualPkgNames.Contains(expectedPkgName))
+                {
+                    WriteWarning(String.Format("Package '{0}' was not installed. Please run the cmdlet with -Verbose for more information", expectedPkgName));
```
alerickson: If we get to this point, does that mean that the package simply wasn't found in any of the repositories?  Should we specify that and then suggest running -Verbose for more info if desired?

alerickson (approved):

src/code/InstallPSResource.cs
```diff
@@ -266,6 +267,15 @@ private void ProcessInstallHelper(string[] pkgNames, bool pkgPrerelease, string[
                 savePkg: false,
                 pathsToInstallPkg: _pathsToInstallPkg);
 
+            List<string> actualPkgNames = installedPkgs.Select(x => x.Name).ToList();
```
```suggestion
            List<string> installedPkgNames = installedPkgs.Select(x => x.Name).ToList();
```

src/code/InstallPSResource.cs
PaulHigin: I think Linq is fine in this case.
However, we should emit an error if a requested resource package could not be installed.
It would be nice to include the reason the install failed (not found, file error, etc.), maybe InstallerHelper.InstallPackages() can return an array of error records?

For example:
```powershell
Install-PSResource -Name MyResource5 -Repository PSGallery
Error: MyResource5 package could not be installed with error: Resource not found.
```

PaulHigin (changes requested):

src/code/InstallPSResource.cs
anamnavi: I changed the logic altogether, please see edited description in the PR Summary.

PaulHigin (approved):
LGTM

Issue #515, opened 2021-10-19
Install-PSResource: implement -RequiredResource and -RequiredResourceFile 


SydneyhSmith:
For Install-PSResource we still need to implement -RequiredResource and -RequiredResourceFile parameters

StevenBucher98:
Adding additional details from [blog posts](https://devblogs.microsoft.com/powershell/powershellget-3-0-preview-12-release/):

Install-PSResource will accept a path to a psd1 or JSON file (using -RequiredResourceFile), or a hash table or JSON (using -RequiredResource) where the key is the module name and the value is either the required version specified using Nuget version range syntax or a hash table where repository is set to the URL of the repository and version contains the Nuget version range syntax.
```powershell
Install-PSResource -RequiredResource @{
  Configuration = "[1.3.1,2.0]"

  Pester        = @{
    version = "[4.4.2,4.7.0]"
    repository = "https://www.powershellgallery.com"
    credential = $cred
    allowPrerelease = $true
  }
}
```
In this case the modules named “Configuration”, and “Pester” will be installed. The JSON format will be the same as if this hashtable is passed to ConvertTo-Json:

```JSON
{
  "Configuration": "[1.3.1,2.0]",

  "Pester": {
    "version": "[4.4.2,4.7.0]",
    "credential": null,
    "repository": "https://www.powershellgallery.com",
    "allowPrerelease": true
  }
}
```

Issue #146, opened 2020-05-19
[beta3] Install-PSResource -RequiredResource


JustinGrote:
Trying the examples from the RFC, I can't get any to work, the matching doesn't seem to be operating. Am I doing the syntax incorrectly?

![image](https://user-images.githubusercontent.com/15258962/82367421-a2ad5980-99c8-11ea-8004-a9af56c1f03f.png)

alerickson:
Sydney and I just [updated the RFC](https://github.com/PowerShell/PowerShell-RFC/blob/b118eca8edb56a4988dd1072b208cd8dbbfebe76/2-Draft-Accepted/RFCxxxx-PowerShellGet-3.0.md#dependencies-and-version-management) yesterday, try using single quotes instead of double

JustinGrote:
Doesn't json require double quotes?
https://www.json.org/json-en.html

Also I'm doing this by passing a hashtable to requiredresource, so singlequotes shouldn't matter (it's just how it evaluates).

I see that you're accepting RequiredResource as a [Object], are you expecting to support multiple object types like a string, etc.?

JustinGrote:
I did finally get it to work as a json string (using double quotes):
`install-psresource -verbose -RequiredResource '{"Configuration": "[1.3.1,2.0)"}'`

However saving that same string to a file did not work, note that it said the parameter I used was `RequiredResource` when I clearly used `RequiredResourceFile`
![image](https://user-images.githubusercontent.com/15258962/82376872-a2688a80-99d7-11ea-9b9b-b90022a2a814.png)

So the scenarios so far are:

- [ ] RequiredResource as HashTable
- [x] RequiredResource as JSON String
- [ ] RequiredResourceFile as Json File


Hashtable-to-string workaround also works:
`Install-PSResource -RequiredResource ([String](@{Configuration='[1.3.1,2.0)'} | ConvertTo-Json -Compress)) -verbose`

SydneyhSmith:
@JustinGrote thanks again for the detailed feedback, after chatting with Amber we have worked through the scenarios and hope to get this fixed up, in addition we discussed the need for better error messaging about what may be wrong with the format provided by the user.

PR #592, opened 2022-01-14
Implement RequiredResourceFile and RequiredResource parameters

alerickson:
# PR Summary
This is the implementation for the -RequiredResourceFile and -RequiredResource parameters as specified in the [RFC](https://github.com/PowerShell/PowerShell-RFC/blob/17ecbf7c091dca0c00bad7eb5054491d9d5a0639/2-Draft-Accepted/RFCxxxx-PowerShellGet-3.0.md#dependencies-and-version-management).

One note here:  the RFC specifies that -RequiredResourceFile take in a .psd1.  This is not implemented yet as the specs defining what exactly should be done here is not yet complete.  The idea behind accepting a .psd1 file is to mimic the behavior of [PSDepend](https://www.powershellgallery.com/packages/PSDepend/0.3.2).  Again, because there's still a lot of unknowns here, I've chosen to leave that out of this PR.   

## PR Context

This resolves issues https://github.com/PowerShell/PowerShellGet/issues/515, https://github.com/PowerShell/PowerShellGet/issues/275, and https://github.com/PowerShell/PowerShellGet/issues/146.

## PR Checklist

    - [x] None
    - **OR**
        - [ ] Issue filed: 
- **User-facing changes**
    - [ ] Not Applicable
    - **OR**
        - [ ] Issue filed: 
    - **OR**
- **Tooling**
    - **OR**

### DIFF
--- a/src/code/InstallPSResource.cs
+++ b/src/code/InstallPSResource.cs
@@ -10,6 +10,7 @@ using System.Collections.Generic;
 using System.IO;
 using System.Linq;
 using System.Management.Automation;
+using Microsoft.PowerShell.Commands;
 
 using Dbg = System.Diagnostics.Debug;
 
@@ -271,8 +272,7 @@ namespace Microsoft.PowerShell.PowerShellGet.Cmdlets
                     Hashtable pkgsInJsonFile = null;
                     try
                     {
-                        pkgsInJsonFile = JsonConvert.DeserializeObject<Hashtable>(requiredResourceFileStream, new JsonSerializerSettings { MaxDepth = 6 });
-
+                        pkgsInJsonFile = Utils.ConvertJsonToHashtable(this, requiredResourceFileStream);
                     }
                     catch (Exception)
                     {
@@ -303,7 +303,7 @@ namespace Microsoft.PowerShell.PowerShellGet.Cmdlets
                         Hashtable pkgsHash = null;
                         try
                         {
-                            pkgsHash = JsonConvert.DeserializeObject<Hashtable>(_requiredResourceJson, new JsonSerializerSettings { MaxDepth = 6 });
+                            pkgsHash = Utils.ConvertJsonToHashtable(this, _requiredResourceJson);
                         }
                         catch (Exception)
                         {
@@ -353,21 +353,8 @@ namespace Microsoft.PowerShell.PowerShellGet.Cmdlets
             {
                 var pkgParamInfo = reqResourceHash[pkgName];
 
-                // Input format was json, and not a hashtable
+                // Format should now be a hashtable, whether the original input format was json or hashtable
                 if (!(pkgParamInfo is Hashtable pkgInstallInfo))
-                {
-                    pkgInstallInfo = new Hashtable { };
-
-                    // Input format was JSON
-                    var pkgParamJTokens = pkgParamInfo as Newtonsoft.Json.Linq.JToken;
-                    foreach (Newtonsoft.Json.Linq.JProperty val in pkgParamJTokens)
-                    {
-                        pkgInstallInfo[val.Name] = val.Value.ToString();
-                    }
-                }
-                // Otherwise, the input format is a hashtable
-
-                if (pkgInstallInfo == null)
                 {
                     return;
                 }
--- a/src/code/InstallPSResource.cs
+++ b/src/code/InstallPSResource.cs
@@ -2,9 +2,12 @@
 // Licensed under the MIT License.
 
 using Microsoft.PowerShell.PowerShellGet.UtilClasses;
+using Newtonsoft.Json;
 using NuGet.Versioning;
 using System;
+using System.Collections;
 using System.Collections.Generic;
+using System.IO;
 using System.Linq;
 using System.Management.Automation;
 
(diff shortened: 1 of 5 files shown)


src/code/InstallPSResource.cs
```diff
@@ -113,6 +116,62 @@ class InstallPSResource : PSCmdlet
...
         public PSResourceInfo InputObject { get; set; }
 
+        /// <summary>
+        /// Installs resources based on input from a JSON file.
+        /// </summary>
+        [Parameter(ParameterSetName = RequiredResourceFileParameterSet)]
+        [ValidateNotNullOrEmpty]
+        public String RequiredResourceFile
+        {
+            get {
```
```suggestion
            get
            {
```

src/code/InstallPSResource.cs
```diff
@@ -113,6 +116,62 @@ class InstallPSResource : PSCmdlet
...
+            {
+                string resolvedPath = string.Empty;
+                if (!string.IsNullOrWhiteSpace(value))
+                {
+                    resolvedPath = SessionState.Path.GetResolvedPSPathFromPSPath(value).First().Path;
+                }
+
+                if (!File.Exists(resolvedPath))
+                {
+                    var exMessage = String.Format("The RequiredResourceFile does not exist.  Please try specifying a path to a valid .json or .psd1 file");
```
PaulHigin: We don't currently support .psd1 files, correct?

src/code/InstallPSResource.cs
```diff
@@ -113,6 +116,62 @@ class InstallPSResource : PSCmdlet
...
+            }
+        }
+
+        /// <summary>
+        ///  Installs resources in a hashtable or JSON string format.
+        /// </summary>
+        [Parameter(ParameterSetName = RequiredResourceParameterSet)]
+        public Object RequiredResource  // takes either string (json) or hashtable
+        {
+            get { return _requiredResourceHash != null ? (Object)_requiredResourceHash : (Object)_requiredResourceJson; }
```
PaulHigin: Cast to 'Object' should not be needed here.

src/code/InstallPSResource.cs
```diff
@@ -113,6 +116,62 @@ class InstallPSResource : PSCmdlet
...
+
+            set
+            {
+                if (value.GetType().Name.Equals("String"))
```
PaulHigin: Suggest:
```csharp
if (value is String jsonResource)
{
    _requiredResourceJson = jsonResource;
}
else if (value is Hashtable hashResource)
{
    _requiredResourceHash = hashResource;
}
else ...
```

src/code/InstallPSResource.cs
```diff
@@ -177,37 +242,190 @@ protected override void ProcessRecord()
...
+                                "credential": null
+                              }
+                            }
+                        */
+                                              
+                        Hashtable pkgsInJson = null;
+                        try
+                        {
+                            //pkgsInJson = JsonConvert.DeserializeObject<Dictionary<string, InstallPkgParams>>(_requiredResourceJson, new JsonSerializerSettings { MaxDepth = 6 });
+                            pkgsInJson = JsonConvert.DeserializeObject<Hashtable>(_requiredResourceJson, new JsonSerializerSettings { MaxDepth = 6 });
```
PaulHigin: Rather than taking a dependency on `Newtonsoft.Json`, I feel we should just use PowerShell's `ConvertFrom-Json`.  This can be a Utility function, like: https://github.com/PowerShell/PowerShellGet/blob/1242151789400fc6cda38cb16e68f8a6451038f7/src/code/Utils.cs#L741

We should look at removing the `Newtonsoft.Json.dll` file from the binaries we ship if we don't need it.

src/code/InstallPSResource.cs
```diff
@@ -177,37 +242,190 @@ protected override void ProcessRecord()
...
+                    string requiredResourceFileStream = string.Empty;
+                    using (StreamReader sr = new StreamReader(_requiredResourceFile))
+                    {
+                        requiredResourceFileStream = sr.ReadToEnd();
+                    }
+                    
+                    Hashtable pkgsInJsonFile = null;
+                    try
+                    {
+                        pkgsInJsonFile = JsonConvert.DeserializeObject<Hashtable>(requiredResourceFileStream, new JsonSerializerSettings { MaxDepth = 6 });
```
PaulHigin: Same comment to not use `JsonConvert.DeserializeObject`.

src/code/InstallPSResource.cs
PaulHigin: This converts json to Hashtable, right?  So variable name should be `pkgsHash` or something.

src/code/InstallPSResource.cs
```diff
@@ -177,37 +242,190 @@ protected override void ProcessRecord()
...
+
+                if (pkgParamInfo is Hashtable)
+                {
+                    // Input format was Hashtable
+                    pkgInstallInfo = pkgParamInfo as Hashtable;
+                }
+                else
+                {
+                    // Input format was JSON
+                    var pkgParamJTokens = pkgParamInfo as Newtonsoft.Json.Linq.JToken;
```
PaulHigin: Not sure why this is needed.  But if some Hashtable values return json, please use `ConvertFrom-Json` utility to convert to hash table.

src/code/InstallPSResource.cs
```diff
@@ -177,37 +242,190 @@ protected override void ProcessRecord()
...
+        private void RequiredResourceHelper(Hashtable reqResourceHash)
+        {
+            var pkgNames = reqResourceHash.Keys;
+
+            foreach (string pkgName in pkgNames)
+            {
+                var pkgParamInfo = reqResourceHash[pkgName];
+                Hashtable pkgInstallInfo = new Hashtable { };
+
+                if (pkgParamInfo is Hashtable)
```
```csharp
if (pkgParameInfo is Hashtable pkgInstallHash)
{
    ...
}
```

src/code/InstallPkgParams.cs
```diff
@@ -0,0 +1,100 @@
...
+    public string Repository { get; set; }
+    public bool AcceptLicense { get; set; } 
+    public bool Prerelease { get; set; }
+    public ScopeType Scope { get; set; }
+    public bool Quiet { get; set; }
+    public bool Reinstall { get; set; }
+    public bool Force { get; set; }
+    public bool TrustRepository { get; set; }
+    public bool NoClobber { get; set; }
+	public bool SkipDependencyCheck { get; set; }
```
PaulHigin:
```suggestion
public bool SkipDependencyCheck { get; set; }

```
This is using tab character for indent.  Please be sure to conform to tab == whitespace convention so our source file formats are consistent.

src/code/InstallPkgParams.cs
```diff
@@ -0,0 +1,100 @@
...
+
+            case "noclobber":
+                bool.TryParse(propertyValue, out bool noClobberTmp);
+                NoClobber = noClobberTmp;
+                break;
+
+            case "skipdependencycheck":
+                bool.TryParse(propertyValue, out bool skipDependencyCheckTmp);
+                SkipDependencyCheck = skipDependencyCheckTmp;
+                break;
```
PaulHigin, line 97: Please add `default:` case.  We may want to assert if an unknown name is unexpected.

PaulHigin (changes requested):

src/code/InstallPSResource.cs
alerickson: I'll try that, this cast was needed to extract values out of the object.  But probably moot since this JSON conversion will be changed anyway.

src/code/InstallPSResource.cs
alerickson: Yup, that makes much more sense, I'll do that and try to completely remove Newtonsoft

src/code/InstallPSResource.cs
alerickson: Correct, I'll remove that section

PaulHigin (approved):
LGTM

Issue #594, opened 2022-01-19
-RequiredResourceFile for .psd1


SydneyhSmith:
### Summary of the new feature / enhancement

`Install-PSResource` can accept a path to a psd1 or json file (using -RequiredResourceFile), or a hashtable or json (using -RequiredResource) where the key is the module name and the value is either the required version specified using Nuget version range syntax or a hash table where repository is set to the URL of the repository and version contains the Nuget version range syntax.

This issue is specifically for the `-RequiredResourceFile` parameter using a psd1 file. 

## Scenarios 

- Incorporate the popular PSDepends functionality 
- Clear module configuration for PowerShell projects
- Configurable abstraction for module management
- Parity with other languages (like requirements.txt for Python)

## User Experience
```
Install-PSResource -RequiredResource @{
  'Configuration' = '[1.3.1,2.0)'
  'Pester'        = @{
    version = '[4.4.2,4.7.0]'
    repository = 'https://www.powershellgallery.com'
    credential = $cred
    allowPrerelease = $true
  }
}
In this case the modules named "Configuration", and "Pester" will be installed. The json format will be the same as if this hashtable is passed to ConvertTo-Json:

{
  'Configuration': '[1.3.1,2.0)',
  'Pester': {
    'version': '[4.4.2,4.7.0]',
    'credential': null,
    'repository': 'https://www.powershellgallery.com',
    'prerelease': true
  }
}
```
The older System.Version four part version type will be supported to retain compatibility with existing published modules using that format.

Declared dependencies are searched and installed using the same trust algorithm as for Install-PSResource.

PR #610, opened 2022-02-12
Allow .psd1 to be passed into -RequiredResourceFile

alerickson:
# PR Summary

Adding functionality to allow .psd1 files to be passed into -RequiredResourceFile in Install-PSResource. 

## PR Context

Addresses issue: https://github.com/PowerShell/PowerShellGet/issues/594
## PR Checklist

    - [x] None
    - **OR**
        - [ ] Issue filed: 
- **User-facing changes**
    - [x] Not Applicable
    - **OR**
        - [ ] Issue filed: 
    - **OR**
- **Tooling**
    - **OR**


src/code/PublishPSResource.cs
```diff
@@ -470,7 +470,7 @@ private bool IsValidModuleManifest(string moduleManifestPath)
             if (!isScript)
             {
                 // Parse the module manifest and *replace* the passed-in metadata with the module manifest metadata.
-                if (!Utils.TryParseModuleManifest(
+                if (!Utils.TryParsePsd1(
```
PaulHigin: Suggest name: `TryParsePSDataFile`.

src/code/InstallPSResource.cs
```diff
@@ -138,13 +138,22 @@ public String RequiredResourceFile
...
                 {
-                    var exMessage = String.Format("The RequiredResourceFile does not exist.  Please try specifying a path to a valid .json file");
+                    var exMessage = String.Format("The RequiredResourceFile does not exist.  Please try specifying a path to a valid .json or .psd1 file");
                     var ex = new ArgumentException(exMessage);
                     var RequiredResourceFileDoesNotExist = new ErrorRecord(ex, "RequiredResourceFileDoesNotExist", ErrorCategory.ObjectNotFound, null);
 
                     ThrowTerminatingError(RequiredResourceFileDoesNotExist);
                 }
 
+                if (!resolvedPath.EndsWith(".json") && !resolvedPath.EndsWith(".psd1"))
```
PaulHigin: You could set a private enum here that determines what kind of file:
```csharp
private enum ResourceFileType
{
    UnknownFile = 0,
    JsonFile,
    PSDataFile
};
```
And then use that value instead of checking for file extension each time.

PaulHigin: String compare should be case insensitive:
```csharp
resolvedPath.EndsWith(".json", StringComparison.InvariantCultureIgnoreCase);
```

PaulHigin (changes requested):

src/code/PublishPSResource.cs
anamnavi: I like this idea as well

test/InstallPSResource.Tests.ps1
```diff
@@ -381,10 +382,28 @@ Describe 'Test Install-PSResource for Module' {
           $res3.Version | Should -Be "0.0.5"
     }
 
+    It "Install modules using -RequiredResourceFile with PSD1 file" {
+        $rrFilePSD1 = ".\$RequiredResourcePSD1FileName"
```
anamnavi: this way it doesn't matter where you run tests from, it should be able to find this required resource psd1 file unlike with the relative path.
```suggestion
        $rrFilePSD1 = Join-Path -Path $psscriptroot -ChildPath "test" -AdditionalChildPath ".\$RequiredResourcePSD1FileName"

```

test/TestRequiredResourceFile.psd1
```diff
@@ -0,0 +1,15 @@
...
+   
+      TestModulePrerelease = @{
+        version = "[0.0.0,0.0.5]"
+        repository = "PoshTestGallery"
+        prerelease = "true"
+      }
+   
+    TestModule99 = @{
+    }
+}
```
anamnavi: nitpick, but add newline

anamnavi (changes requested):
Some minor suggestions, and I also agree with Paul's idea of a private enum for type of file though

src/code/InstallPSResource.cs
alerickson: How exactly would the private enum work?  Right now I have variables in the class instead of the hardcoded extension names, but I'm a little confused as to how the enum type works to categorize ".json" and ".psd1"

PaulHigin: I was thinking of something like:

```csharp
ResourceFileType fileType;
if (resolvedPath.EndsWith(".json", StringComparison.InvariantCultureIgnoreCase)
{
    fileType = ResourceFileType.JsonFile;
}
else if (resolvePath.EndsWith(".psd1", StringComparison.InvariantCultureIgnoreCase)
{
    fileType = ResourceFileType.PSDataFile;
}
else
{
    // Or maybe just throw here, if no further processing can be done.
    fileType = ResourceFileType.UnknownFile;
}

// No need to check path anymore, if further processing needs to check file type again.

```

alerickson: Ah, yes.  This makes sense, I can add this

src/code/InstallPSResource.cs
```diff
@@ -185,20 +194,29 @@ public String RequiredResourceFile
 
         #endregion
 
+        #region Enums
```
PaulHigin, line 197: Please add white space after 'Enums'

src/code/InstallPSResource.cs
```diff
@@ -185,20 +194,29 @@ public String RequiredResourceFile
...
+        public enum ResourceFileType
```
PaulHigin: Enum should be `private` since it is not useful for a user and is needed internally only.

PaulHigin (approved):
Looks good after some minor changes.

test/TestRequiredResourceFile.psd1
anamnavi: still needs to be added if you can, thanks!

anamnavi (approved):
LGTM, just one minor feedback

Issue #519, opened 2021-10-19
Install-PSResource: implement -SkipPublisherCheck


SydneyhSmith:
For Install-PSResource we still need to implement -SkipPublisherCheck

PR #632, opened 2022-04-22
Add authenticode validation to Install-PSResource

alerickson:
# PR Summary
Adding -AuthenticodeCheck parameter to `Install-PSResource` to validate signed files and catalog files at install, save and update time.

Still needed:
- [x] Tests 
- [x] Proper error messaging on failure
- [x] Discussion as to whether the default behavior should be to validate signatures (ie -SkipPublisherCheck flag) or to not validate signatures (ie opt in via -PublisherCheck)
- [x]  Discussion as to whether Update-PSResource and/or Save-PSResource should also have publisher validations

## PR Context
See issues #519, #489

## PR Checklist
    - [x] None
    - **OR**
        - [ ] Issue filed: 
- **User-facing changes**
    - [ ] Not Applicable
    - **OR**
        - [ ] Issue filed: 
    - **OR**
- **Tooling**
    - **OR**


src/code/InstallPSResource.cs
```diff
@@ -103,6 +103,12 @@ class InstallPSResource : PSCmdlet
         /// </summary>
         [Parameter]
         public SwitchParameter SkipDependencyCheck { get; set; }
+        
+        /// <summary>
+        /// Skips the check for package validation.
```
anamnavi: Should this message be updated, since we're performing authentication check not really package validation as a whole?
```suggestion
        /// Check validation for signed and catalog files
```

src/code/SavePSResource.cs
```diff
@@ -131,6 +131,12 @@ public string Path
         [Parameter]
         public SwitchParameter SkipDependencyCheck { get; set; }
 
+        /// <summary>
+        /// Skips the check for package validation.
```
anamnavi:
```suggestion
        /// Check validation for signed and catalog files

```

src/code/UpdatePSResource.cs
```diff
@@ -110,6 +110,12 @@ public sealed class UpdatePSResource : PSCmdlet
         [Parameter]
         public SwitchParameter SkipDependencyCheck { get; set; }
 
+        /// <summary>
+        /// Skips the check for package validation.
```
anamnavi:
```suggestion
        /// Check validation for signed and catalog files

```

src/code/InstallHelper.cs
```diff
@@ -1,898 +1,1229 @@
...
+
+                if (!sourceTrusted && !yesToAll)
+                {
+                    continue;
+                }
+
+                _cmdletPassedIn.WriteVerbose("Untrusted repository accepted as trusted source.");
+
+                // If it can't find the pkg in one repository, it'll look for it in the next repo in the list
+                var isLocalRepo = repo.Uri.AbsoluteUri.StartsWith(Uri.UriSchemeFile + Uri.SchemeDelimiter, StringComparison.OrdinalIgnoreCase);
```
anamnavi, line 192: Could we just do this instead? Or does it end up including cases (local server based paths) that we don't want?
```suggestion
                var isLocalRepo = repo.Uri.Scheme == Uri.UriSchemeFile;
```

src/code/InstallHelper.cs
```diff
@@ -1,898 +1,1229 @@
...
+                    var exMessage = String.Format("The catalog file '{0}' is invalid.", pkgName + ".cat");
+                    var ex = new ArgumentException(exMessage);
+
+                    errorRecord = new ErrorRecord(ex, "TestFileCatalogError", ErrorCategory.InvalidResult, null);
+                    return false;
+                }
+            }
+
+
+            // TODO: validate all files (need to figure out what all files are (ie what files shouldn't be validated) : list of extensions
```
anamnavi: we can remove this TODO comment now I think

anamnavi (approved):
LGTM, just some questions/suggestions

src/code/InstallHelper.cs
```diff
@@ -1,898 +1,1227 @@
...
-            }
-        }
-
-        #endregion
-    }
-}
+        public const string PSScriptFileExt = ".ps1";
+        private const string MsgRepositoryNotTrusted = "Untrusted repository";
+        private const string MsgInstallUntrustedPackage = "You are installing the modules from an untrusted repository. If you trust this repository, change its Trusted value by running the Set-PSResourceRepository cmdlet. Are you sure you want to install the PSresource from '{0}' ?";
+        private readonly string[] certStoreLocations = { "cert:\\LocalMachine\\Root", "cert:\\LocalMachine\\AuthRoot", "cert:\\CurrentUser\\Root", "cert:\\CurrentUser\\AuthRoot" };
```
```suggestion
        private readonly string[] CertStoreLocations = { "cert:\\LocalMachine\\Root", "cert:\\LocalMachine\\AuthRoot", "cert:\\CurrentUser\\Root", "cert:\\CurrentUser\\AuthRoot" };

```

src/code/InstallHelper.cs
```diff
@@ -1,898 +1,1227 @@
...
+                chain.Build(authenticodeSignature.SignerCertificate);
+
+                foreach (X509ChainElement element in chain.ChainElements)
+                {
+                    foreach (var certStoreLocation in certStoreLocations)
+                    {
+                        var results = PowerShellInvoker.InvokeScriptWithHost<object>(
+                                        cmdlet: _cmdletPassedIn,
+                                        script: @"
+                                            param ([string] $certStoreLocation )
```
PaulHigin: `element` parameter is not being passed in as a parameter argument, so $element is always null.

src/code/InstallHelper.cs
```diff
@@ -1,898 +1,1227 @@
...
+                            _cmdletPassedIn.WriteVerbose(String.Format("Successfully deleted '{0}'", tempInstallPath));
+                        }
+                    }
+                }
+            }
+
+            return pkgsSuccessfullyInstalled;
+        }
+
+        private bool CheckAuthenticodeSignature(string pkgName, string tempDirNameVersion, VersionRange versionRange, List<string> pathsToSearch, string installPath, out ErrorRecord errorRecord)
```
PaulHigin: All of the signature check logic should not be in 'InstallHelper.cs', but should be static methods in 'Utils.cs', in its own 'Authenticode Signauture' code region.

src/code/InstallHelper.cs
```diff
@@ -1,898 +1,1227 @@
...
+            string catalogFilePath = Path.Combine(tempDirNameVersion, pkgName + ".cat");
+            if (File.Exists(catalogFilePath))
+            {
+                // Run catalog validation
+                Collection<PSObject> TestFileCatalogResult = new Collection<PSObject>();
+                string moduleBasePath = tempDirNameVersion;
+                try
+                {
+                    // By default "Test-FileCatalog will look through all files in the provided directory, -FilesToSkip allows us to ignore specific files
+                    TestFileCatalogResult = _cmdletPassedIn.InvokeCommand.InvokeScript(
```
PaulHigin: All script invocations should be wrapped in a static method in the 'Utils.cs'.  If very large we could have a 'AuthenticodeUtils.cs' file.

src/code/InstallHelper.cs
```diff
@@ -1,898 +1,1227 @@
...
+
+                    errorRecord = new ErrorRecord(ex, "TestFileCatalogError", ErrorCategory.InvalidResult, null);
+                    return false;
+                }
+            }
+
+            Collection<PSObject> authenticodeSignature = new Collection<PSObject>();
+            try
+            {
+                string[] listOfExtensions = { "*.ps1", "*.psd1", "*.psm1", "*.mof", "*.cat", "*.ps1xml" };
```
PaulHigin: We can't gather all file signatures in a single collection, because we need to ensure each signable file is actually signed with a valid signature.

src/code/InstallHelper.cs
```diff
@@ -1,898 +1,1227 @@
...
+                                new ErrorRecord(
+                                    new PSInvalidOperationException(
+                                        message: $"Install-PSResource encountered an error while authenticating certificate for \"{pkgName}\" from certificate store \"{certStoreLocation}\".",
+                                        innerException: terminatingError),
+                                    "InstallPSResourceCannotReadCertFromStore",
+                                    ErrorCategory.InvalidResult,
+                                    _cmdletPassedIn));
+                        }
+
+                        X509Certificate2 rootCertificateAuthority = results.Any() ? (X509Certificate2)results.FirstOrDefault() : null;
```
PaulHigin: Don't use LINQ, and instead use actual properties on objects.

Also, this should be wrapped in a utility method that returns expected object/type.

src/code/InstallHelper.cs
```diff
@@ -1,898 +1,1227 @@
...
+                                            Microsoft.PowerShell.Management\Get-ChildItem -Path $certStoreLocation |
+                                                Microsoft.PowerShell.Core\Where-Object { ($_.Subject -eq $element.Certificate.Subject) -and ($_.thumbprint -eq $element.Certificate.Thumbprint) }
+                                            ",
+                                        args: new object[] { certStoreLocation, element },
+                                        out Exception terminatingError);
+
+
+                        if (terminatingError != null)
+                        {
+                            ThrowTerminatingError(
```
PaulHigin: Cmdlet exception should be thrown at higher layer.

src/code/InstallHelper.cs
```diff
@@ -1,898 +1,1227 @@
...
+                        File.Delete(Path.Combine(finalModuleVersionDir, pkgInfo.Name + PSScriptFileExt));
+                    }
+                }
+
+                _cmdletPassedIn.WriteVerbose(string.Format("Moving '{0}' to '{1}'", scriptPath, Path.Combine(finalModuleVersionDir, pkgInfo.Name + PSScriptFileExt)));
+                Utils.MoveFiles(scriptPath, Path.Combine(finalModuleVersionDir, pkgInfo.Name + PSScriptFileExt));
+            }
+        }
+
+        private Hashtable GetAuthenticodePublisher(Signature authenticodeSignature, string pkgName)
```
PaulHigin: This should be a static method.  Also, this performs an exhaustive search and is expensive.  Most files will have the same signature certificate, so it should be possible to cache results for a SignerCertificate based on its thumbprint.

src/code/InstallHelper.cs
```diff
@@ -1,898 +1,1227 @@
...
+                foreach (X509ChainElement element in chain.ChainElements)
+                {
+                    foreach (var certStoreLocation in certStoreLocations)
+                    {
+                        var results = PowerShellInvoker.InvokeScriptWithHost<object>(
+                                        cmdlet: _cmdletPassedIn,
+                                        script: @"
+                                            param ([string] $certStoreLocation )
+
+                                            Microsoft.PowerShell.Management\Get-ChildItem -Path $certStoreLocation |
```
PaulHigin: This is an expensive search and results should be cached.  Also include comments to make clear what the search does and what the results mean.

src/code/InstallHelper.cs
```diff
@@ -1,898 +1,1227 @@
...
+                if (safex509ChainHandle != null)
+                {
+                    safex509ChainHandle.Dispose();
+                }
+            }
+
+            return isMicrosoftCert;
+        }
+
+        public static bool IsMicrosoftCertificateHelper(SafeX509ChainHandle pChainContext)
```
PaulHigin: Don't use the word 'helper' in utility methods.

src/code/InstallHelper.cs
```diff
@@ -1,898 +1,1227 @@
...
+                }
+            }
+
+            return publisherInfo;
+        }
+
+        private bool IsMicrosoftCert(Signature authenticodeSignature)
+        {
+            bool isMicrosoftCert = false;
+            if (authenticodeSignature.SignerCertificate != null)
```
PaulHigin: Maybe this is the place to add the cache.  Once a signature certificate has been validated it can be added to a cache based on its thumbprint.

src/code/InstallHelper.cs
```diff
@@ -1,898 +1,1227 @@
...
+
+        [DllImport("Crypt32.dll", CharSet = CharSet.Auto, SetLastError = true)]
+        public extern static
+        bool CertVerifyCertificateChainPolicy(
+            IntPtr pszPolicyOID,
+            SafeX509ChainHandle pChainContext,
+            ref CERT_CHAIN_POLICY_PARA pPolicyPara,
+            ref CERT_CHAIN_POLICY_STATUS pPolicyStatus);
+
+
```
PaulHigin: Remove unneeded white space.

PaulHigin (changes requested):
A good start, but some logic seems broken and all of the signature check functions should be separated from this file, and added as static utility methods either in Utils.cs (in separate code region) or in a new AuthenticodeUtils.cs source file.  Also we need to make use of cache to prevent repeatedly searching the same paths and validating the same signature certificate.

src/code/InstallHelper.cs
alerickson: I addressed the first comment, but not sure if I fully addressed the second comment.

src/code/InstallHelper.cs
alerickson: Lines 721-732 will handle this.  Basically, if any file that ends with the one of the extensions in the array does not have a valid signature, it will still return a Signature object, but the 'SignatureStatus' will not be valid (ie 'NotSigned' or 'UnknownError')

src/code/InstallHelper.cs
alerickson: Added a cache for both validating if a cert is a Microsoft cert and to find the publisher details

src/code/InstallHelper.cs
alerickson: I haven't added comments yet, but will do that

src/code/InstallHelper.cs
PaulHigin: We should also include in the error which file failed the Authenticode signature check and how it failed (no signed or invalid).

src/code/InstallHelper.cs
PaulHigin: This is not needed anymore since we are not validating certs, but just performing Authenticode signature checks.

src/code/InstallHelper.cs
PaulHigin: I don't think this is needed anymore as we are not doing cert checks.

PaulHigin (commented):
LGTM

PR #661, opened 2022-05-26
Psresourcerepository file bug

anamnavi:
# PR Summary

There was a bug where `Url` was changed to `Uri` in RepositorySettings.cs, so when using the PSResourceRepository.xml file for read/update/add/remove operations it would check for a `Uri` attribute, but due to all prior versions of PowerShellGet having `Url` in the PSResourceRepository.xml file, the `Uri` attribute couldn't be found and referencing it was causing a null pointer exception to be thrown.

This PR:
- fixes that bug and reverts back to using Url as the default for read/write operations to PSResourceRepository.xml file.
- we check for both `Url` and `Uri` when read/writing info for a repository to the xml file.
- also add error handling to check that other expected attributes, i.e Name, Priority, Trusted, exist before referencing them and provide proper error handling for those cases too.

## PR Checklist

    - [ ] None
    - **OR**
        - [ ] Issue filed: 
- **User-facing changes**
    - [ ] Not Applicable
    - **OR**
        - [ ] Issue filed: 
    - **OR**
- **Tooling**
    - **OR**

### DIFF
--- a/src/code/RepositorySettings.cs
+++ b/src/code/RepositorySettings.cs
@@ -98,7 +98,7 @@ namespace Microsoft.PowerShell.PowerShellGet.UtilClasses
                 XElement newElement = new XElement(
                     "Repository",
                     new XAttribute("Name", repoName),
-                    new XAttribute("Uri", repoUri),
+                    new XAttribute("Url", repoUri),
                     new XAttribute("Priority", repoPriority),
                     new XAttribute("Trusted", repoTrusted)
                     );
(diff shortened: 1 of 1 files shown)


src/code/RepositorySettings.cs
```diff
@@ -139,17 +139,60 @@ public static PSRepositoryInfo Update(string repoName, Uri repoUri, int repoPrio
...
 
+                // Check that repository node we are attempting to update has all required attributes: Name, Url (or Uri), Priority, Trusted.
+                // Name attribute is already checked for in FindRepositoryElement()
+
+                if (node.Attribute("Priority") == null)
+                {
+                    throw new ArgumentException("Repository element does not contain neccessary 'Priority' attribute, in file located at path: {0}. Fix this in your file and run again.", FullRepositoryPath);
+                }
+                
+                if (node.Attribute("Trusted") == null)
```
alerickson, line 150: I think we should also add this null check for 'name' as well just in case we change something in the future or a user modifies their own xml

src/code/RepositorySettings.cs
```diff
@@ -139,17 +139,60 @@ public static PSRepositoryInfo Update(string repoName, Uri repoUri, int repoPrio
...
+                {
+                    throw new ArgumentException("Repository element does not contain neccessary 'Trusted' attribute, in file located at path: {0}. Fix this in your file and run again.", FullRepositoryPath);
+                }                
+
+                bool urlAttributeExists = node.Attribute("Url") != null;
+                bool uriAttributeExists = node.Attribute("Uri") != null;
```
alerickson, line 156: I think we can get rid of the 'Uri' checks and just check for 'Url'.  We can just let users know if they encounter the null object error that they can reset their xml (presumably this is a very small number of people anyway)

src/code/RepositorySettings.cs
```diff
@@ -257,9 +294,32 @@ public static List<PSRepositoryInfo> Remove(string[] repoNames, out string[] err
                 {
                     repoCredentialInfo = new PSCredentialInfo(node.Attribute("VaultName").Value, node.Attribute("SecretName").Value);
                 }
+
+                if (node.Attribute("Priority") == null)
```
alerickson, line 298: Same comments here as above

alerickson (changes requested):

src/code/RepositorySettings.cs
```diff
@@ -139,17 +139,60 @@ public static PSRepositoryInfo Update(string repoName, Uri repoUri, int repoPrio
                     throw new ArgumentException("Cannot find the repository because it does not exist. Try registering the repository using 'Register-PSResourceRepository'");
                 }
 
+                // Check that repository node we are attempting to update has all required attributes: Name, Url (or Uri), Priority, Trusted.
+                // Name attribute is already checked for in FindRepositoryElement()
+
+                if (node.Attribute("Priority") == null)
```
PaulHigin, line 145: I think we can simplify this code by first assigning all expected xml attributes to local variables, and not perform look ups multiple times for each attribute.  Then we can check the local variables for null values.

src/code/RepositorySettings.cs
```diff
@@ -139,17 +139,60 @@ public static PSRepositoryInfo Update(string repoName, Uri repoUri, int repoPrio
...
+                {
+                    throw new ArgumentException("Repository element does not contain neccessary 'Priority' attribute, in file located at path: {0}. Fix this in your file and run again.", FullRepositoryPath);
```
PaulHigin, line 147: We should have a common message for any missing repository values that are required.  This should be very unlikely and indicates the repository file got corrupted.  I feel the only sensible thing to do in this case is to drop the corrupted entry and report the error to the user.  We will need to re-write the repository with good data, and it is possible that the entire file is non-readable, in which case we should revert to the default state whatever it is (empty, or with PSGallery?).

Message suggestion, something like:

'The repository entry for [RepositoryName if available] is unreadable and will be dropped from the repository list.'

src/code/RepositorySettings.cs
```diff
@@ -290,17 +350,58 @@ public static List<PSRepositoryInfo> Read(string[] repoNames, out string[] error
...
 
-            if (repoNames == null || !repoNames.Any() || string.Equals(repoNames[0], "*") || repoNames[0] == null)
+            if (repoNames == null || repoNames.Length == 0 || string.Equals(repoNames[0], "*") || repoNames[0] == null)
             {
                 // Name array or single value is null so we will list all repositories registered
                 // iterate through the doc
                 foreach (XElement repo in doc.Descendants("Repository"))
                 {
-                    if (!Uri.TryCreate(repo.Attribute("Uri").Value, UriKind.Absolute, out Uri thisUri))
+                    if (repo.Attribute("Name") == null)
```
PaulHigin, line 359: There should be a utility method that 'harvests' all expected repository attributes into a data structure, and also validates the required attributes.  The returned repository data structure can then be returned and assumed to reflect a valid repository entry.

PaulHigin (changes requested):

PaulHigin (approved):
I think these changes as-is are fine for fixing the 'Url', 'Uri' problem.  Let's ignore my other comments on error message and doing remediation when file entries are unreadable, and leave those improvements for later.

PR #691, opened 2022-07-07
Convert IEnumerables to Lists for perf improvements

### LINKED ISSUE #654: [Performance improvements for find/install] Fix IEnumerable issues

Misuse of IEnumerable is causing multiple server hits in 'Find'.
- LINQ is generics and can cause multiple compiled methods calling search APIs.
- Find code always returned IEnumerable interface, causing LINQ calls to call
search APIs multiple times to manipulate IEnumerable as collections.
- Dependency search returns IEnumerable which is then used as collection and calls server X times.
retrievedPkgs = pkgMetadataResource.GetMetadataAsync(...)
if (retrievedPkgs == null || retrievedPkgs.Count() == 0) ...
foundPackagesMetadata.AddRange(retrievedPkgs.ToList());

_No response_

alerickson:
# PR Summary
Pervasive use of IEnumerables in primarily FindHelper and InstallHelper were causing some unexpected performances degradations.   Occasionally the same call was (unexpectedly) occurring multiple times due to the fact that IEnumerable does not return a collection.  This was particularly painful because it meant multiple, unnecessary server calls, which can hit the server hard, but also decreases performance of PowerShellGet.

## PR Context
Resolves #654:

- LINQ is generics and can cause multiple compiled methods calling search APIs.
- Find code always returned IEnumerable interface, causing LINQ calls to call search APIs multiple times to manipulate IEnumerable as collections. 
- Dependency search returns IEnumerable which is then used as collection and calls server X times.
retrievedPkgs = pkgMetadataResource.GetMetadataAsync(...)
if (retrievedPkgs == null || retrievedPkgs.Count() == 0) ...
foundPackagesMetadata.AddRange(retrievedPkgs.ToList());

## PR Checklist

    - [x] None
    - **OR**
        - [ ] Issue filed: 
- **User-facing changes**
    - [x] Not Applicable
    - **OR**
        - [ ] Issue filed: 
    - **OR**
- **Tooling**
    - **OR**

### DIFF
--- a/src/code/FindHelper.cs
+++ b/src/code/FindHelper.cs
@@ -384,7 +384,7 @@ namespace Microsoft.PowerShell.PowerShellGet.Cmdlets
                     yield break;
                 }
 
-                foundPackagesMetadata.AddRange(retrievedPkgs.ToList());
+                foundPackagesMetadata.AddRange(retrievedPkgs);
 
                 // _pkgsLeftToFind.Remove(pkgName);
 
@@ -427,7 +427,7 @@ namespace Microsoft.PowerShell.PowerShellGet.Cmdlets
                                 skip: SearchAsyncMaxTake,
                                 take: GalleryMax,
                                 log: NullLogger.Instance,
-                                cancellationToken: _cancellationToken).GetAwaiter().GetResult().ToList());
+                                cancellationToken: _cancellationToken).GetAwaiter().GetResult());
                     }
                 }
                 catch (HttpRequestException ex)
@@ -450,7 +450,7 @@ namespace Microsoft.PowerShell.PowerShellGet.Cmdlets
                 // perhaps validate in Find-PSResource, and use debugassert here?
                 WildcardPattern nameWildcardPattern = new WildcardPattern(pkgName, WildcardOptions.IgnoreCase);
                 foundPackagesMetadata.AddRange(wildcardPkgs.Where(
-                    p => nameWildcardPattern.IsMatch(p.Identity.Id)).ToList());
+                    p => nameWildcardPattern.IsMatch(p.Identity.Id)));
 
                 if (!_repositoryNameContainsWildcard)
                 {
--- a/src/code/FindHelper.cs
+++ b/src/code/FindHelper.cs
@@ -351,7 +351,7 @@ namespace Microsoft.PowerShell.PowerShellGet.Cmdlets
             if (!pkgName.Contains("*"))
             {
                 // case: searching for specific package name i.e "Carbon"
-                IEnumerable<IPackageSearchMetadata> retrievedPkgs = null;
+                List<IPackageSearchMetadata> retrievedPkgs = null;
                 try
                 {
                     // GetMetadataAsync() API returns all versions for a specific non-wildcard package name
@@ -362,7 +362,7 @@ namespace Microsoft.PowerShell.PowerShellGet.Cmdlets
                         includeUnlisted: false,
                         sourceCacheContext: sourceContext,
                         log: NullLogger.Instance,
-                        token: _cancellationToken).GetAwaiter().GetResult();
+                        token: _cancellationToken).GetAwaiter().GetResult().ToList();
                 }
                 catch (HttpRequestException ex)
                 {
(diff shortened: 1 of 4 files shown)


src/code/FindHelper.cs
```diff
@@ -415,17 +415,19 @@ public FindHelper(CancellationToken cancellationToken, PSCmdlet cmdletPassedIn)
...
-                            log: NullLogger.Instance,
-                            cancellationToken: _cancellationToken).GetAwaiter().GetResult());
+                        wildcardPkgs.AddRange(
+                            pkgSearchResource.SearchAsync(
+                                searchTerm: pkgName,
+                                filters: searchFilter,
+                                skip: SearchAsyncMaxTake,
+                                take: GalleryMax,
+                                log: NullLogger.Instance,
+                                cancellationToken: _cancellationToken).GetAwaiter().GetResult().ToList());
```
PaulHigin: It is not necessary to convert to List here, since the IEnumerable iterator is used only once with `wildcardPkgs.AddRange()`, which actually takes an IEnumerable type as argument.

src/code/InstallHelper.cs
```diff
@@ -192,15 +192,15 @@ public InstallHelper(PSCmdlet cmdletPassedIn)
...
+                List<PSResourceInfo> pkgsFromRepoToInstall = findHelper.FindByResourceName(
                     name: _pkgNamesToInstall.ToArray(),
                     type: ResourceType.None,
                     version: _versionRange != null ? _versionRange.OriginalString : null,
                     prerelease: _prerelease,
                     tag: null,
                     repository: new string[] { repoName },
                     credential: credential,
-                    includeDependencies: !skipDependencyCheck);
+                    includeDependencies: !skipDependencyCheck).ToList();
```
PaulHigin, line 203: If we never need IEnumerable from FindByResourceName(), maybe we should just change it to return List<> instead of IEnumerable<>.

src/code/InstallHelper.cs
```diff
@@ -222,7 +222,7 @@ public InstallHelper(PSCmdlet cmdletPassedIn)
                 // Check to see if the pkgs (including dependencies) are already installed (ie the pkg is installed and the version satisfies the version range provided via param)
                 if (!_reinstall)
                 {
-                    pkgsFromRepoToInstall = FilterByInstalledPkgs(pkgsFromRepoToInstall);
+                    pkgsFromRepoToInstall = FilterByInstalledPkgs(pkgsFromRepoToInstall).ToList();
```
PaulHigin: I see `FilterByInstalledPkgs` was changed to return List<>.  So `ToList()` is no longer needed here.

src/code/InstallHelper.cs
```diff
@@ -308,11 +309,11 @@ private IEnumerable<PSResourceInfo> FilterByInstalledPkgs(IEnumerable<PSResource
                 _pkgNamesToInstall.RemoveAll(x => x.Equals(pkg.Name, StringComparison.InvariantCultureIgnoreCase));
             }
 
-            return filteredPackages.Values.ToArray();
+            return filteredPackages.Values.ToList();
```
PaulHigin: This seems pretty inefficient.  I think it would be better to create a HashSet and then add to a final List<> only those packages that are not already installed.

src/code/InstallHelper.cs
```diff
@@ -689,11 +690,11 @@ private bool DetectClobber(string pkgName, Hashtable parsedMetadataHashtable)
             bool foundClobber = false;
             GetHelper getHelper = new GetHelper(_cmdletPassedIn);
             // selectPrereleaseOnly is false because even if Prerelease is true we want to include both stable and prerelease, never select prerelease only.
-            IEnumerable<PSResourceInfo> pkgsAlreadyInstalled = getHelper.GetPackagesFromPath(
+            List<PSResourceInfo> pkgsAlreadyInstalled = getHelper.GetPackagesFromPath(
```
PaulHigin: Since `pkgsAlreadyInstalled` is iterated over only once, converting to List<> is not needed here.

PaulHigin (changes requested):

src/code/InstallHelper.cs
alerickson: Updated this to reference a HashSet.

alerickson:
Closing this PR in favor of #728

PR #704, opened 2022-07-13
Expand acceptable Path arguments for Publish-PSResource

### LINKED ISSUE #305: Publishing with publish-psresource requires specific folder name [3.0.0-beta10]
When using publish-psresource, the folder/directory name must be the same as the directory folder, and publishing from a folder with the version name will result in an error

Example:
```
gh:\psrapid\psrAPId> publish-psresource -path C:\Users\adrian.andersson\git\psrapid\psrAPId\1.0.1\psrAPId.psd1 -Repository powershellgetbeta -Credential $pshelf.credential -APIKey $pshelf.NuGetApiKey -verbose
Publish-PSResource: Value cannot be null. (Parameter 'path')
[14:28 | 0.02]
gh:\psrapid\psrAPId> publish-psresource -path C:\Users\adrian.andersson\git\psrapid\psrAPId\1.0.1\ -Repository powershellgetbeta -Credential $pshelf.credential -APIKey $pshelf.NuGetApiKey -verbose
Publish-PSResource: No file with a .psd1 extension was found in C:\Users\adrian.andersson\git\psrapid\psrAPId\1.0.1\1.0.1.psd1.  Please specify a path to a valid modulemanifest.
```

Works fine if the folder name matches the psd1 name. Seems its parsing the folder name and searching for a manifest file using that name, which makes having a versioned release folder a bit more complicated

Bonus Oddness:
It is possible to omit the repository name and publish-psresource will not throw an error, even if there are multiple repositories with the same priority weight

alerickson:
# PR Summary
This PR expands what `Publish-PSResource` will accept as a path for a resource to be published.  
Acceptable paths:
- Modules with a file path pointing to the root directory eg:
    `C:\user\ModuleToBePublished`
- Modules with a file path pointing to the module manifest, eg:
   `C:\users\ModuleToBePublished\ModuleToBePublished.psd1`
   `C:\users\ADifferentDirectoryName\ModuleToBePublished.psd1`
- Scripts with a file path pointing to that script, eg:
   `C:\users\ScriptToBePublished.psd1`

## PR Context

Resolves #305 

## PR Checklist

    - [x] None
    - **OR**
        - [ ] Issue filed: 
- **User-facing changes**
    - [x] Not Applicable
    - **OR**
        - [ ] Issue filed: 
    - **OR**
- **Tooling**
    - **OR**


alerickson:
This PR is ready for review

src/code/PublishPSResource.cs
```diff
@@ -138,17 +138,19 @@ protected override void BeginProcessing()
...
             {
-                // condition 1: we point to a folder when publishing a module
-                // condition 2: we point to a .ps1 file directly when publishing a script, but not to .psd1 file (for publishing a module)
                 _path = resolvedPath;
             }
             else
             {
                 // unsupported file path
-                var exMessage = string.Format("Either the path to the resource to publish does not exist or is not in the correct format, for scripts point to .ps1 file and for modules point to folder containing .psd1");
+                var exMessage = string.Format("Either the path to the resource to publish does not exist or is not in the correct format, point to a the path or file of the module or script to publish.");
```
PaulHigin: The above `GetResolvedPSPathFRomPSPath` will throw a `MethodInvocationException` if the path cannot be resolved (does not exist).  So we can remove the `does not exist` part of the message here, or alternatively catch the exception above and let this check notify the user if it does not exist.

src/code/PublishPSResource.cs
```diff
@@ -212,13 +214,35 @@ protected override void EndProcessing()
                 }
 
                 // remove '.ps1' extension from file name
-                _pkgName = pkgFileOrDir.Name.Remove(pkgFileOrDir.Name.Length - 4);
+                _pkgName = pkgFileOrDir.Name.Remove(pkgFileOrDir.Name.Length - PSScriptFileExt.Length);
```
PaulHigin:
```suggestion
                _pkgName = Path.GetFileNameWithoutExtension(pkgFileOrDir.Name);

```

src/code/PublishPSResource.cs
```diff
@@ -212,13 +214,35 @@ protected override void EndProcessing()
...
-                _pkgName = pkgFileOrDir.Name.Remove(pkgFileOrDir.Name.Length - 4);
+                _pkgName = pkgFileOrDir.Name.Remove(pkgFileOrDir.Name.Length - PSScriptFileExt.Length);
             }
             else
             {
-                _pkgName = pkgFileOrDir.Name;
-                resourceFilePath = System.IO.Path.Combine(_path, _pkgName + PSDataFileExt);
+                // parsedMetadata needs to be initialized for modules, will later be passed in to create nuspec
                 parsedMetadata = new Hashtable();
+                if (pkgFileOrDir.Name.EndsWith(PSDataFileExt))
```
PaulHigin:
```suggestion
                if (pkgFileOrDir.Name.EndsWith(PSDataFileExt, StringComparison.OrdinalIgnoreCase))

```

src/code/PublishPSResource.cs
```diff
@@ -212,13 +214,35 @@ protected override void EndProcessing()
...
+                    _pkgName = pkgFileOrDir.Name.Remove(pkgFileOrDir.Name.Length - PSDataFileExt.Length);
+
+                }
+                else {
+                    // directory
+                    // search for module manifest
+                    List<FileInfo> childFiles = pkgFileOrDir.EnumerateFiles().ToList();
+
+                    resourceFilePath = string.Empty;
+                    foreach (FileInfo file in childFiles)
```
PaulHigin, line 254: No need to create a list object.
```suggestion
                    foreach (FileInfo childFile in pkgFileOrDir.EnumerateFiles())

```

src/code/PublishPSResource.cs
```diff
@@ -212,13 +214,35 @@ protected override void EndProcessing()
...
+                    {
+                        if (file.Name.EndsWith(PSDataFileExt))
+                        {
+                            resourceFilePath = file.FullName;
+                            _pkgName = file.Name.Remove(file.Name.Length - PSDataFileExt.Length);
```
PaulHigin:
```suggestion
                            _pkgName = Path.GetFileNameWithoutExtension(file.Name);

```

src/code/PublishPSResource.cs
```diff
@@ -212,13 +214,35 @@ protected override void EndProcessing()
...
+                }
+                else {
+                    // directory
+                    // search for module manifest
+                    List<FileInfo> childFiles = pkgFileOrDir.EnumerateFiles().ToList();
+
+                    resourceFilePath = string.Empty;
+                    foreach (FileInfo file in childFiles)
+                    {
+                        if (file.Name.EndsWith(PSDataFileExt))
```
PaulHigin:
```suggestion
                        if (file.Name.EndsWith(PSDataFileExt, StingComparer.OrdinalIgnoreCase))

```

src/code/PublishPSResource.cs
```diff
@@ -138,17 +138,19 @@ protected override void BeginProcessing()
 
             string resolvedPath = SessionState.Path.GetResolvedPSPathFromPSPath(Path).First().Path;
 
-            if (Directory.Exists(resolvedPath) || 
-                (File.Exists(resolvedPath) && resolvedPath.EndsWith(PSScriptFileExt, StringComparison.OrdinalIgnoreCase)))
+            // Condition 1: path is to the root directory of the module to be published
+            // Condition 2: path is to the .psd1 or .ps1 of the module/script to be published  
+            if (Directory.Exists(resolvedPath)
+                    || (File.Exists(resolvedPath)
+                        && (resolvedPath.EndsWith(PSDataFileExt, StringComparison.OrdinalIgnoreCase)
```
PaulHigin: It looks like we check path file extension multiple times.  Ideally, this would only be done once and we set local variables describing the state:
 - what to publish: enum -> script or module
 - path to script file to publish
 - path to module manifest file to publish
 - path to module directory to publish

Instead of if/else statements, it would be better to have two classes that encapsulate publishing a script or module, and that derive from a base class that provides common functions (temporary directory) and a `publish` virtual method.

src/code/PublishPSResource.cs
```diff
@@ -329,27 +353,41 @@ protected override void EndProcessing()
...
                 {
-                    // Create subdirectory structure in temp folder
-                    foreach (string dir in System.IO.Directory.GetDirectories(_path, "*", System.IO.SearchOption.AllDirectories))
+                    try
                     {
-                        var dirName = dir.Substring(_path.Length).Trim(_PathSeparators);
-                        System.IO.Directory.CreateDirectory(System.IO.Path.Combine(outputDir, dirName));
-                    }
+                        // If path is pointing to a file, get the parent directory, otherwise assumption is that path is pointing to the root directory
+                        string rootModuleDir = _path.EndsWith(PSDataFileExt) ? new FileInfo(_path).Directory.FullName : _path;
```
PaulHigin: We have checked the `_path` file extension previously.
```suggestion
                        string rootModuleDir = _path.EndsWith(PSDataFileExt, StringComparer.OriginalIgnoreCase) ? new FileInfo(_path).Directory.FullName : _path;

```

PaulHigin (changes requested):

src/code/PublishPSResource.cs
alerickson: I changed this a little bit so that it catches the `MethodInvocationException` and throws immediately if applicable.  I removed the check to see if the directory or file exists since that was unnecessary anyway.

src/code/PublishPSResource.cs
alerickson: That's a good idea.  I'll clean up the path variables here, but address splitting up the class in a different PR so we can review the design of it.

src/code/PublishPSResource.cs
```diff
@@ -136,22 +140,51 @@ protected override void BeginProcessing()
...
+            {
+                pathToModuleManifestToPublish = resolvedPath;
+                resourceType = ResourceType.Module;
+            }
+            else if (resolvedPath.EndsWith(PSScriptFileExt, StringComparison.OrdinalIgnoreCase))
+            {
+                pathToScriptFileToPublish = resolvedPath;
+                resourceType = ResourceType.Script;
+            }
+            else if (string.IsNullOrEmpty(resolvedPath))
```
PaulHigin: We should do this check first.  There is no point in finding out what kind of path if the string is null or empty.

src/code/PublishPSResource.cs
```diff
@@ -136,22 +140,51 @@ protected override void BeginProcessing()
...
+                            "SourcePathDoesNotExist", 
+                            ErrorCategory.InvalidArgument, 
+                            this));
+            }
 
-            if (Directory.Exists(resolvedPath) || 
-                (File.Exists(resolvedPath) && resolvedPath.EndsWith(PSScriptFileExt, StringComparison.OrdinalIgnoreCase)))
+            // Condition 1: path is to the root directory of the module to be published
+            // Condition 2: path is to the .psd1 or .ps1 of the module/script to be published  
+            _path = resolvedPath;
```
PaulHigin: Not sure why this assignment is necessary.  You already have fully resolved paths for each case...

src/code/PublishPSResource.cs
```diff
@@ -211,19 +238,37 @@ protected override void EndProcessing()
...
+                        }
+                    }
+                }
 
                 // Validate that there's a module manifest
-                if (!File.Exists(resourceFilePath))
+                if (!File.Exists(pathToModuleManifestToPublish))
                 {
-                    var message = String.Format("No file with a .psd1 extension was found in {0}.  Please specify a path to a valid modulemanifest.", resourceFilePath);
+                    var message = String.Format("No file with a .psd1 extension was found in {0}.  Please specify a path to a valid modulemanifest.", _path);
```
PaulHigin:
```suggestion
                    var message = String.Format("No file with a .psd1 extension was found in {0}.  Please specify a path to a valid modulemanifest.", pathToModuleManifestToPublish);

```

src/code/PublishPSResource.cs
```diff
@@ -334,34 +375,48 @@ protected override void EndProcessing()
                     }
                 }
 
-                if (isScript)
+                if (resourceType == ResourceType.Script)
                 {
                     // copy the script file to the temp directory
                     File.Copy(_path, System.IO.Path.Combine(outputDir, _pkgName + PSScriptFileExt), true);
```
PaulHigin:
```suggestion
                    File.Copy(pathToScriptFileToPublish, System.IO.Path.Combine(outputDir, _pkgName + PSScriptFileExt), true);

```

src/code/PublishPSResource.cs
```diff
@@ -334,34 +375,48 @@ protected override void EndProcessing()
...
                 {
-                    // Create subdirectory structure in temp folder
-                    foreach (string dir in System.IO.Directory.GetDirectories(_path, "*", System.IO.SearchOption.AllDirectories))
+                    try
                     {
-                        var dirName = dir.Substring(_path.Length).Trim(_PathSeparators);
-                        System.IO.Directory.CreateDirectory(System.IO.Path.Combine(outputDir, dirName));
-                    }
+                        // If path is pointing to a file, get the parent directory, otherwise assumption is that path is pointing to the root directory
+                        string rootModuleDir = !string.IsNullOrEmpty(pathToModuleManifestToPublish) ? new FileInfo(_path).Directory.FullName : _path;
```
PaulHigin:
```suggestion
                        string rootModuleDir = !string.IsNullOrEmpty(pathToModuleManifestToPublish) ? 
                            Path.GetDirectoryName(pathToModuleManifestToPublish) :
                            pathToModuleDirToPublish;

```

src/code/PublishPSResource.cs
```diff
@@ -334,34 +375,48 @@ protected override void EndProcessing()
...
 
-                    // Copy files over to temp folder
-                    foreach (string fileNamePath in System.IO.Directory.GetFiles(_path, "*", System.IO.SearchOption.AllDirectories))
-                    {
-                        var fileName = fileNamePath.Substring(_path.Length).Trim(_PathSeparators);
+                        // Create subdirectory structure in temp folder
+                        foreach (string dir in System.IO.Directory.GetDirectories(rootModuleDir, "*", System.IO.SearchOption.AllDirectories))
```
PaulHigin, line 391: I wonder if we should make this CopyDir Utility method public and use it here?  It would be nice to not duplicate code if we can help it.

https://github.com/PowerShell/PowerShellGet/blob/115343118fb9eb794b927bb78aa20c031768d900/src/code/Utils.cs#L1135

PaulHigin (changes requested):
Overall looks good.  Just a few comments and suggestions.

src/code/PublishPSResource.cs
alerickson: I just got rid `_path` and am just using `resolvedPath` where `_path` was previously used

src/code/PublishPSResource.cs
alerickson: There's a small issue with the Utility method, which is that it doesn't ignore the `.nuspec` file if it sees it.  The way the publish 'copy' function currently works is that it prioritizes any .nuspec file that it sees in the publish folder, so that if the user specifically wants to use their own nuspec file they can.  We can get rid of this and only use the PSGet generated .nuspec file and if there's feedback that the community wants that functionality, add it back in at that point.

PaulHigin: If the utility method is not suitable then it is fine to have your own custom code.

PaulHigin (approved):
LGTM

PR #717, opened 2022-07-18
Add -Force param to Register-PSResourceRepository and Set-PSResourceRepository

### LINKED ISSUE #612: Register-PSResourceRepository should support -Force

As a devops engineer, I want to be able to configure PSRepositories in a _somewhat declarative_ Register-or-Set way, particularly during automation runs.

```PowerShell
$Repositories = <# import a list from a psd1 file #> @{
    Name = "OurGallery"
    Url = "https://proget.OurDomain.com/nuget/PowerShell"
    Trusted = $true
    Priority = 90
}
Register-PSResourceRepository -Repositories $Repositories
```

## The problem

Currently, if the repository exists, `Register` fails, and I must use `Set` but although I can use the *same syntax* with `Set-PSResourceRepository`, in this case, if the repository does not _already_ exist, it fails.

So I have to:

1. Clear out existing registrations
2. **Separately** re-register the PSGallery correctly
3. Then register the **other** repositories we use

```PowerShell
@('PSGallery') + $Repositories.Name | Unregister-PSResourceRepository
Register-PSResourceRepository -PSGallery -Priority 10 
Register-PSResourceRepository -Repositories $Repositories
```

Or I have to:

1. Put the PSGallery into my config file *without* the URL
2. **Separately** re-register the PSGallery
3. `Register` **and** `Set` all the repositories, ignoring the errors

```PowerShell
Register-PSResourceRepository -PSGallery
Register-PSResourceRepository -Repositories $Repositories -ErrorAction SilentlyContinue -ErrorVariable Failed
if ($Failed) {
Set-PSResourceRepository -Repositories $Repositories
}
```

### Proposed technical implementation details

I'd like a `-Force` switch on either Register or Set that would overwrite existing repositories (in the case of Register) or register them (in the case of Set) so that I can have declarative Register-or-Set syntax for the set of repositories I need, without needing the extra separate call to fix PSGallery, and the extra call to unregister the repositories before re-registering them (or set them after failing to register them)...

alerickson:
# PR Summary
This PR adds a `-Force` parameter to `Register-PSResourceRepository`, which allows a user to update (or set) a resource repository if the repository already exists.

The PR also changes the default behavior of `Set-PSResourceRepository` by creating a new resource repository if a user attempts to set a resource repository that does not already exist. 

## PR Context

Resolves #612 

## PR Checklist

    - [x] None
    - **OR**
        - [ ] Issue filed: 
- **User-facing changes**
    - [x] Not Applicable
    - **OR**
        - [ ] Issue filed: 
    - **OR**
- **Tooling**
    - **OR**


src/code/SetPSResourceRepository.cs
```diff
@@ -89,11 +89,17 @@ public SwitchParameter Trusted
...
+        /// When specified, displays the successfully registered repository and its information.
         /// </summary>
         [Parameter]
         public SwitchParameter PassThru { get; set; }
 
+        /// <summary>
+        /// When specified, will register a new repository with information specified if one currently does not exist.
+        /// </summary>
+        [Parameter]
+        public SwitchParameter Force { get; set; }
```
ThomasNieto: A `Force` parameter isn't needed for `Set-PSResourceRepository`. If the repository exists then it will update the repo config if the repo doesn't exist then it should create the repo if all the required values (Uri) are passed.

For example take a look at `New-Variable` and `Set-Variable`:

```powershell
PS C:\> Get-Variable -Name Test
Get-Variable: Cannot find a variable with the name 'Test'.
PS C:\> New-Variable -Name Test -Value $true
PS C:\> New-Variable -Name Test -Value $false
New-Variable: A variable with name 'Test' already exists.
PS C:\> New-Variable -Name Test -Value $false -Force
PS C:\> Get-Variable -Name Test

Test                           False

PS C:\> Set-Variable -Name Test -Value $true
PS C:\> Get-Variable -Name Test

Test                           True

PS C:\> Remove-Variable -Name Test
PS C:\> Set-Variable -Name Test -Value $false
PS C:\> Get-Variable -Name Test

Test                           False
```

`New-Variable` will create the variable but fail if it already exists unless the `Force` flag is passed. `Set-Variable` will either set the variable value or create the variable if it doesn't exist.

`Register-PSRespositoy` and `Set-PSRepository` should match `New-Variable` and `Set-Variable` behavior.

alerickson: That makes sense!  Just removed the -Force parameter for Set-PSResourceRepository but kept the implementation so that set will still create a new repository by default if it doesn't already exist.

src/code/RegisterPSResourceRepository.cs
```diff
@@ -141,9 +147,19 @@ protected override void ProcessRecord()
                         ThrowTerminatingError(errorRecord);
                     }
 
+                    string[] errorMsgs;
                     try
                     {
-                        items.Add(NameParameterSetHelper(Name, _uri, Priority, Trusted, CredentialInfo));
+                        items.Add(RepositorySettings.AddRepository(Name, _uri, Priority, Trusted, CredentialInfo, Force, this, out errorMsgs));
+
+                        if (errorMsgs.Length > 0)
```
anamnavi: Can we only have 1 error message returned from this? If so then I think we should have the `errorMsg` variable be of type `string` instead of `string[]`. If instead we could have more than 1 error message, then should we write those other errors out too?

src/code/RegisterPSResourceRepository.cs
```diff
@@ -200,71 +216,30 @@ protected override void ProcessRecord()
...
+            var addedRepo = RepositorySettings.AddToRepositoryStore(PSGalleryRepoName, 
+                psGalleryUri, 
+                repoPriority, 
+                repoTrusted, 
+                repoCredentialInfo: null, 
+                Force, 
+                this, 
+                out string[] errorMsgs);
+
+            if (errorMsgs.Length > 0)
```
anamnavi: same comment as above about plurality of error(s) and whether we need to report all if them back?

src/code/RegisterPSResourceRepository.cs
```diff
@@ -368,11 +343,25 @@ private PSRepositoryInfo RepoValidationHelper(Hashtable repo)
...
                     repoUri,
                     repo.ContainsKey("Priority") ? Convert.ToInt32(repo["Priority"].ToString()) : defaultPriority,
                     repo.ContainsKey("Trusted") ? Convert.ToBoolean(repo["Trusted"].ToString()) : defaultTrusted,
-                    repoCredentialInfo);
+                    repoCredentialInfo,
+                    Force,
+                    this,
+                    out string[] errorMsgs);
+
+                if (errorMsgs.Length > 0)
```
anamnavi: same comment as above about plurality of error(s) and whether we need to report all if them back?

src/code/RepositorySettings.cs
```diff
@@ -74,20 +74,178 @@ public static void CheckRepositoryStore()
             }
         }
 
+        public static PSRepositoryInfo AddRepository(string repoName, Uri repoUri, int repoPriority, bool repoTrusted, PSCredentialInfo repoCredentialInfo, bool force, PSCmdlet cmdletPassedIn, out string[] errorMsgs)
+        {
+            errorMsgs = new string[] { };
```
anamnavi: instead of initializer brackets:
```suggestion
            errorMsgs = Utils.EmptyStrArray;
```

src/code/RepositorySettings.cs
```diff
@@ -74,20 +74,178 @@ public static void CheckRepositoryStore()
...
+                throw new ArgumentException("Cannot register PSGallery with -Name parameter. Try: Register-PSResourceRepository -PSGallery");
+            }
+
+            return AddToRepositoryStore(repoName, repoUri, repoPriority, repoTrusted, repoCredentialInfo, force, cmdletPassedIn, out errorMsgs);
+        }
+
+
+        public static PSRepositoryInfo AddToRepositoryStore(string repoName, Uri repoUri, int repoPriority, bool repoTrusted, PSCredentialInfo repoCredentialInfo, bool force, PSCmdlet cmdletPassedIn, out string[] errorMsgs)
+        {
+            errorMsgs = new string[] { };
```
anamnavi:
```suggestion
            errorMsgs = Utils.EmptyStrArray;
```

src/code/RepositorySettings.cs
```diff
@@ -74,20 +74,178 @@ public static void CheckRepositoryStore()
...
+
+                if (!isSecretManagementModuleAvailable)
+                {
+                    cmdletPassedIn.WriteWarning($"Microsoft.PowerShell.SecretManagement module cannot be found. Make sure it is installed before performing PSResource operations in order to successfully authenticate to PSResourceRepository \"{repoName}\" with its CredentialInfo.");
+                }
+            }
+
+            // determine if either 1 of 4 values are attempting to be set: Uri, Priority, Trusted, CredentialInfo.
+            // if none are (i.e only Name parameter was provided, write error)
+            if (repoUri == null && repoPriority == defaultPriority && _trustedNullable == null && repoCredentialInfo == null)
```
anamnavi, line 205: could a user do:
`Set-PSResourceRepository -Name "MyRepo" -Force`

to attempt to create the repo? i.e  do we allow users to solely create the repo with Set? If so remove this condition, otherwise this condition will ensure they are trying to set atleast one property for the repo aside from Name (whether the repo exists or not, using Force)

src/code/RepositorySettings.cs
```diff
@@ -126,8 +284,9 @@ public static PSRepositoryInfo Add(string repoName, Uri repoUri, int repoPriorit
         /// Updates a repository name, Uri, priority, installation policy, or credential information
         /// Returns:  void
         /// </summary>
-        public static PSRepositoryInfo Update(string repoName, Uri repoUri, int repoPriority, bool? repoTrusted, PSCredentialInfo repoCredentialInfo)
+        public static PSRepositoryInfo Update(string repoName, Uri repoUri, int repoPriority, bool? repoTrusted, PSCredentialInfo repoCredentialInfo, PSCmdlet cmdletPassedIn, out string[] errorMsgs)
         {
+            errorMsgs = new string[] { };
```
anamnavi:
```suggestion
            errorMsgs = Utils.EmptyStrArray;
```

src/code/RepositorySettings.cs
```diff
@@ -136,7 +295,9 @@ public static PSRepositoryInfo Update(string repoName, Uri repoUri, int repoPrio
                 XElement node = FindRepositoryElement(doc, repoName);
                 if (node == null)
                 {
-                    throw new ArgumentException("Cannot find the repository because it does not exist. Try registering the repository using 'Register-PSResourceRepository'");
+                    bool repoIsTrusted = repoTrusted == null || repoTrusted == false ? false : true;
```
anamnavi: I think this could simplify, but not sure if this is less readable?
```suggestion
                    bool repoIsTrusted = !(repoTrusted == null || repoTrusted == false);
```

test/RegisterPSResourceRepository.Tests.ps1
```diff
@@ -234,23 +234,94 @@ Describe "Test Register-PSResourceRepository" {
...
     }
 
-    $testCases2 = @{Type = "-Name is not specified";                     IncorrectHashTable = @{Uri = $tmpDir1Path};                           ErrorId = "NullNameForRepositoriesParameterSetRegistration,Microsoft.PowerShell.PowerShellGet.Cmdlets.RegisterPSResourceRepository"},
-                  @{Type = "-Name is PSGallery";                         IncorrectHashTable = @{Name = $PSGalleryName; Uri = $tmpDir1Path};    ErrorId = "PSGalleryProvidedAsNameRepoPSet,Microsoft.PowerShell.PowerShellGet.Cmdlets.RegisterPSResourceRepository"},
-                  @{Type = "-Uri not specified";                         IncorrectHashTable = @{Name = $TestRepoName1};                        ErrorId = "NullUriForRepositoriesParameterSetRegistration,Microsoft.PowerShell.PowerShellGet.Cmdlets.RegisterPSResourceRepository"},
-                  @{Type = "-Uri is not valid scheme";                   IncorrectHashTable = @{Name = $TestRepoName1; Uri="www.google.com"};  ErrorId = "InvalidUri,Microsoft.PowerShell.PowerShellGet.Cmdlets.RegisterPSResourceRepository"}
+    It "not register incorrectly formatted Name type repo among correct ones when incorrect type is -Name is not specified" {
```
anamnavi: may sound clearer:
```suggestion
    It "not register incorrectly formatted -Name type repo among correct ones, where incorrect one is missing -Name" {
```

test/RegisterPSResourceRepository.Tests.ps1
```diff
@@ -234,23 +234,94 @@ Describe "Test Register-PSResourceRepository" {
...
+
+        $res2 = Get-PSResourceRepository -Name $TestRepoName3
+        $res2.Name | Should -Be $TestRepoName3
+
+        $res3 = Get-PSResourceRepository -Name $PSGalleryName
+        $res3.Name | Should -Be $PSGalleryName
+        $res3.Priority | Should -Be 30
+    }
+
+    It "not register incorrectly formatted Name type repo among correct ones when incorrect type is -Name is PSGallery" {
```
anamnavi:
```suggestion
    It "not register incorrectly formatted -Name type repo among correct ones, where incorrect type has -Name of PSGallery" {
```

test/SetPSResourceRepository.Tests.ps1
```diff
@@ -241,6 +241,17 @@ Describe "Test Set-PSResourceRepository" {
         $Res.Uri.LocalPath | Should -Contain "\\hcgg.rest.of.domain.name\test\ITxx\team\NuGet\"
     }
 
+    It "set repository should register repository if it does not already exist" {
+        $testRepoName = "NewForceTestRepo"
+        Set-PSResourceRepository -Name $testRepoName -Uri $tmpDir1Path -PassThru
```
anamnavi, line 257: Would we not need to use `-Force` with this? How is this creating this right now without `-Force` and passing the test?

anamnavi: Or do we create create with `Set` even if repo doesn't exist, *without* Force? Because that seems like it would be problematic..

anamnavi (changes requested):
It's looking great. I just had some design questions about:
1) `Set` creating a repo (if it doesn't already exist) *without* -Force
2)  Set being run with Name, Force and no other parameter ..is this a valid case (because at that point you are setting no properties, just registering the repository). I feel it would not be a valid case..but would love to discuss

src/code/RepositorySettings.cs
ThomasNieto: Set wouldn't have a force parameter, see this [discussion](https://github.com/PowerShell/PowerShellGet/pull/717#discussion_r924933746). If the user runs `Set-PSResourceRepository -Name MyRepo` and the repo doesn't exist it should throw an error saying `Uri` is required to register the repository.

test/SetPSResourceRepository.Tests.ps1
alerickson: Read @ThomasNieto's comment above as for the reasoning to not include -Force for set. I initially included a -Force parameter for set, but was convinced by his explanation as to why it should be the default behavior and not functionality only added with the -Force param

src/code/RepositorySettings.cs
alerickson: Yup, Set is just calling 'Add' repository, so anything that fails for adding a new repository would fail for setting a repository that does not exist.

>  If so remove this condition, otherwise this condition will ensure they are trying to set atleast one property for the repo aside from Name (whether the repo exists or not, using Force)

We still want to leave the checks in because we want to ensure the properties passed in (eventaully to 'Add') are valid

src/code/RegisterPSResourceRepository.cs
alerickson: Only 1 error message.  I can change it to string, I left it as string[] to be consistent with the other out error message variables, but can update this to just string

src/code/SetPSResourceRepository.cs
SydneyhSmith: Thanks @ThomasNieto we like this design, what are your thoughts on if there should be a warning that a new PSResourceRepository has been registered?

ThomasNieto: No warning should be written.

src/code/RepositorySettings.cs
```diff
@@ -74,20 +74,178 @@ public static void CheckRepositoryStore()
...
+
+            var repo = RepositorySettings.Add(repoName, repoUri, repoPriority, repoTrusted, repoCredentialInfo, force);
+
+            return repo;
+        }
+
+
+        public static PSRepositoryInfo UpdateRepositoryStore(string repoName, Uri repoUri, int repoPriority, bool repoTrusted, bool isSet, int defaultPriority, PSCredentialInfo repoCredentialInfo, PSCmdlet cmdletPassedIn, out string[] errorMsgs)
+        {
+            var tempErrors = new List<string>() { };
```
anamnavi: I believe you don't need bracket initializer
```suggestion
            var tempErrors = new List<string>();
```

src/code/RepositorySettings.cs
```diff
@@ -136,7 +295,9 @@ public static PSRepositoryInfo Update(string repoName, Uri repoUri, int repoPrio
                 XElement node = FindRepositoryElement(doc, repoName);
                 if (node == null)
                 {
-                    throw new ArgumentException("Cannot find the repository because it does not exist. Try registering the repository using 'Register-PSResourceRepository'");
+                    bool repoIsTrusted = !(repoTrusted == null || repoTrusted == false);
+                    repoPriority = repoPriority < 0 ? defaultPriority : repoPriority;
+                    return AddToRepositoryStore(repoName, repoUri, repoPriority, repoIsTrusted, repoCredentialInfo, force:true, cmdletPassedIn, out errorMsgs);
```
anamnavi: do you mind removing L317 below? Not sure how the `bool urlAttributeExists = node.Attribute("Url") != null;` got added twice (probably a typo), thanks!

anamnavi (approved):
LGTM, minor nitpicks

src/code/RegisterPSResourceRepository.cs
anamnavi: I think error string would be better

src/code/RepositorySettings.cs
anamnavi: Ignore my comment here, I just re-read this and one line is to check for uri and the other for url existing

PR #748, opened 2022-08-10
Fix -PassThru for Set-PSResourceRepository cmdlet to return all properties

### LINKED ISSUE #715: Set-PSResourceRepository -PassThru not returning all properties
### Prerequisites

### Steps to reproduce

If `Set-PSResourceRepository` is called the `Uri` property is not populated when using `-PassThru` parameter. I don't know if `CredentialInfo` is also impacted.

### Expected behavior

```console
PS C:\> Set-PSResourceRepository -Name PSGallery -Trusted:$false -PassThru

Name      Uri                                      Trusted Priority
----      ---                                      ------- --------
PSGallery https://www.powershellgallery.com/api/v2 False   50
```

### Actual behavior

```console
PS C:\> Set-PSResourceRepository -Name PSGallery -Trusted:$false -PassThru

Name      Uri Trusted Priority
----      --- ------- --------
PSGallery     False   50
```

### Error details

```console
N/A
```

### Environment data

```powershell
Name          Version
----          -------
PowerShellGet 3.0.14

PSVersion                      7.2.5
PSEdition                      Core
GitCommitId                    7.2.5
OS                             Microsoft Windows 10.0.19043
Platform                       Win32NT
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0…}
PSRemotingProtocolVersion      2.3
SerializationVersion           1.1.0.1
WSManStackVersion              3.0
```

### Visuals

_No response_

alyssa1303:
# PR Summary

This PR fixes `-PassThru` parameter for `Set-PSResourceRepository` to return all properties, specifically the missing `Uri`.

## PR Context

Resolve #715 

## PR Checklist

    - [x] None
    - **OR**
        - [ ] Issue filed: 
- **User-facing changes**
    - [x] Not Applicable
    - **OR**
        - [ ] Issue filed: 
    - **OR**
- **Tooling**
    - **OR**


src/code/RepositorySettings.cs
```diff
@@ -168,28 +168,27 @@ public static PSRepositoryInfo Update(string repoName, Uri repoUri, int repoPrio
...
-                        }
+                    }
+                    if (!Uri.TryCreate(node.Attribute("Uri").Value, UriKind.Absolute, out thisUrl))
+                    {
+                        throw new PSInvalidOperationException(String.Format("Unable to read incorrectly formatted Uri for repo {0}", repoName));
                     }
                 }
 
+                
+
```
alerickson: You can get rid of all these new lines (lines 190-194), we should have 1 new line at the end of the file, but we can delete everything else.

test/SetPSResourceRepository.Tests.ps1
```diff
@@ -250,6 +250,15 @@ Describe "Test Set-PSResourceRepository" {
         $res.Trusted | Should -Be False
```
alerickson, line 272: Just a comment, unrelated to this PR-- it's kind of strange that we were actually testing to see if -PassThru returned the correct Uri path, but we still somehow still managed to have a bug with that scenario.  Do either of you know why that's the case? @alyssa1303 @anamnavi?

alerickson (approved):
Looks good!  Just have the one request to remove whitespace before merging

test/SetPSResourceRepository.Tests.ps1
anamnavi: Unfortunately I was testing `-PassThru` with setting the Uri. The issue is reproduced only when setting anything but the Uri, as that's when it detects Uri is not being set and incorrectly sets it to null.

src/code/RepositorySettings.cs
```diff
@@ -168,28 +168,27 @@ public static PSRepositoryInfo Update(string repoName, Uri repoUri, int repoPrio
 
                 // determine if existing repository node (which we wish to update) had Url to Uri attribute
```
anamnavi: nitpick, I think I left this typo in last time
```suggestion
                // determine if existing repository node (which we wish to update) had Url or Uri attribute
```

src/code/RepositorySettings.cs
```diff
@@ -168,28 +168,27 @@ public static PSRepositoryInfo Update(string repoName, Uri repoUri, int repoPrio
...
+                    if (repoUri != null) {
                         node.Attribute("Url").Value = repoUri.AbsoluteUri;
-                        // Create Uri from node Uri attribute to create PSRepositoryInfo item to return.
-                        if (!Uri.TryCreate(node.Attribute("Url").Value, UriKind.Absolute, out thisUrl))
-                        {
-                            throw new PSInvalidOperationException(String.Format("Unable to read incorrectly formatted Url for repo {0}", repoName));
-                        }
                     }
-                    else
+                    if (!Uri.TryCreate(node.Attribute("Url").Value, UriKind.Absolute, out thisUrl))
```
anamnavi: Your code sets the node's Uri (or Url) attribute to the Uri passed in by the user, then we try to validate the node's new Uri. If it fails we throw but we've changed the node.Attribute(Uri) (or Url) already and don't revert it back. On the plus side though, we *always* validate that the value present for Uri (or Url) is valid, which the original code didn't do.

Looping @alerickson @PaulHigin into this for what they also think the flow should be, but I think if we detect the new Uri value provided by the user isn't a valid Uri we should throw an error and keep the original. If the user isn't trying to set the Uri, in my opinion, we should still validate the existing Uri (or Url) and throw if needed, but keep the original value.

anamnavi: One way to do what I suggested could be that in the original code, we had a condition for `if (repoUri != null)`...I think we need an `else` to that condition which would try to validate the current Uri- throw if invalid and if valid then set `thisUri` to the current Uri value. This way when we create `updatedRepo` and use `thisUri` it's not possible for `thisUri` to be null.

src/code/RepositorySettings.cs
anamnavi: Agreed, sorry this must've been a typo in the original code to have so many newlines.

test/SetPSResourceRepository.Tests.ps1
```diff
@@ -250,6 +250,15 @@ Describe "Test Set-PSResourceRepository" {
...
     }
 
+    It "set repository given Priority and see updated repository with -PassThru" {
```
anamnavi, line 275: test looks great!!

anamnavi (changes requested):
I think we should account for the case where the user provides a Uri to be set for the repository, it's an invalid Uri, so we revert to the original value Uri before throwing. Just a thought on the flow, happy to discuss more :)

src/code/RepositorySettings.cs
alerickson: I mentioned this a little earlier, but I don't think we need to validate what the current contents of uri/url is.  I think you can just put lines 175-178 before lines 172-174 so that the error gets thrown before the property gets set. Same with 184-187 getting placed before 181-183.  That should fix the  original problem right?  Good catch on this bug though

PaulHigin: I feel the creation of the Uri from user input should be done before checking whether an existing Uri exists in the store (Can we ever have a repository store entry without a Uri?  If not then this check should not be needed).  I see that the `Uri.TryCreate()` is called twice, but this can be done just once and throw if needed, before anything else is done (since that is a terminating error condition).

alerickson: > Can we ever have a repository store entry without a Uri? 

Nope, only if a user manually modifies their file

alyssa1303: The reason why `Uri.TryCreate()` is called twice in my previous code is because there are 2 different conditions, one for Uri and the other for Url. I updated the code and now `Uri.TryCreate()` is called 4 times because for each condition, I need to check if the user input Uri/Url is a valid one before I update the value. I call `Uri.TryCreate()` before `if (repo != null)` because if user doesn't input a Uri, `thisUri` will be `null`, which causes the current bug.

src/code/RepositorySettings.cs
```diff
@@ -166,33 +166,35 @@ public static PSRepositoryInfo Update(string repoName, Uri repoUri, int repoPrio
...
-                // determine if existing repository node (which we wish to update) had Url to Uri attribute
+                // determine if existing repository node (which we wish to update) had Url or Uri attribute
                 Uri thisUrl = null;
-                if (repoUri != null)
+                if (urlAttributeExists) 
                 {
-                    if (urlAttributeExists)
+                    Uri.TryCreate(node.Attribute("Url").Value, UriKind.Absolute, out thisUrl);
+
+                    if (repoUri != null)
```
PaulHigin: We can move this `if()` statement outside the `if (urlAttributeExists)` if, since it is duplicated in both if/else blocks.
Also, we are only updating an existing Uri, if the user provided one.  If not we don't need to do anything and leave the previous Uri (Url) ... we have already validated that there is an existing Uri or Url.

```csharp
if (repoUri != null)
{
    // Update the existing Uri or Url with the new one provided by the user.
    if (!Uri.TryCreate(repoUri.AbsoluteUri, UriKind.Absolute, out userUri))
    {
        throw ...
    }

    if (urlAttributeExists)
    {
        node.Attribute("Url").Value = userUri;
    }
    else
    {
        node.Attribute("Uri").Value = userUri;
    }
}
```

PaulHigin (changes requested):

src/code/RepositorySettings.cs
PaulHigin: I feel need to only call `Uri.TryCreate()` one time for the Uri provided by the user.  Also I think we only need to update the existing Uri or Url if the user provided one (see my comment below), and do nothing otherwise.

src/code/RepositorySettings.cs
alyssa1303: Yes, I can refactor it like you recommended. However, I think I still need to keep the `else` part for when `repo` is `null` because even though the existing Uri (Url) is not affected, when the `updatedRepo` is created and returned in line 245 - 259, `thisUrl` will be `null` when user uses `-PassThru`. The `updatedRepo` takes a `Uri` argument, not `string`, so I'm not sure if there's a better way than just assign the existing Uri (Url) to `thisUri`.

anamnavi (approved):
LGTM

alerickson (approved):

src/code/RepositorySettings.cs
```diff
@@ -166,33 +166,34 @@ public static PSRepositoryInfo Update(string repoName, Uri repoUri, int repoPrio
...
                         {
                             throw new PSInvalidOperationException(String.Format("Unable to read incorrectly formatted Url for repo {0}", repoName));
                         }
+
+                    if (urlAttributeExists)
+                    {
+                        node.Attribute("Url").Value = thisUrl.AbsoluteUri;
                     }
-                    else
+                    else{
```
PaulHigin: Please be consistent with bracket placement (on following line).

src/code/RepositorySettings.cs
```diff
@@ -166,33 +166,34 @@ public static PSRepositoryInfo Update(string repoName, Uri repoUri, int repoPrio
...
+                {
+                    if (urlAttributeExists)
                     {
-                        // Uri attribute exists
-                        node.Attribute("Uri").Value = repoUri.AbsoluteUri;
-                        if (!Uri.TryCreate(node.Attribute("Uri").Value, UriKind.Absolute, out thisUrl))
-                        {
-                            throw new PSInvalidOperationException(String.Format("Unable to read incorrectly formatted Uri for repo {0}", repoName));
-                        }
+                        Uri.TryCreate(node.Attribute("Url").Value, UriKind.Absolute, out thisUrl);
```
PaulHigin: I think we should test for failure on Uri creation.  Seems unlikely it would fail, but if it does that means the repository entry is corrupted and useless.  At minimum we should throw an error informing the user.  We might consider just removing the entry, but that may be too drastic.

PaulHigin (changes requested):

src/code/RepositorySettings.cs
anamnavi: Agreed on not removing the entry, just throwing an error

src/code/RepositorySettings.cs
alyssa1303: Will fix that!

src/code/RepositorySettings.cs
alyssa1303: I made the change you requested. I'm not sure if I throw the right kind of error with the right message so I'd really appreciate your input on that.

src/code/RepositorySettings.cs
```diff
@@ -356,11 +356,17 @@ public static PSRepositoryInfo Update(string repoName, Uri repoUri, int repoPrio
                 {
                     if (urlAttributeExists)
                     {
-                        Uri.TryCreate(node.Attribute("Url").Value, UriKind.Absolute, out thisUrl);
+                        if(!Uri.TryCreate(node.Attribute("Url").Value, UriKind.Absolute, out thisUrl))
+                        {
+                            throw new PSInvalidOperationException(String.Format("Current 'Url' for repo {0} is an invalid Uri. Fix this so the file is not corrupted.", repoName));
```
PaulHigin: Maybe:

"The 'Url' for repository {0} is invalid and the repository cannot be used. Please update the Url field or remove the repository entry."

PaulHigin (changes requested):

src/code/RepositorySettings.cs
alyssa1303: Sounds good! I'll change to that

PaulHigin (approved):
LGTM

PR #750, opened 2022-08-15
Install-PSResource: Installation path for scripts not included in environment PATH variable

### LINKED ISSUE #709: Install-psresource does not install PSGalleryScripts
### Prerequisites

### Steps to reproduce

This issue only affects scripts that are published to the powershell gallery, or psgalleryscripts.  For the purposes of this issue, the example script I will use is "Export-PSObjectToINI,"  You can use "Find-Psresource" to find the "Export-PSObjectToINI" script hosted in the powershell gallery easily:
2022-07-14 15_26_48-Administrator_ Windows PowerShell ISE
But when trying to install the script using "install-psresource", the module doesn't return any error or output, which would typically indicate the resource was successfully installed, but attempting to run the script reveals that it was not installed at all:

2022-07-14 15_29_36-Administrator_ Windows PowerShell ISE

This is a problem that needs to be resolved, as the only way to install scripts from the powershell gallery as of now is to import the old version of powershellget and use the "install-script" command.  

### Expected behavior

```console
The script should have successfully installed on the local machine.
```

### Actual behavior

```console
The script was not installed on the local machine
```

### Error details

```console
No error to be had.
```

### Environment data

```powershell
get-module powershellget; $PSVersionTable

ModuleType Version    Name                                ExportedCommands                                                                                               
---------- -------    ----                                ----------------                                                                                               
Binary     3.0.14     PowerShellGet                       {Find-PSResource, Get-PSResource, Get-PSResourceRepository, Install-PSResource...}                             

Key   : PSVersion
Value : 5.1.22000.653
Name  : PSVersion


Key   : PSEdition
Value : Desktop
Name  : PSEdition


Key   : PSCompatibleVersions
Value : {1.0, 2.0, 3.0, 4.0...}
Name  : PSCompatibleVersions


Key   : BuildVersion
Value : 10.0.22000.653
Name  : BuildVersion


Key   : CLRVersion
Value : 4.0.30319.42000
Name  : CLRVersion


Key   : WSManStackVersion
Value : 3.0
Name  : WSManStackVersion


Key   : PSRemotingProtocolVersion
Value : 2.3
Name  : PSRemotingProtocolVersion


Key   : SerializationVersion
Value : 1.1.0.1
Name  : SerializationVersion
```

### Visuals

_No response_

anamnavi:
# PR Summary

Installing a script with Install-PSResource happens successfully however the scripts installation path isn't added to the environment PATH variable as was in V2. 

Upon discussion from the team, we feel it is best for consistency across Windows and Linux to not attempt to modify the user's/machine's environment PATH variable, rather emit a warning indicating how the user could do so and specifying the path they would need to add. The script is still installed successfully, but the user would have to modify the environment PATH variable for discoverability.

## PR Context

Resolves #709 

## PR Checklist

    - [ ] None
    - **OR**
        - [ ] Issue filed: 
- **User-facing changes**
    - [ ] Not Applicable
    - **OR**
        - [ ] Issue filed: 
    - **OR**
- **Tooling**
    - **OR**


src/code/InstallHelper.cs
```diff
@@ -575,6 +575,24 @@ private List<PSResourceInfo> FilterByInstalledPkgs(List<PSResourceInfo> packages
...
                     _cmdletPassedIn.WriteVerbose(String.Format("Successfully installed package '{0}' to location '{1}'", pkg.Name, installPath));
                     pkgsSuccessfullyInstalled.Add(pkg);
+
+                    if (!isModule)
+                    {
+                        // Add Scripts install path to Path environment variable so that it can be ran without prepending the path
+                        if (!Environment.GetEnvironmentVariable("PATH").Contains(installPath))
+                        {
+                            // need to add Scripts install path
+                            string currentPathEnvVarValue = Environment.GetEnvironmentVariable("PATH", EnvironmentVariableTarget.User);
```
alerickson: I think we should also check the scope and accommodate for AllUsers scope as well, similar to what v2 does here: https://github.com/PowerShell/PowerShellGetv2/blob/5aba7531f526b6e60b2757c36799995597b6dcec/src/PowerShellGet/private/functions/ValidateAndSet-PATHVariableIfUserAccepts.ps1#L34

alerickson:
V2 prompts for user to accept/confirm the action before setting the environment variable.  I'm not a fan of prompts but since it's changing the env var, I think it might be important to include (see: https://github.com/PowerShell/PowerShellGetv2/blob/5aba7531f526b6e60b2757c36799995597b6dcec/src/PowerShellGet/PSGet.Resource.psd1#L222 for the v2 prompt)

src/code/InstallHelper.cs
```diff
@@ -575,6 +575,24 @@ private List<PSResourceInfo> FilterByInstalledPkgs(List<PSResourceInfo> packages
 
                     _cmdletPassedIn.WriteVerbose(String.Format("Successfully installed package '{0}' to location '{1}'", pkg.Name, installPath));
                     pkgsSuccessfullyInstalled.Add(pkg);
+
+                    if (!isModule)
+                    {
+                        // Add Scripts install path to Path environment variable so that it can be ran without prepending the path
+                        if (!Environment.GetEnvironmentVariable("PATH").Contains(installPath))
```
alerickson: Also looks like v2 checks that that both `installPath` and `installPath\` are not in `PATH` before adding `installPath` to `PATH` (see: https://github.com/PowerShell/PowerShellGetv2/blob/5aba7531f526b6e60b2757c36799995597b6dcec/src/PowerShellGet/private/functions/ValidateAndSet-PATHVariableIfUserAccepts.ps1#L71)

src/code/InstallHelper.cs
alerickson: It also looks like on top of adding the scope specific target, it is also adding the path to the env var for the current process (see: https://github.com/PowerShell/PowerShellGetv2/blob/5aba7531f526b6e60b2757c36799995597b6dcec/src/PowerShellGet/private/functions/ValidateAndSet-PATHVariableIfUserAccepts.ps1#L110)

alerickson (changes requested):

src/code/InstallHelper.cs
anamnavi: I added a check to accommodate for AllUsers scope, and to add the path to current process target.

src/code/InstallHelper.cs
anamnavi: added this check, good catch thank you!

alerickson:
This looks good to me!  Should be good once the tests are resolved.

src/code/InstallHelper.cs
```diff
@@ -35,6 +35,8 @@ internal class InstallHelper
         public const string PSScriptFileExt = ".ps1";
         private const string MsgRepositoryNotTrusted = "Untrusted repository";
         private const string MsgInstallUntrustedPackage = "You are installing the modules from an untrusted repository. If you trust this repository, change its Trusted value by running the Set-PSResourceRepository cmdlet. Are you sure you want to install the PSresource from '{0}' ?";
+        private const string ScriptPATHPromptQuery = "Your system has not been configured with a default script installation path yet, which means you can only run a script by specifying the full path to the script file. This action places the script into the folder '{0}', and adds that folder to your PATH environment variable. Do you want to add the script installation path '{0}' to the PATH environment variable?";
```
```suggestion
        private const string ScriptPATHPromptQuery = "The installation path for the script does not appear in the {scope} path environment variable. Do you want the script installation path, {path}, added to the {scope} PATH environment variable?";

```

src/code/InstallHelper.cs
```diff
@@ -580,6 +587,43 @@ private List<PSResourceInfo> FilterByInstalledPkgs(List<PSResourceInfo> packages
 
                     _cmdletPassedIn.WriteVerbose(String.Format("Successfully installed package '{0}' to location '{1}'", pkg.Name, installPath));
                     pkgsSuccessfullyInstalled.Add(pkg);
+
+                    if (!_savePkg && !isModule)
+                    {
+                        // Add Scripts install path to Path environment variable so that it can be ran without prepending the path
```
```suggestion
                        // Add Scripts install path to Path environment variable so that it can be discovered.

```

src/code/InstallHelper.cs
```diff
@@ -580,6 +587,43 @@ private List<PSResourceInfo> FilterByInstalledPkgs(List<PSResourceInfo> packages
 
                     _cmdletPassedIn.WriteVerbose(String.Format("Successfully installed package '{0}' to location '{1}'", pkg.Name, installPath));
                     pkgsSuccessfullyInstalled.Add(pkg);
+
+                    if (!_savePkg && !isModule)
+                    {
+                        // Add Scripts install path to Path environment variable so that it can be ran without prepending the path
+                        string envPathValue = Environment.GetEnvironmentVariable("PATH"); // todo anam- does this get all targets?
```
PaulHigin: This retrieves environment variable in the current process. We want to check that both process and scope PATH variables contain the path, and if not then update those paths.  We only need to prompt the user for non-process (scope) PATH variable changes.

src/code/InstallHelper.cs
```diff
@@ -580,6 +587,43 @@ private List<PSResourceInfo> FilterByInstalledPkgs(List<PSResourceInfo> packages
 
                     _cmdletPassedIn.WriteVerbose(String.Format("Successfully installed package '{0}' to location '{1}'", pkg.Name, installPath));
                     pkgsSuccessfullyInstalled.Add(pkg);
+
+                    if (!_savePkg && !isModule)
+                    {
+                        // Add Scripts install path to Path environment variable so that it can be ran without prepending the path
+                        string envPathValue = Environment.GetEnvironmentVariable("PATH"); // todo anam- does this get all targets?
+                        string installPathwithBackSlash = installPath + "\\";
+                        if (!envPathValue.Contains(installPath) && !envPathValue.Contains(installPathwithBackSlash))
```
PaulHigin: This code is complex enough to be in its own private static method.  We also need error handling if the environment variable cannot be retrieved or set (access denied).  For example a non-Admin user will not be able to get/set machine scope variables.  Access error would probably happen during installation, but we still need to handle access errors here.

PaulHigin: The private method should also be able to override the user prompt via a `force` parameter.

test/InstallPSResource.Tests.ps1
```diff
@@ -181,32 +181,71 @@ Describe 'Test Install-PSResource for Module' {
...
     # Windows only
-    It "Install resource under no specified scope - Windows only" -Skip:(!(Get-IsWindows)) {
+    It "Install module resource under no specified scope - Windows only" -Skip:(!(Get-IsWindows)) {
         Install-PSResource -Name $testModuleName -Repository $PSGalleryName -TrustRepository 
         $pkg = Get-PSResource $testModuleName
         $pkg.Name | Should -Be $testModuleName
         $pkg.InstalledLocation.ToString().Contains("Documents") | Should -Be $true
     }
 
+    # Windows only
```
PaulHigin: We should also have tests that verify environment PATH variables exist (for non-interactive case, force case).

PaulHigin (changes requested):

test/InstallPSResource.Tests.ps1
anamnavi: No longer applicable with warning based approach, so resolving.

PaulHigin (approved):
LGTM

alyssa1303 (approved):
LGTM too

alerickson (approved):

PR #763, opened 2022-08-16
Add -TemporaryPath parameter to Install-PSResource, Save-PSResource, and Update-PSResource

### LINKED ISSUE #650: Make the temp folder for Save-Module configurable
### Prerequisites

### Steps to reproduce

In Azure Functions, we use the `Save-Module` cmdlet to download the function app dependencies. At times, the modules to be installed are very large and the operation fails because there is not enough space in the `temp` directory. Having the ability to set an alternative to the `temp` directory via an environment variable could help us in this scenario. For reference, please see https://github.com/Azure/azure-functions-powershell-worker/issues/771.

### Expected behavior

```console
N/A
```

### Actual behavior

```console
N/A
```

### Error details

```console
N/A
```

### Environment data

```powershell
N/A
```

### Visuals

_No response_

alyssa1303:
# PR Summary

This PR adds a parameter called `-TemporaryPath` for `Install-PSResource`, `Save-PSResource`, and `Update-PSResource` cmdlets, which allows users to set the directory for temporary installation.  

## PR Context

Resolves #650 

## PR Checklist

    - [x] None
    - **OR**
        - [ ] Issue filed: 
- **User-facing changes**
    - [x] Not Applicable
    - **OR**
        - [ ] Issue filed: 
    - **OR**
- **Tooling**
    - **OR**

### DIFF
--- a/src/code/InstallHelper.cs
+++ b/src/code/InstallHelper.cs
@@ -53,6 +53,7 @@ namespace Microsoft.PowerShell.PowerShellGet.Cmdlets
         private bool _savePkg;
         List<string> _pathsToSearch;
         List<string> _pkgNamesToInstall;
+        private string _tmpPath;
 
         #endregion
 
@@ -82,7 +83,8 @@ namespace Microsoft.PowerShell.PowerShellGet.Cmdlets
             bool skipDependencyCheck,
             bool authenticodeCheck,
             bool savePkg,
-            List<string> pathsToInstallPkg)
+            List<string> pathsToInstallPkg,
+            string tmpPath)
         {
             _cmdletPassedIn.WriteVerbose(string.Format("Parameters passed in >>> Name: '{0}'; Version: '{1}'; Prerelease: '{2}'; Repository: '{3}'; " +
                 "AcceptLicense: '{4}'; Quiet: '{5}'; Reinstall: '{6}'; TrustRepository: '{7}'; NoClobber: '{8}'; AsNupkg: '{9}'; IncludeXml '{10}'; SavePackage '{11}'",
@@ -113,6 +115,7 @@ namespace Microsoft.PowerShell.PowerShellGet.Cmdlets
             _includeXml = includeXml;
             _savePkg = savePkg;
             _pathsToInstallPkg = pathsToInstallPkg;
+            _tmpPath = tmpPath;
 
             // Create list of installation paths to search.
             _pathsToSearch = new List<string>();
--- a/src/code/InstallHelper.cs
+++ b/src/code/InstallHelper.cs
@@ -87,7 +87,7 @@ namespace Microsoft.PowerShell.PowerShellGet.Cmdlets
             string tmpPath)
         {
             _cmdletPassedIn.WriteVerbose(string.Format("Parameters passed in >>> Name: '{0}'; Version: '{1}'; Prerelease: '{2}'; Repository: '{3}'; " +
-                "AcceptLicense: '{4}'; Quiet: '{5}'; Reinstall: '{6}'; TrustRepository: '{7}'; NoClobber: '{8}'; AsNupkg: '{9}'; IncludeXml '{10}'; SavePackage '{11}'",
+                "AcceptLicense: '{4}'; Quiet: '{5}'; Reinstall: '{6}'; TrustRepository: '{7}'; NoClobber: '{8}'; AsNupkg: '{9}'; IncludeXml '{10}'; SavePackage '{11}'; TemporaryPath '{12}'",
                 string.Join(",", names),
                 versionRange != null ? (versionRange.OriginalString != null ? versionRange.OriginalString : string.Empty) : string.Empty,
                 prerelease.ToString(),
@@ -99,7 +99,8 @@ namespace Microsoft.PowerShell.PowerShellGet.Cmdlets
                 noClobber.ToString(),
                 asNupkg.ToString(),
                 includeXml.ToString(),
-                savePkg.ToString()));
+                savePkg.ToString(),
+                tmpPath != null ? tmpPath.ToString() : string.Empty));
 
             _versionRange = versionRange;
             _prerelease = prerelease;
@@ -115,7 +116,7 @@ namespace Microsoft.PowerShell.PowerShellGet.Cmdlets
             _includeXml = includeXml;
             _savePkg = savePkg;
             _pathsToInstallPkg = pathsToInstallPkg;
-            _tmpPath = tmpPath;
+            _tmpPath = tmpPath ?? Path.GetTempPath();
 
             // Create list of installation paths to search.
             _pathsToSearch = new List<string>();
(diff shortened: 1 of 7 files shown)


src/code/InstallPSResource.cs
```diff
@@ -67,6 +67,32 @@ class InstallPSResource : PSCmdlet
...
+            {
+                if (!string.IsNullOrEmpty(value))
+                {
+                    string resolvedPath = SessionState.Path.GetResolvedPSPathFromPSPath(value).First().Path;
+
+                    // Path where resource is temporarily saved must be a directory
+                    if (Directory.Exists(resolvedPath))
+                    {
+                        _tmpPath = resolvedPath;
+                    }
```
alerickson: We might want to emit a warning, verbose message or non-terminating error if the directory does not exist.  I do think the default behavior should be to use the default temp directory if the one provided by the user does not exist.  I'm just not sure if we should provide some kind of messaging to the user to that we are doing that or just specify what the expected behavior is in our docs.

alerickson (commented):
Everything looks good-- just have 1 comment on providing user messaging.  @anamnavi and @PaulHigin may have some opinions on that as well.

src/code/InstallPSResource.cs
anamnavi: This is a good question. On one hand, I think it's reasonable to try and use the default temp path if the user provided one does not exist. On the other hand, for a team like Azure functions or whoever is using this to provide a directory with more space, if we use the default temp path it's pointless for them...but the argument can be made that they should've ensured their desired directory was created. Is there a scenario where we would want to try to create the user provided temp path directory if it doesn't exist and throw/warn if that creation failed?

Essentially not sure what best practice is here, perhaps @PaulHigin may have more idea.

src/code/SavePSResource.cs
```diff
@@ -109,6 +109,32 @@ public string Path
         }
         private string _path;
 
+        /// <summary>
+        /// The destination where the resource is to be temporarily installed
```
anamnavi:
```suggestion
        /// The destination where the resource is to be temporarily saved to.

```

src/code/UpdatePSResource.cs
```diff
@@ -68,6 +69,32 @@ public sealed class UpdatePSResource : PSCmdlet
         [Parameter]
         public ScopeType Scope { get; set; }
 
+        /// <summary>
+        /// The destination where the resource is to be temporarily installed
```
anamnavi:
```suggestion
        /// The destination where the resource is to be temporarily installed to while updating.

```

anamnavi (commented):
Not sure about whether we warn/throw/write error if the temp path directory does not exist, but aside from that looks great to me.

src/code/InstallHelper.cs
```diff
@@ -97,7 +99,8 @@ public InstallHelper(PSCmdlet cmdletPassedIn)
                 noClobber.ToString(),
                 asNupkg.ToString(),
                 includeXml.ToString(),
-                savePkg.ToString()));
+                savePkg.ToString(),
+                tmpPath != null ? tmpPath.ToString() : string.Empty));
```
PaulHigin: `ToString()` is not needed since `tmpPath` is a string type.
```suggestion
                tmpPath ?? string.Empty));

```

src/code/InstallHelper.cs
```diff
@@ -113,6 +116,7 @@ public InstallHelper(PSCmdlet cmdletPassedIn)
             _includeXml = includeXml;
             _savePkg = savePkg;
             _pathsToInstallPkg = pathsToInstallPkg;
+            _tmpPath = tmpPath ?? Path.GetTempPath();
```
PaulHigin, line 120: I don't think we need the `_tmpPath` class field since it appears that tmpPath is only used here locally.

src/code/InstallPSResource.cs
```diff
@@ -67,6 +67,32 @@ class InstallPSResource : PSCmdlet
...
+        [Parameter]
+        [ValidateNotNullOrEmpty]
+        public string TemporaryPath
+        {
+            get
+            { return _tmpPath; }
+
+            set
+            {
+                if (!string.IsNullOrEmpty(value))
```
PaulHigin: This check is not needed because the above `[ValidateNotNullOrEmpty]` attribute will ensure the argument is not null or empty.

src/code/InstallPSResource.cs
PaulHigin: I agree with @anamnavi that if a temporary path is provided and that path does not exist (or some other error like access) then this is a terminating error at parameter binding time.  The user specified an invalid path and so the cmdlet operation cannot continue.

Also, the `GetResolvedPSPathFromPSPath` API will throw for a number of errors, including if it cannot resolve the path, so the `DirectoryExists()` is not needed.  But only if the path does not contain wildcard characters, which we don't want to support so we should throw if there are any.  I think something like:

```csharp
if (WildcardPattern.ContainsWildcardCharacters(value))
{
    throw PSArgumentException("Wildcard characters are not allowed in the temporary path.");
}

// This will throw if path cannot be resolved.
_tmpPath = SessionState.Path.GetResolvedPSPathFromPSPath(value).First().Path;
```

PaulHigin (changes requested):
I also noticed that the `InstallHelper` class is derived from `PSCmdlet`, and this is incorrect because it is not being processed as a PowerShell cmdlet.  I think this was done so that the PSCmdlet `ThrowTerminatingError()` and `WriteError()` methods could be called.  But this is incorrect (and I am not sure what the behavior is).  `InstallHelper` should throw a normal C# exceptions and the top level cmdlets should catch and rethrow.  Also `InstallHelper` should return a `string[]` for any errors to be reported by the top level cmdlet.

@alyssa1303 Can you make these changes, or create a new `Issue` to make these changes later (since this is not really in scope of this PR)?

anamnavi:
> I also noticed that the `InstallHelper` class is derived from `PSCmdlet`, and this is incorrect because it is not being processed as a PowerShell cmdlet. I think this was done so that the PSCmdlet `ThrowTerminatingError()` and `WriteError()` methods could be called. But this is incorrect (and I am not sure what the behavior is). `InstallHelper` should throw a normal C# exceptions and the top level cmdlets should catch and rethrow. Also `InstallHelper` should return a `string[]` for any errors to be reported by the top level cmdlet.
> 
> @alyssa1303 Can you make these changes, or create a new `Issue` to make these changes later (since this is not really in scope of this PR)?

I think creating an issue for that (and a separate PR) would be better.

src/code/InstallHelper.cs
alyssa1303: I created `_tmpPath` class field because the tmpPath needs to be passed from `InstallPackages()` to `InstallPackage()`. If you think it's better to add a parameter to `ProcessRepositories()` to pass `tmpPath` from `InstallPackages()` to `ProcessRepositories()` and then from `ProcessRepositories()` to `InstallPackage()`, I can change it that way.

alyssa1303:
I created an issue #767

src/code/InstallHelper.cs
PaulHigin: Oh, I see that now.  I am fine leaving it as is.

anamnavi (approved):
LGTM

PaulHigin (approved):
LGTM

PR #780, opened 2022-08-30
Make -Path a required parameter for Save-PSResource cmdlet

alyssa1303:
# PR Summary

This PR makes `-Path` a required parameter for `Save-PSResource` cmdlet and changes its logic.

## PR Checklist

    - [x] None
    - **OR**
        - [ ] Issue filed: 
- **User-facing changes**
    - [x] Not Applicable
    - **OR**
        - [ ] Issue filed: 
    - **OR**
- **Tooling**
    - **OR**

### DIFF
--- a/src/code/SavePSResource.cs
+++ b/src/code/SavePSResource.cs
@@ -81,7 +81,7 @@ namespace Microsoft.PowerShell.PowerShellGet.Cmdlets
         /// <summary>
         /// The destination where the resource is to be installed. Works for all resource types.
         /// </summary>
-        [Parameter]
+        [Parameter(Mandatory = true)]
         [ValidateNotNullOrEmpty]
         public string Path
         {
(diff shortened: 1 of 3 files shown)


anamnavi (approved):
LGTM

src/code/SavePSResource.cs
```diff
@@ -90,21 +90,13 @@ public string Path
...
-                string resolvedPath = string.Empty;
-                if (string.IsNullOrEmpty(value))
-                {
-                    // If the user does not specify a path to save to, use the user's current working directory
-                    resolvedPath = SessionState.Path.GetResolvedPSPathFromPSPath(SessionState.Path.CurrentFileSystemLocation.Path).First().Path;
-                }
-                else {
-                    resolvedPath = SessionState.Path.GetResolvedPSPathFromPSPath(value).First().Path;
-                }
+                if (WildcardPattern.ContainsWildcardCharacters(value))
```
alerickson, line 93: I think instead of checking for wildcards here we should try to resolve the path and if that throws, catch and display the error.

alerickson: That way if there's any issues at all with the path we can properly handle and emit the error.

PaulHigin: The problem with wildcards in the path is that they can return multiple matches, and then which path do we use?  Whatever we choose will not be apparent to the user, unless we write out a message.

I feel we should not support wildcards in this case, and require the user to provide a single path that resolves to a real location.  If not then that is a terminating error.  We don't need to test the path and return an error, because `GetResolvedPSPathFromPSPath` already does this and throws an exception with a readable error:
```powershell
"Cannot find path 'C:\temp\temp' because it does not exist."
```

PaulHigin (approved):
LGTM

src/code/SavePSResource.cs
alerickson: That's true, this makes sense

alerickson (approved):

PR #833, opened 2022-10-18
Add SupportsWildcards() attribute to parameters accepting wildcards

### LINKED ISSUE #751: Some parameters accept wildcards need but don’t have the [SupportsWildcards()] attribute
### Prerequisites

### Steps to reproduce

The following parameters accept wildcards need but don’t have the [SupportsWildcards()] attribute
•	Find-PSResource -Repository
•	Find-PSResource -Version
•	Save-PSResource -Repository
•	Update-ModuleManifest -AliasesToExport
•	Update-ModuleManifest -CmdletsToExport
•	Update-ModuleManifest -DscResourcesToExport
•	Update-ModuleManifest -FunctionsToExport
•	Update-ModuleManifest -VariablesToExport
•	Update-PSResource -Repository

### Expected behavior

```console
NA
```

### Actual behavior

```console
NA
```

### Error details

_No response_

### Environment data

```powershell
NA
```

### Visuals

_No response_

alyssa1303:
# PR Summary

This PR adds `[SupportsWildcards()]` attribute to parameters that accept wildcards but aren't currently having it.

## PR Context

Resolves #751 

## PR Checklist

    - [x] None
    - **OR**
        - [ ] Issue filed: 
- **User-facing changes**
    - [x] Not Applicable
    - **OR**
        - [ ] Issue filed: 
    - **OR**
- **Tooling**
    - **OR**

### DIFF
--- a/src/code/SavePSResource.cs
+++ b/src/code/SavePSResource.cs
@@ -47,7 +47,7 @@ namespace Microsoft.PowerShell.PowerShellGet.Cmdlets
         public string Version { get; set; }
 
         /// <summary>
-        /// Specifies to allow saveing of prerelease versions
+        /// Specifies to allow saving of prerelease versions
         /// </summary>
         [Parameter(ParameterSetName = NameParameterSet)]
         public SwitchParameter Prerelease { get; set; }
@@ -55,6 +55,7 @@ namespace Microsoft.PowerShell.PowerShellGet.Cmdlets
         /// <summary>
         /// Specifies the specific repositories to search within.
         /// </summary>
+        [SupportsWildcards]
         [Parameter(ParameterSetName = NameParameterSet)]
         [ArgumentCompleter(typeof(RepositoryNameCompleter))]
         [ValidateNotNullOrEmpty]
--- a/src/code/UpdatePSResource.cs
+++ b/src/code/UpdatePSResource.cs
@@ -58,6 +58,7 @@ namespace Microsoft.PowerShell.PowerShellGet.Cmdlets
         /// Specifies one or more repository names to update packages from.
         /// If not specified, search will include all currently registered repositories in order of highest priority.
         /// </summary>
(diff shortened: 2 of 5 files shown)


src/code/SavePSResource.cs
```diff
@@ -47,14 +47,15 @@ public sealed class SavePSResource : PSCmdlet
         public string Version { get; set; }
```
anamnavi, line 47: The `Version` parameter also supports wildcards. Running the cmdlet with `-Version "2.2.*"` saves the latest version that satisfies the wildcard. Do we want this to be the expected behavior? If so we should add `[SupportsWildcards]` attribute here. If instead, we want only specific versions to be allowed we should fix this in the code.

src/code/UpdatePSResource.cs
```diff
@@ -58,6 +58,7 @@ public sealed class UpdatePSResource : PSCmdlet
         /// Specifies one or more repository names to update packages from.
```
anamnavi, line 58: The `-Version` parameter also *takes* wildcards, but do we want this as expected behavior? I tried running `Update-PSResource "folder_test" -Version "4.5.*" -Repository PSGallery` but it updates to version 6.5.0.0 which is the latest version. So it doesn't make sense. If we want wildcard support for Version we should correct this. If we don't want that as expected behavior we should ensure wildcards are not passed in for the Version parameter.

anamnavi (changes requested):
Looking great so far! I left some questions we should consider. Also, Get-PSResource -Version parameter also supports wildcards so if we can add that would be great, thanks!

src/code/UpdatePSResource.cs
alyssa1303: I agree that the current behavior doesn't make sense, so we should discuss this in our next meeting.

alerickson (approved):

PR #839, opened 2022-10-26
Add Get-PSScriptFileInfo cmdlet

### LINKED ISSUE #762: Add Get-PSScriptFileInfo cmdlet

As a user I want a `Get-PSScriptFileInfo` cmdlet to return a `PSScriptFileInfo` object from a specified script. This functionality was present in PowerShellGet v2 using the `Test-ScriptFileInfo` cmdlet.

_No response_

alyssa1303:
# PR Summary

This PR adds `Get-PSScriptFileInfo` cmdlet which takes a `-Path` parameter and returns a `PSScriptFileInfo` object. 

## PR Context

Resolves #762 

## PR Checklist

    - [x] None
    - **OR**
        - [ ] Issue filed: 
- **User-facing changes**
    - [x] Not Applicable
    - **OR**
        - [ ] Issue filed: 
    - **OR**
- **Tooling**
    - **OR**


src/code/GetPSScriptFileInfo.cs
```diff
@@ -0,0 +1,86 @@
...
+
+        #endregion
+
+        #region Methods
+
+        protected override void EndProcessing()
+        {            
+            if (!Path.EndsWith(".ps1", StringComparison.OrdinalIgnoreCase))
+            {
+                var exMessage = "Path needs to end with a .ps1 file. Example: C:/Users/john/x/MyScript.ps1";
```
PaulHigin:
```suggestion
                var exMessage = "The script file pathname must end with a .ps1 file extension. Example: C:/Users/john/x/MyScript.ps1";
```

src/code/GetPSScriptFileInfo.cs
```diff
@@ -0,0 +1,86 @@
...
+    public sealed class GetPSScriptFileInfo : PSCmdlet
+    {
+        #region Parameters
+
+        /// <summary>
+        /// The path to the .ps1 file to retrieve.
+        /// </summary>
+        [Parameter(Position = 0, Mandatory = true)]
+        [ValidateNotNullOrEmpty]
+        public string Path { get; set; }
```
PaulHigin, line 24: We're not very consistent about it, but when expecting a file pathname, we generally use 'FilePath'

```suggestion
        public string FilePath { get; set; }
```

src/code/GetPSScriptFileInfo.cs
```diff
@@ -0,0 +1,86 @@
...
+using System.Management.Automation;
+using Microsoft.PowerShell.PowerShellGet.UtilClasses;
+
+namespace Microsoft.PowerShell.PowerShellGet.Cmdlets
+{
+    /// <summary>
+    /// Retrieve the contents of a .ps1 file
+    /// </summary>
+    [Cmdlet(VerbsCommon.Get, "PSScriptFileInfo")]
+    public sealed class GetPSScriptFileInfo : PSCmdlet
```
PaulHigin, line 15: If we return a known type, we generally include this:
```powershell
[OutputType(typeof(PSScriptFileInfo))]
```

src/code/GetPSScriptFileInfo.cs
```diff
@@ -0,0 +1,86 @@
...
+                var exMessage = "Path needs to end with a .ps1 file. Example: C:/Users/john/x/MyScript.ps1";
+                var ex = new ArgumentException(exMessage);
+                var InvalidPathError = new ErrorRecord(ex, "InvalidPath", ErrorCategory.InvalidArgument, null);
+                ThrowTerminatingError(InvalidPathError);   
+            }
+
+            var resolvedPaths = SessionState.Path.GetResolvedPSPathFromPSPath(Path);
+            if (resolvedPaths.Count != 1)
+            {
+                var exMessage = "Error: Could not resolve provided Path argument into a single path.";
```
PaulHigin, line 43:
```suggestion
                var exMessage = "Error: Could not resolve provided FilePath argument into a single file.";
```

src/code/GetPSScriptFileInfo.cs
```diff
@@ -0,0 +1,86 @@
...
+                scriptFileInfoPath: resolvedPath,
+                parsedScript: out PSScriptFileInfo psScriptFileInfo,
+                errors: out ErrorRecord[] errors,
+                out string[] verboseMsgs);
+
+            if (!isValidScript)
+            {
+                foreach (ErrorRecord error in errors)
+                {
+                    WriteVerbose("The .ps1 script file passed in was not valid due to: " + error.Exception.Message);
```
PaulHigin: We generally use dotNet string formatting for error messages:
```suggestion
                    WriteVerbose($"The '{fileName}' script file passed in was not valid due to: {error.Exception.Message}");
```

PaulHigin: Actually, since these are errors, they should be written as errors (WriteError()).  I would write them after the initial 'invalid file' error message below.  I see that you write 'verbose' messages later below.  I would just write them here to make the code a bit clearer.

src/code/GetPSScriptFileInfo.cs
```diff
@@ -0,0 +1,86 @@
...
+                }
+                
+                var exMessage = "Error: Invalid .ps1 script file. Verify that the script file has Version, Guid, Description and Author properties.";
+                var ex = new PSArgumentException(exMessage);
+                var InvalidPSScriptFileError = new ErrorRecord(ex, "InvalidPSScriptFile", ErrorCategory.InvalidArgument, null);
+                ThrowTerminatingError(InvalidPSScriptFileError);
+            }
+            else
```
PaulHigin: 'else' is not needed here since conditional block above returns early with a terminating error exception.

src/code/GetPSScriptFileInfo.cs
```diff
@@ -0,0 +1,86 @@
...
+                out string[] verboseMsgs);
+
+            if (!isValidScript)
+            {
+                foreach (ErrorRecord error in errors)
+                {
+                    WriteVerbose("The .ps1 script file passed in was not valid due to: " + error.Exception.Message);
+                }
+                
+                var exMessage = "Error: Invalid .ps1 script file. Verify that the script file has Version, Guid, Description and Author properties.";
```
PaulHigin:
```suggestion
                var exMessage = $"Error: '{fileName}' script file is invalid. The script file must include Version, Guid, Description and Author properties.";
```

PaulHigin (changes requested):

src/code/GetPSScriptFileInfo.cs
alyssa1303: In that case, should I change the parameter name for `Test-PSScriptFileInfo`, `New-PSScriptFileInfo`, and `Update-PSScriptFileInfo` since they all use `Path`? Otherwise, `Get-PSScriptFileInfo` will be inconsistent with them.

PaulHigin: Oh, no.  If the other PSGet use 'Path', let's be consistent with that.

src/code/GetPSScriptFileInfo.cs
alyssa1303: Thanks for the suggestion! I'll write them after the `invalid file` error message then. As for the verbose message at the end, they were written later because they should be written whether the script is a valid one or not. Those verbose message are for errors like an invalid projectUri. This type of error doesn't make the script invalid but will return a null projectUri object, which is why it's written as verbose message instead of error.

PaulHigin (approved):
LGTM

src/code/GetPSScriptFileInfo.cs
```diff
@@ -0,0 +1,86 @@
...
+            }
+
+            PSObject psScriptFileInfoWithName = new PSObject(psScriptFileInfo);
+
+            string Name = System.IO.Path.GetFileNameWithoutExtension(resolvedPath);
+            psScriptFileInfoWithName.Properties.Add(new PSNoteProperty(nameof(Name), Name));
+
+            WriteObject(psScriptFileInfoWithName);
+
+            foreach (string msg in verboseMsgs)
```
anamnavi, line 76: I think the verbose messages (potentially about Uris being invalid) should be written before the `WriteObject()` call that writes the script file object. That way all errors + verbose messages can be displayed and out of the way before the object is shown,

anamnavi (approved):
LGTM, just one request for change

Issue #634, opened 2022-05-01
Find-PSResource Name parameter must be provided


ThomasNieto:
### Prerequisites

- [X] Write a descriptive title.
- [X] Make sure you are able to repro it on the [latest released version](https://www.powershellgallery.com/packages/PowerShellGet)
- [X] Search the existing issues.

### Steps to reproduce

Find-PSResource errors out without a `-Name` parameter passed.

### Expected behavior

```console
Returns resources
```

### Actual behavior

```console
Find-PSResource: Name parameter must be provided.
```

### Error details

```console
Exception             :
    Type        : System.Management.Automation.PSInvalidOperationException
    ErrorRecord :
        Exception             :
            Type    : System.Management.Automation.ParentContainsErrorRecordException
            Message : Name parameter must be provided.
            HResult : -2146233087
        CategoryInfo          : InvalidOperation: (:) [], ParentContainsErrorRecordException
        FullyQualifiedErrorId : InvalidOperation
    TargetSite  :
        Name          : ThrowTerminatingError
        DeclaringType : System.Management.Automation.MshCommandRuntime, System.Management.Automation, Version=7.2.3.500, Culture=neutral, PublicKeyToken=31bf3856ad364e35
        MemberType    : Method
        Module        : System.Management.Automation.dll
    Message     : Name parameter must be provided.
    Source      : System.Management.Automation
    HResult     : -2146233079
    StackTrace  :
   at System.Management.Automation.MshCommandRuntime.ThrowTerminatingError(ErrorRecord errorRecord)
TargetObject          : Microsoft.PowerShell.PowerShellGet.Cmdlets.FindPSResource
CategoryInfo          : InvalidOperation: (Microsoft.PowerShel…lets.FindPSResource:FindPSResource) [Find-PSResource], PSInvalidOperationException
FullyQualifiedErrorId : NameParameterNotProvided,Microsoft.PowerShell.PowerShellGet.Cmdlets.FindPSResource
InvocationInfo        :
    MyCommand        : Find-PSResource
    ScriptLineNumber : 1
    OffsetInLine     : 1
    HistoryId        : 1
    Line             : Find-PSResource
    PositionMessage  : At line:1 char:1
                       + Find-PSResource
                       + ~~~~~~~~~~~~~~~
    InvocationName   : Find-PSResource
    CommandOrigin    : Internal
ScriptStackTrace      : at <ScriptBlock>, <No file>: line 1
```

### Environment data

```powershell
[...]

Issue #244, opened 2020-09-08
Document Proxy support


SydneyhSmith:

alerickson:
we should not implement the parmas right now and encourage users to configure this setting themselves, either through an environment variable or by modifying their nuget.config file.

The two implementations I was looking at were 1) modifying the process environment variable or 2) modifying the user's nuget.config file.

1) Initially modifying the environment variable seemed like the best route to take, but after doing more thorough research, it seems like a lot of folks have had issues configuring the value and getting NuGet to read it properly. NuGet has very little documentation here, but it looks like the `http_proxy` variable is one that's widely used and typically takes one of the following formats:
`http://<proxy>`
`http://<proxy>:<port>`
`http://<username>:<password>@<proxy>`
`http://<username>:<password>@<proxy>:<port>`
PSGet would need to do a lot of manipulation to get this configured properly. There's a lot of points of failure here which is concerning, especially given that there's no direct guidance on how this should be set up, and no easy way to thoroughly test scenarios. I can see this being a point of frustration and inconsistently working (ie working for some users but not others), which makes me hesitant.

2) Modifying the nuget.config file, seems to be a more reliable way to configure a proxy because we're adding a proxy key (which can include port number), adding a username key, and adding a password key, all as separate variables that are easily modifiable by the user since we store what's directly passed in from the user.  The primary issue with this implementation is that there's a potential security vulnerability with PSGet writing passwords to the nuget.config file. This implementation also feels a bit hacky.

What we'd like to do is encourage users to configure proxies themselves and if we get feedback that it's too frustrating or they need a smoother experience, we can come back to one of these implementations. The big todo here now is clear documentation on how users can properly set up one of these two configurations.

alerickson:
The above two linked PRs show an implementation for the two options.

sdwheeler:
It looks like `Publish-PSResource` is the only cmdlet that has proxy parameters.

PR #922, opened 2023-02-02
Update docs for beta19

sdwheeler:
Update docs for beta19 - reviewed all cmdlets for changes. Fixed some formatting and wording.

I addressed the following docs issues even though these changes are not in beta19 yet.

- PowerShell/PowerShellGet#634
  - Changed syntax and parameter metadata to show -Name as required

- PowerShell/PowerShellGet#244
  - Removed -Proxy and -ProxyCredential from Publish-PSResource doc


help/Find-PSResource.md
```diff
@@ -198,7 +207,7 @@ Takeown         1.0.2.0            PSGallery  Crescendo Powershell wrapper of ta
 
 ### -CommandName
 
-The name of the command to search for.
+The name of the command to search for. The name is case-sensitive.
```
alerickson, line 210: It's a bug that it's case sensitive.  I think we need to fix this before the next release so I don't know if it's worth adding this in here.

alerickson (changes requested):

help/Find-PSResource.md
sdwheeler: This is for beta19. I plan another edit pass for beta20 when it gets integrated with 7.4-preview.

alerickson: Should we let you know before we release beta20 so we can have updated docs for that release?

alerickson (approved):

PR #1089, opened 2023-05-03
Remove nuget dependencies

anamnavi:
# PR Summary

Remove NuGet package dependencies that were previously taken for Find and Install when using NuGet APIs

We can't remove dependencies to the following nuget packages as they are used by PublishPSResource.cs
NuGet.Common 
NuGet.Commands
NuGet.Configuration
NuGet.Packaging
NuGet.Protocol
NuGet.Frameworks

Also removes dead code that was previously called for the NuGet APIs' logic.
## PR Checklist

    - [ ] None
    - **OR**
        - [ ] Issue filed: 
- **User-facing changes**
    - [ ] Not Applicable
    - **OR**
        - [ ] Issue filed: 
    - **OR**
- **Tooling**
    - **OR**


PaulHigin (commented):
In what way is PSPublishResource.cs dependent on all of these NuGet binaries?  Is it because of `NuGet.Versioning`?  What would we have to do to get rid of the rest of the NuGet dependencies?

anamnavi:
> In what way is PSPublishResource.cs dependent on all of these NuGet binaries? Is it because of `NuGet.Versioning`? What would we have to do to get rid of the rest of the NuGet dependencies?

PublishPSResource.cs PackNupkg() and PushNupkg() methods use NuGet.Common, NuGet.Commands, NuGet.Configuration and NuGet.Packaging libraries. Those packages take dependencies on NuGet.Protocol, NuGet.ProjectModel, and NuGet.Frameworks. 

To remove all NuGet dependencies we would need to write our own APIs for packing and pushing a .nupkg file. On the plus side, we did remove all NuGet dependencies for the Find and Install scenarios (except for NuGet.Versioning which has no dependencies of its own)

src/code/InstallPSResource.cs
```diff
@@ -519,7 +519,7 @@ private void ProcessInstallHelper(string[] pkgNames, string pkgVersion, bool pkg
                 return;
             }
 
-            var installedPkgs = _installHelper.InstallPackages(
+            var installedPkgs = _installHelper.BeginInstallingPackages(
```
alerickson: I think we should change this name to 'BeginInstallPackages' just because we use 'Find', 'Install', etc. elsewhere in the code base

alerickson (approved):

alerickson:
@PaulHigin @anamnavi I think we need to explore getting rid of the NuGet APIs in Publish as a separate task because that's going to be quite a bit of work.

PaulHigin (approved):

PR #1276, opened 2023-07-20
Publish-PSResource should create nupkg packages with lowercased name

anamnavi:
# PR Summary

## PR Checklist

    - [ ] None
    - **OR**
        - [ ] Issue filed: 
- **User-facing changes**
    - [ ] Not Applicable
    - **OR**
        - [ ] Issue filed: 
    - **OR**
- **Tooling**
    - **OR**

### DIFF
--- a/src/code/PublishPSResource.cs
+++ b/src/code/PublishPSResource.cs
@@ -444,7 +444,8 @@ namespace Microsoft.PowerShell.PSResourceGet.Cmdlets
                     {
                         var exMessage = string.Format("Destination path does not exist: {0}", DestinationPath);
                         var ex = new ArgumentException(exMessage);
-                        var InvalidDestinationPath = new ErrorRecord(ex, "InvalidDestinationPath", ErrorCategory.InvalidArgument, null);
+                        var InvalidDestinationPath = new ErrorRecord(ex, "InvalidDestinationPath", ErrorCategory.InvalidArgument, targetObject: null);
+
                         WriteError(InvalidDestinationPath);
                         return;
                     }
--- a/src/code/PublishPSResource.cs
+++ b/src/code/PublishPSResource.cs
@@ -440,6 +440,15 @@ namespace Microsoft.PowerShell.PSResourceGet.Cmdlets
                 // If -DestinationPath is specified then also publish the .nupkg there
                 if (!string.IsNullOrWhiteSpace(DestinationPath))
                 {
+                    if (!Directory.Exists(DestinationPath))
+                    {
+                        var exMessage = string.Format("Destination path does not exist and cannot be created: {0}", DestinationPath);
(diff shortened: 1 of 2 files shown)


alerickson (approved):

src/code/PublishPSResource.cs
```diff
@@ -443,7 +443,10 @@ protected override void EndProcessing()
                     try
                     {
                         var nupkgName = _pkgName + "." + _pkgVersion.ToNormalizedString() + ".nupkg";
-                        File.Copy(System.IO.Path.Combine(outputNupkgDir, nupkgName), System.IO.Path.Combine(DestinationPath, nupkgName));
+                        string srcPath = System.IO.Path.Combine(outputNupkgDir, nupkgName);
+                        // The file that gets created upon Copy should have lowercased package name.
+                        string destPath = System.IO.Path.Combine(DestinationPath, nupkgName.ToLower());
```
adityapatwardhan, line 458: Do we care about the destination path not matching case?

anamnavi: I don't see why we would care...the file path at the destination should always have lower case, as that's what we put there and expect to see after publish. The srcPath has to match the case as it's the file name with the casing the (Windows or Unix) system sees and has to copy

src/code/PublishPSResource.cs
```diff
@@ -440,6 +440,15 @@ protected override void EndProcessing()
                 // If -DestinationPath is specified then also publish the .nupkg there
                 if (!string.IsNullOrWhiteSpace(DestinationPath))
                 {
+                    if (!Directory.Exists(DestinationPath))
+                    {
+                        var exMessage = string.Format("Destination path does not exist: {0}", DestinationPath);
+                        var ex = new ArgumentException(exMessage);
+                        var InvalidDestinationPath = new ErrorRecord(ex, "InvalidDestinationPath", ErrorCategory.InvalidArgument, null);
```
adityapatwardhan:
```suggestion
                        var InvalidDestinationPath = new ErrorRecord(ex, "InvalidDestinationPath", ErrorCategory.InvalidArgument, targetObject: null);

```

adityapatwardhan: Not very sure about the name "targetObject".. please check on that.

adityapatwardhan (approved):

src/code/PublishPSResource.cs
anamnavi: From the docs, "This is the object against which the cmdlet or provider was operating when the error occurred. This is optional."

We have it as null or the cmdlet throughout the code. For the other DestinationPath does not exist error in this file, the value is also null. I can change all of them to be cmdlet instead of null across the code in a separate PR.

Issue #1584, opened 2024-03-04
ACR should support prerelease versions


anamnavi:
### Prerequisites

- [X] Write a descriptive title.
- [X] Make sure you are able to repro it on the [latest released version](https://www.powershellgallery.com/packages/Microsoft.PowerShell.PSResourceGet)
- [X] Search the existing issues.

### Steps to reproduce

Find-PSResource, Install-PSResource and Publish-PSResource should work with prerelease versions.

### Expected behavior

```console
Find-PSResource, Install-PSResource and Publish-PSResource should work with prerelease versions.
```

### Actual behavior

```console
These cmdlets don't work with prerelease versions yet.
```

### Environment data

```powershell
all platforms.
```

Issue #1582, opened 2024-02-29
ACR: version last digit 0 bug produces version mismatch


anamnavi:
### Prerequisites

- [X] Write a descriptive title.
- [X] Make sure you are able to repro it on the [latest released version](https://www.powershellgallery.com/packages/Microsoft.PowerShell.PSResourceGet)
- [X] Search the existing issues.

### Steps to reproduce

Publish a resource with version 1.0.0.0. Publish-PSResource will succeed, but Find-PSResource with -Version "1.0", "1.0.0", "1.0.0.0" will all fail. This is because PSResourceGet code converts the user inputted version to a Normalized Version, so "1.0" becomes "1.0.0" or "1.0.0.0" becomes "1.0.0" and ACR requires the exact tag to find the package, so anything aside from "1.0.0.0" will fail and give a Package Not Found error.

This occurs if the version has 4 digits and ends with 0, or if the version has 2 digits and ends with 0.

### Expected behavior

```console
Given that a package myPkg with version "1.0.0.0" is published, all the following should work

Find-PSResource -Name "myPkg" -Version "1.0"
Find-PSResource -Name "myPkg" -Version "1.0.0"
Find-PSResource -Name "myPkg" -Version "1.0.0.0"
```

### Actual behavior

```console
Given that a package myPkg with version "1.0.0.0" is published, the following do not work and package is not found:

Find-PSResource -Name "myPkg" -Version "1.0"
Find-PSResource -Name "myPkg" -Version "1.0.0"
Find-PSResource -Name "myPkg" -Version "1.0.0.0"
```

### Environment data

```powershell
all platforms
```

PR #1588, opened 2024-03-06
ACR: Ensure versions returned are sorted and can be correctly compared if number of trailing zeroes differ

### LINKED ISSUE #1581: ACR: Versions are not in order
### Prerequisites

- [X] Make sure you are able to repro it on the [latest released version](https://www.powershellgallery.com/packages/Microsoft.PowerShell.PSResourceGet)

### Steps to reproduce

Versions (which are ACR tags) are not guaranteed to be in alphabetical or numerical order, just chronological wrt to when they were published. When performing search with globbing patterns, ensure versions are in order.

### Expected behavior

```console
Versions should be sorted into some order, so that they are returned in decreasing numerical order.
```

### Actual behavior

```console
Versions are in order of when they were published.
```

### Error details

_No response_

### Environment data

```powershell
all plaforms.
```

### Visuals

_No response_

anamnavi:
Versions stored in ACR are not present in numerical order, instead they use chronological order of when the repository tag (i.e version in our context) is published. This PR sorts tags (i.e versions) returned in FindACRImageTags().

The tag is a string, but we use it for version and version can be a NuGetVersion that may have different number of trailing 0's but the normalized version would be the same. So if a version "1.0.0" is published, the user input of version "1.0" or "1.0.0.0" should also match it but previously did not. This PR fixes this by finding the exact tag (i.e version) from the server and using that when we call GetACRMetadata().

Also add support for prerelease versions.

# PR Summary

## PR Context
Resolves #1581 #1582 #1584

## PR Checklist

    - [ ] None
    - **OR**
        - [ ] Issue filed: 
- **User-facing changes**
    - [ ] Not Applicable
    - **OR**
        - [ ] Issue filed: 
    - **OR**
- **Tooling**
    - **OR**

### DIFF
--- a/src/code/ACRServerAPICalls.cs
+++ b/src/code/ACRServerAPICalls.cs
@@ -154,9 +154,9 @@ namespace Microsoft.PowerShell.PSResourceGet
             List<JToken> allVersionsList = foundTags["tags"].ToList();
 
             SortedDictionary<NuGet.Versioning.SemanticVersion, string> sortedQualifyingPkgs = GetPackagesWithRequiredVersion(allVersionsList, VersionType.VersionRange, VersionRange.All, specificVersion: null, packageName, includePrerelease, out errRecord);
-            foreach(KeyValuePair<NuGet.Versioning.SemanticVersion, string> s in sortedQualifyingPkgs.Reverse())
+            foreach(KeyValuePair<NuGet.Versioning.SemanticVersion, string> pkgVersionTag in sortedQualifyingPkgs.Reverse())
             {
-                string exactTagVersion = s.Value.ToString();
+                string exactTagVersion = pkgVersionTag.Value.ToString();
                 Hashtable metadata = GetACRMetadata(Registry, packageNameLowercase, exactTagVersion, acrAccessToken, out errRecord);
                 if (errRecord != null || metadata.Count == 0)
                 {
@@ -260,9 +260,9 @@ namespace Microsoft.PowerShell.PSResourceGet
             List<Hashtable> latestVersionResponse = new List<Hashtable>();
             List<JToken> allVersionsList = foundTags["tags"].ToList();
             SortedDictionary<NuGet.Versioning.SemanticVersion, string> sortedQualifyingPkgs = GetPackagesWithRequiredVersion(allVersionsList, VersionType.VersionRange, versionRange, specificVersion: null, packageName, includePrerelease, out errRecord);
-            foreach(KeyValuePair<NuGet.Versioning.SemanticVersion, string> s in sortedQualifyingPkgs.Reverse())
+            foreach(KeyValuePair<NuGet.Versioning.SemanticVersion, string> pkgVersionTag in sortedQualifyingPkgs.Reverse())
             {
-                string exactTagVersion = s.Value.ToString();
+                string exactTagVersion = pkgVersionTag.Value.ToString();
                 Hashtable metadata = GetACRMetadata(Registry, packageName.ToLower(), exactTagVersion, acrAccessToken, out errRecord);
                 if (errRecord != null || metadata.Count == 0)
                 {
@@ -320,9 +320,9 @@ namespace Microsoft.PowerShell.PSResourceGet
             List<JToken> allVersionsList = foundTags["tags"].ToList();
 
             SortedDictionary<NuGet.Versioning.SemanticVersion, string> sortedQualifyingPkgs = GetPackagesWithRequiredVersion(allVersionsList, VersionType.SpecificVersion, VersionRange.All, requiredVersion, packageName, includePrereleaseVersions, out errRecord);
-            foreach(KeyValuePair<NuGet.Versioning.SemanticVersion, string> s in sortedQualifyingPkgs.Reverse())
+            foreach(KeyValuePair<NuGet.Versioning.SemanticVersion, string> pkgVersionTag in sortedQualifyingPkgs.Reverse())
             {
-                string exactTagVersion = s.Value.ToString();
+                string exactTagVersion = pkgVersionTag.Value.ToString();
                 Hashtable metadata = GetACRMetadata(Registry, packageNameLowercase, exactTagVersion, acrAccessToken, out errRecord);
                 if (errRecord != null || metadata.Count == 0)
                 {
--- a/src/code/ACRServerAPICalls.cs
+++ b/src/code/ACRServerAPICalls.cs
@@ -154,7 +154,9 @@ namespace Microsoft.PowerShell.PSResourceGet
             List<JToken> allVersionsList = foundTags["tags"].ToList();
 
             SortedDictionary<NuGet.Versioning.SemanticVersion, string> sortedQualifyingPkgs = GetPackagesWithRequiredVersion(allVersionsList, VersionType.VersionRange, VersionRange.All, specificVersion: null, packageName, includePrerelease, out errRecord);
-            foreach(KeyValuePair<NuGet.Versioning.SemanticVersion, string> pkgVersionTag in sortedQualifyingPkgs.Reverse())
+            var pkgsInDescendingOrder = sortedQualifyingPkgs.Reverse();
+
+            foreach(var pkgVersionTag in pkgsInDescendingOrder)
             {
                 string exactTagVersion = pkgVersionTag.Value.ToString();
                 Hashtable metadata = GetACRMetadata(Registry, packageNameLowercase, exactTagVersion, acrAccessToken, out errRecord);
@@ -260,7 +262,9 @@ namespace Microsoft.PowerShell.PSResourceGet
             List<Hashtable> latestVersionResponse = new List<Hashtable>();
             List<JToken> allVersionsList = foundTags["tags"].ToList();
             SortedDictionary<NuGet.Versioning.SemanticVersion, string> sortedQualifyingPkgs = GetPackagesWithRequiredVersion(allVersionsList, VersionType.VersionRange, versionRange, specificVersion: null, packageName, includePrerelease, out errRecord);
-            foreach(KeyValuePair<NuGet.Versioning.SemanticVersion, string> pkgVersionTag in sortedQualifyingPkgs.Reverse())
+            var pkgsInDescendingOrder = sortedQualifyingPkgs.Reverse();
+
+            foreach(var pkgVersionTag in pkgsInDescendingOrder)
             {
                 string exactTagVersion = pkgVersionTag.Value.ToString();
                 Hashtable metadata = GetACRMetadata(Registry, packageName.ToLower(), exactTagVersion, acrAccessToken, out errRecord);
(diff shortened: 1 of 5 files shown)


src/code/ACRServerAPICalls.cs
```diff
@@ -1687,6 +1675,69 @@ private string CreateMetadataContent(string manifestFilePath, ResourceType resou
...
+                    }
+
+                    _cmdletPassedIn.WriteDebug($"'{packageName}' version parsed as '{pkgVersion}'");
+
+                    if (isSpecificVersionSearch)
+                    {
+                        if (pkgVersion.ToNormalizedString() == specificVersion.ToNormalizedString())
+                        {
+                            // NuGet.Versioning.SemanticVersion semanticVersionOfPkg = new SemanticVersion(pkgVersion.Major, pkgVersion.Minor, pkgVersion.Patch, pkgVersion.IsPrerelease ? pkgVersion.Release: String.Empty);
+                            // sortedPkgs.Add(semanticVersionOfPkg, pkgVersionElement.ToString());
```
adityapatwardhan:
```suggestion
```

src/code/ACRServerAPICalls.cs
```diff
@@ -1687,6 +1675,69 @@ private string CreateMetadataContent(string manifestFilePath, ResourceType resou
...
+                            sortedPkgs.Add(pkgVersion, pkgVersionElement.ToString());
+                            break;
+                        }
+                    }
+                    else
+                    {
+                        if (versionRange.Satisfies(pkgVersion) && (!pkgVersion.IsPrerelease || includePrerelease))
+                        {
+                            // NuGet.Versioning.SemanticVersion semanticVersionOfPkg = new SemanticVersion(pkgVersion.Major, pkgVersion.Minor, pkgVersion.Patch, pkgVersion.IsPrerelease ? pkgVersion.Release: String.Empty);
+                            // sortedPkgs.Add(semanticVersionOfPkg, pkgVersionElement.ToString());
```
adityapatwardhan:
```suggestion
```

src/code/ACRServerAPICalls.cs
```diff
@@ -1687,6 +1675,69 @@ private string CreateMetadataContent(string manifestFilePath, ResourceType resou
...
+                            sortedPkgs.Add(pkgVersion, pkgVersionElement.ToString());
+                        }
+                    }
+
```
adityapatwardhan:
```suggestion
```

src/code/ACRServerAPICalls.cs
```diff
@@ -1687,6 +1675,69 @@ private string CreateMetadataContent(string manifestFilePath, ResourceType resou
...
+                }
+            }
+
+            return sortedPkgs;
```
adityapatwardhan, line 1701: Maybe we should just reverse here as we always need it in descending order.

src/code/ACRServerAPICalls.cs
```diff
@@ -1687,6 +1675,69 @@ private string CreateMetadataContent(string manifestFilePath, ResourceType resou
...
+                            new InvalidOrEmptyResponse($"Response does not contain version element ('name') for package '{packageName}' in '{Repository.Name}'."),
+                            "FindNameFailure",
+                            ErrorCategory.InvalidResult,
+                            this);
+
+                        return null;
+                    }
+
+                    // determine if the package version that is a repository tag is a valid NuGetVersion
+                    if (!NuGetVersion.TryParse(pkgVersionElement.ToString(), out NuGetVersion pkgVersion))
```
adityapatwardhan: we seem to be doing `pkgVersionElement.ToString()` a few times. Can we just store it in a string as reuse the string.

src/code/ACRServerAPICalls.cs
```diff
@@ -143,59 +143,30 @@ public override FindResults FindName(string packageName, bool includePrerelease,
...
                 return new FindResults(stringResponse: new string[] { }, hashtableResponse: emptyHashResponses, responseType: acrFindResponseType);
             }
 
             List<Hashtable> latestVersionResponse = new List<Hashtable>();
             List<JToken> allVersionsList = foundTags["tags"].ToList();
-            allVersionsList.Reverse();
 
-            foreach (var pkgVersionTagInfo in allVersionsList)
+            SortedDictionary<NuGet.Versioning.SemanticVersion, string> sortedQualifyingPkgs = GetPackagesWithRequiredVersion(allVersionsList, VersionType.VersionRange, VersionRange.All, specificVersion: null, packageName, includePrerelease, out errRecord);
+            var pkgsInDescendingOrder = sortedQualifyingPkgs.Reverse();
```
adityapatwardhan: Null reference exception when an error happens in GetPackagesWithRequiredVersion and it returns null

src/code/ACRServerAPICalls.cs
```diff
@@ -143,59 +143,30 @@ public override FindResults FindName(string packageName, bool includePrerelease,
...
             {
                 return new FindResults(stringResponse: new string[] { }, hashtableResponse: emptyHashResponses, responseType: acrFindResponseType);
             }
 
             List<Hashtable> latestVersionResponse = new List<Hashtable>();
             List<JToken> allVersionsList = foundTags["tags"].ToList();
-            allVersionsList.Reverse();
 
-            foreach (var pkgVersionTagInfo in allVersionsList)
+            SortedDictionary<NuGet.Versioning.SemanticVersion, string> sortedQualifyingPkgs = GetPackagesWithRequiredVersion(allVersionsList, VersionType.VersionRange, VersionRange.All, specificVersion: null, packageName, includePrerelease, out errRecord);
```
adityapatwardhan: Check for errRecord befor doing anything

src/code/ACRServerAPICalls.cs
```diff
@@ -143,59 +143,30 @@ public override FindResults FindName(string packageName, bool includePrerelease,
...
-                        Hashtable metadata = GetACRMetadata(registry, packageNameLowercase, pkgVersion, acrAccessToken, out errRecord);
-                        if (errRecord != null || metadata.Count == 0)
-                        {
-                            return new FindResults(stringResponse: new string[] { }, hashtableResponse: emptyHashResponses, responseType: acrFindResponseType);
-                        }
-
-                        latestVersionResponse.Add(metadata);
-                        break;
-                    }
+                    return new FindResults(stringResponse: new string[] { }, hashtableResponse: emptyHashResponses, responseType: acrFindResponseType);
```
adityapatwardhan: Add a comment to explain what is happening here.

src/code/ACRServerAPICalls.cs
```diff
@@ -281,53 +252,28 @@ public override FindResults FindVersionGlobbing(string packageName, VersionRange
...
             {
                 return new FindResults(stringResponse: new string[] { }, hashtableResponse: emptyHashResponses, responseType: acrFindResponseType);
             }
 
             List<Hashtable> latestVersionResponse = new List<Hashtable>();
             List<JToken> allVersionsList = foundTags["tags"].ToList();
-            allVersionsList.Reverse();
-            foreach (var packageVersion in allVersionsList)
+            SortedDictionary<NuGet.Versioning.SemanticVersion, string> sortedQualifyingPkgs = GetPackagesWithRequiredVersion(allVersionsList, VersionType.VersionRange, versionRange, specificVersion: null, packageName, includePrerelease, out errRecord);
+            var pkgsInDescendingOrder = sortedQualifyingPkgs.Reverse();
```
adityapatwardhan: Same as above, check for errRecord and potential null reference exception

adityapatwardhan (changes requested):

test/PublishPSResourceTests/PublishPSResourceACRServer.Tests.ps1
```diff
@@ -371,10 +371,9 @@ Describe "Test Publish-PSResource" -tags 'CI' {
         $results[0].Version | Should -Be $scriptVersion
     }
 
-    <# This test does not work currently due to a bug if the last digit is 0. Link to issue: https://github.com/PowerShell/PSResourceGet/issues/1582
     It "Should publish a script without lines in between comment blocks locally" {
         $scriptName = "ScriptWithoutEmptyLinesBetweenCommentBlocks"
-        $scriptVersion = "1.0.0"
+        $scriptVersion = "1.0"
```
alerickson, line 376: Just out of curiosity, does it only work if version is a 2 digit number? Should we open an issue for this?

src/code/PSResourceInfo.cs
```diff
@@ -849,6 +852,36 @@ public sealed class PSResourceInfo
 
                     metadata["NormalizedVersion"] = parsedNormalizedVersion;
```
alerickson: This isn't a part of your changes, but just wondering if parsedNormalizedVersion should be parsedNormalizedVersion.ToNormalizedString(), because I don't think the value itself is normalized yet.

src/code/PSResourceInfo.cs
```diff
@@ -849,6 +852,36 @@ public sealed class PSResourceInfo
...
+                    if (!NuGetVersion.TryParse(versionValue, out NuGetVersion parsedNormalizedVersion) && pkgVersion == null)
+                    {
+                        errorMsg = string.Format(
+                            CultureInfo.InvariantCulture,
+                            @"TryConvertFromACRJson: Cannot parse NormalizedVersion or System.Version from version in metadata.");
+
+                        return false;
+                    }
+
+                    metadata["NormalizedVersion"] = parsedNormalizedVersion;
```
alerickson: Same question as above here, should this be parsedNormalizedVersion.ToNormalizedString()?

src/code/ACRServerAPICalls.cs
```diff
@@ -365,19 +314,31 @@ public override FindResults FindVersion(string packageName, string version, Reso
...
+            }
 
-            if (errRecord != null)
+            List<Hashtable> latestVersionResponse = new List<Hashtable>();
+            List<JToken> allVersionsList = foundTags["tags"].ToList();
+
+            SortedDictionary<NuGet.Versioning.SemanticVersion, string> sortedQualifyingPkgs = GetPackagesWithRequiredVersion(allVersionsList, VersionType.SpecificVersion, VersionRange.All, requiredVersion, packageName, includePrereleaseVersions, out errRecord);
+            var pkgsInDescendingOrder = sortedQualifyingPkgs.Reverse();
+
+            foreach(var pkgVersionTag in pkgsInDescendingOrder)
```
alerickson: Lines 327-339 happens get repeated 2 other times in this file, consider separating out into a method that can be called.

src/code/ACRServerAPICalls.cs
```diff
@@ -711,10 +687,22 @@ internal Hashtable GetACRMetadata(string registry, string packageName, NuGetVers
                     return requiredVersionResponse;
                 }
 
-                if (!NuGetVersion.TryParse(pkgVersionElement.ToString(), out NuGetVersion pkgVersion))
+
```
alerickson:
```suggestion
```

src/code/ACRServerAPICalls.cs
anamnavi: do you mean leave a comment about why we return from the error case? or a comment about why we reverse order of packages?

src/code/ACRServerAPICalls.cs
anamnavi: Since `.Reverse()` is a LINQ method the return type of the method would then become `IEnumerable<KeyValuePair<NuGet.Versioning.SemanticVersion, string>>` ...is that too messy to have? If not I can change the method return type and return a reversed collection.

src/code/PSResourceInfo.cs
anamnavi: You're right, good catch! I added that fix. I also factored out some of the shared code.

src/code/ACRServerAPICalls.cs
adityapatwardhan: Keep it as is.

src/code/ACRServerAPICalls.cs
adityapatwardhan: Do we always return an empty FindResult response when we dont find one. If it is consistent, then its ok. Maybe instead of a new FindResults() everytime have a static one.

adityapatwardhan (approved):

src/code/ACRServerAPICalls.cs
anamnavi: We always return an empty FindResult, so this is consistent. But I agree having a static one would be way better. I can do that in a separate PR.

alerickson (approved):

PR #1594, opened 2024-03-11
ACR: Add resource type to returned PSResourceInfo object

### LINKED ISSUE #1592: Return artifact info from Find/Install PSResource for ACR repos

Create a class for the metadata that gets returned from ACR servers.  Artifact type gets published to the server manifest, but does not get returned as a part of the `PSResourceInfo` object.

_No response_

alerickson:
# PR Summary
This PR retrieves the `resourceType` metadata in the `annotations` section of a package manifest for ContainterRegistry servers and parses it into the `PSResourceInfo` object returned by `Find-PSResource`.

It also changes all instances of "artifactType" to "resourceType" to be consistent with naming. 

## PR Context
Resolves #1592 

## PR Checklist

    - [ ] None
    - **OR**
        - [ ] Issue filed: 
- **User-facing changes**
    - [ ] Not Applicable
    - **OR**
        - [ ] Issue filed: 
    - **OR**
- **Tooling**
    - **OR**


src/code/ContainerRegistryInfo.cs
```diff
@@ -0,0 +1,50 @@
...
+				{ "Metadata", Metadata },
+				{ "ResourceType", ResourceType }
+			};
+
+            return hashtable;
+		}
+
+        #endregion
+    }
+}
```
anamnavi:
```suggestion
}

```

src/code/ACRResponseUtil.cs
```diff
@@ -42,20 +42,24 @@ public override IEnumerable<PSResourceResult> ConvertToPSResourceResult(FindResu
...
 
-                foreach (DictionaryEntry entry in response)
+                if (!response.ContainsKey("Metadata"))
                 {
-                    packageName = (string)entry.Key;
-                    packageMetadata = (string)entry.Value;
+                    yield return new PSResourceResult(returnedObject: pkg, exception: new ConvertToPSResourceException("Error retrieving package metadata from response."), isTerminatingError: true);
                 }
 
+                ResourceType? resourceType = response.ContainsKey("ResourceType") ? response["ResourceType"] as ResourceType? : ResourceType.None;
```
anamnavi, line 56: I don't think we need to make `ResourceType resourceType` nullable. We only get a hashtable from `ContainerRegistryInfo.ToHashtable()` which will always have a entry for "ResourceType" with ResourceType.None being the default. So it should technically never be null, like the other keys' values.

src/code/PSResourceInfo.cs
```diff
@@ -810,6 +810,7 @@ public sealed class PSResourceInfo
         public static bool TryConvertFromACRJson(
           string packageName,
           JsonDocument packageMetadata,
+          ResourceType? resourceType,
```
anamnavi, line 813: I don't think ResourceType here needs to be nullable, see comment from before. It should default to ResourceType.None when the variable is created by the caller method

src/code/PSResourceInfo.cs
```diff
@@ -996,7 +997,7 @@ public sealed class PSResourceInfo
                     repository: repository.Name,
                     repositorySourceLocation: repository.Uri.ToString(),
                     tags: metadata["Tags"] as string[],
-                    type: ResourceType.None,
+                    type: resourceType ?? ResourceType.None,
```
anamnavi, line 1000: same comment about this not needing to be nullable

test/FindPSResourceTests/FindPSResourceACRServer.Tests.ps1
```diff
@@ -192,4 +192,13 @@ Describe 'Test HTTP Find-PSResource for ACR Server Protocol' -tags 'CI' {
         $res.Version | Should -Be "3.5.0"
         $res.Prerelease | Should -Be "alpha"
     }
+
+    It "Should find and return correct resource type" {
```
anamnavi: could you also add a test for script type? The package `testscript` should have the right type, thanks!

src/code/ACRResponseUtil.cs
anamnavi: we can do a similar check to ensure the hashtable has 'ResourceType' key. Again the way the ContainerRegistryInfo is created it should and will have all the keys. For the "name" and "metadata" keys it's possible that the values can be empty strings and that's really the only thing that's really needed to be checked for.

anamnavi (changes requested):

test/FindPSResourceTests/FindPSResourceACRServer.Tests.ps1
alerickson: Will add this test

src/code/ACRResponseUtil.cs
alerickson: Because it's casting from a Hashtable value it needs to be able to potentially handle a null value.  I tried just doing ResourceType and was forced to make it nullable.

src/code/ACRServerAPICalls.cs
```diff
@@ -563,18 +563,16 @@ internal Hashtable GetACRMetadata(string registry, string packageName, string ex
...
                 errRecord = new ErrorRecord(exception, "FindNameFailure", ErrorCategory.InvalidResult, this);
 
                 return requiredVersionResponse;
             }
 
-            string metadataPkgName = metadataTuple.Item1;
-            string metadata = metadataTuple.Item2;
             string pkgVersionString = String.Empty;
-            using (JsonDocument metadataJSONDoc = JsonDocument.Parse(metadata))
+            using (JsonDocument metadataJSONDoc = JsonDocument.Parse(serverPkgInfo.Metadata))
```
adityapatwardhan: This can throw.. we need to handle it.

adityapatwardhan (approved):

anamnavi (approved):
LGTM

Issue #1633, opened 2024-04-12
Unable to install packages with dependencies from Artifactory PSGallery mirror (NuGet v2)


sean-r-williams:
### Prerequisites

- [X] Write a descriptive title.
- [X] Make sure you are able to repro it on the [latest released version](https://www.powershellgallery.com/packages/Microsoft.PowerShell.PSResourceGet)
- [X] Search the existing issues.

### Steps to reproduce

* Configure Artifactory with a remote feed pointing at PSGallery
* Connect to said feed with PSResourceGet (using NuGet v2)
* Install a package with dependencies (e.g. `Microsoft.Graph.Teams`, which depends on `Microsoft.Graph.Authentication`)

### Expected behavior

```console
PS> Install-PSResource Microsoft.Graph.Teams -Repository Artifactory
PS> <# module installs without issue #>
```

### Actual behavior

```console
Install-PSResource: Package(s) 'microsoft.graph.teams' could not be installed from repository 'Artifactory-Remote'.
```

### Error details

```console
PS> Get-Error

Exception             :
    Type    : Microsoft.PowerShell.PSResourceGet.UtilClasses.ResourceNotFoundException
    Message : Package(s) 'microsoft.graph.teams' could not be installed from repository 'Artifactory-Remote'.
    HResult : -2146233088
TargetObject          : Microsoft.PowerShell.PSResourceGet.Cmdlets.InstallPSResource
CategoryInfo          : InvalidData: (Microsoft.PowerShel…s.InstallPSResource:InstallPSResource) [Install-PSResource], ResourceNotFoundException
FullyQualifiedErrorId : InstallPackageFailure,Microsoft.PowerShell.PSResourceGet.Cmdlets.InstallPSResource
InvocationInfo        :
    MyCommand        : Install-PSResource
    ScriptLineNumber : 1
    OffsetInLine     : 1
    HistoryId        : 4
    Line             : install-psresource microsoft.graph.teams -Repository Artifactory-Remote -debug -Reinstall
    Statement        : install-psresource microsoft.graph.teams -Repository Artifactory-Remote -debug -Reinstall
    PositionMessage  : At line:1 char:1
                       + install-psresource microsoft.graph.teams -Repository Artifactory-Remo …
                       + ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    InvocationName   : install-psresource
[...]

PR #1644, opened 2024-04-24
V2ServerAPICalls: fix unnecessary `and` in FindVersionGlobbing w/ Artifactory

sean-r-williams:
# PR Summary

This PR fixes a bug in `V2ServerAPICalls.FindVersionGlobbing()` wherein an `and` was being emitted unnecessarily.

Fixes #1633.

## PR Context

`FindVersionGlobbing()` currently builds query filters by repeatedly appending filter clauses to the URL. Most NuGet v2 server types expect `Id` to be provided in the `$filter` parameter, however Artifactory does not. As a result, the `filterQuery` string is still null at line 1153 and prepending an `and` yields an invalid filter.

## PR Checklist

- [X] [PR has a meaningful title](https://github.com/PowerShell/PowerShell/blob/master/.github/CONTRIBUTING.md#pull-request---submission)
    - Use the present tense and imperative mood when describing your changes
- [X] [Summarized changes](https://github.com/PowerShell/PowerShell/blob/master/.github/CONTRIBUTING.md#pull-request---submission)
- [ ] [Make sure all `.h`, `.cpp`, `.cs`, `.ps1` and `.psm1` files have the correct copyright header](https://github.com/PowerShell/PowerShell/blob/master/.github/CONTRIBUTING.md#pull-request---submission)
- [X] This PR is ready to merge and is not [Work in Progress](https://github.com/PowerShell/PowerShell/blob/master/.github/CONTRIBUTING.md#pull-request---work-in-progress).
    - If the PR is work in progress, please add the prefix `WIP:` or `[ WIP ]` to the beginning of the title (the `WIP` bot will keep its status check at `Pending` while the prefix is present) and remove the prefix when the PR is ready.
- **[Breaking changes](https://github.com/PowerShell/PowerShell/blob/master/.github/CONTRIBUTING.md#making-breaking-changes)**
    - [X] None
    - **OR**
    - [ ] [Documentation needed]()
        - [ ] Issue filed: 
- **User-facing changes**
    - [X] Not Applicable
    - **OR**
    - [ ] [Documentation needed]()
        - [ ] Issue filed: 
- **Testing - New and feature**
    - [X] N/A or can only be tested interactively
    - **OR**
    - [ ] [Make sure you've added a new test if existing tests do not effectively test the code changed](https://github.com/PowerShell/PowerShell/blob/master/.github/CONTRIBUTING.md#before-submitting)
- **Tooling**
    - [X] I have considered the user experience from a tooling perspective and don't believe tooling will be impacted.
[...]

PR #1645, opened 2024-04-24
Refactor V2/NuGetServerAPICalls to use object-oriented query/filter builder

### LINKED ISSUE #1643: Refactor V2ServerAPICalls/NuGetServerAPICalls to use object-oriented query builder (instead of string manipulation)

Currently, `V2ServerAPICalls` and `NuGetServerAPICalls` build NuGet v2 API call URLs through direct manipulation of the query URL. This has several limitations:
* Potential URL encoding errors
* Fencepost-class problems where combining operators (`and`, etc.) are emitted at inappropriate places (see #1633 for an example of this)

Moving to an object-oriented approach would make these problems easier to avoid altogether.

Refactor V2ServerAPICalls/NuGetServerAPICalls to:
* Utilize some type of container (`List<String>`, `HashSet<String>`, or similar) to hold query clauses, then build a formatted query with combining operators in the appropriate places
* Handle URL encoding through either:
  * Off-label use of `HttpValueCollection` to dynamically build a query string (with built-in URL-safe encoding)
  * Direct serialization of key-value pairs with `Uri.EscapeDataString`

sean-r-williams:
# PR Summary

This PR replaces the existing NuGet v2 API call-building mechanism based on string manipulation (as used in members of `V2ServerAPICalls` and `NuGetServerAPICalls`) with a separate pair of classes (`NuGetV2QueryBuilder` and `NuGetV2FilterBuilder`) that construct a query-string by aggregating user-supplied criteria.

Resolves #1643.

## PR Context

The current implementation of V2 API calling classes builds API call URLs (including parameters) by directly performing string-manipulation on the URL itself. This has several drawbacks:
* **Performance**: In the CLR, strings are immutable. Calling `+=` on a string means a completely new string is created (with the suffix appended), and the prior string remains in memory until GC. 
  * `NuGetV2FilterBuilder` uses `StringBuilder` to avoid costly reallocation of large strings.
* **Sanitization**: OData filters, being URL query parameters, must be URL-encoded to prevent unexpected behavior. Operating on the URL directly means each relevant method has to URL-encode, which adds boilerplate or security risks if methods mis-encode.
  * `NuGetV2QueryBuilder` uses a derivative of `System.Collections.Specialized.NameValueCollection` to automatically URL-encode any configured parameters.
  * This does not include user input sanitization at the OData level - see below.
* **Flexibility**: Different NuGet server implementations have different API call requirements, as seen by the quirk flag system used today. Different combinations of these quirk flags have caused manual combination of filter clauses to misbehave (#1633) when a clause previously expected to be "always present" is no longer present.
  * `NuGetV2FilterBuilder` joins filter clauses together after the full set is available, so there's no duplicated/inappropriately-placed joining operators.

There are some specific limitations in the current implementation:
* I had to add a project reference to `System.Web` to build against TFM `net472`, as we need `System.Web.HttpUtility`. To my knowledge, System.Web should be in the GAC on .NET 4.5 and above, and PowerShell 7 already has System.Web.HttpUtility.dll linked as-is.  
  I'm not an expert in MSBuild, however, so if there's a better way to ensure `HttpUtility` is available I'm all ears.
* `NuGetV2FilterBuilder` doesn't ensure that filter clauses are properly escaping OData-relevant syntax (parentheses, quotes, etc.). 
  I wanted to focus namely on the aggregation of clauses versus the clauses themselves in the first pass - now that we have purpose-built classes for these, it would be significantly easier to expose an interface for filter clauses (`INuGetV2FilterCriterion`?) with some inheritance for unary-/binary-operator criteria.

## PR Checklist

- [X] [PR has a meaningful title](https://github.com/PowerShell/PowerShell/blob/master/.github/CONTRIBUTING.md#pull-request---submission)
- [X] [Summarized changes](https://github.com/PowerShell/PowerShell/blob/master/.github/CONTRIBUTING.md#pull-request---submission)
- [X] [Make sure all `.h`, `.cpp`, `.cs`, `.ps1` and `.psm1` files have the correct copyright header](https://github.com/PowerShell/PowerShell/blob/master/.github/CONTRIBUTING.md#pull-request---submission)
- [X] This PR is ready to merge and is not [Work in Progress](https://github.com/PowerShell/PowerShell/blob/master/.github/CONTRIBUTING.md#pull-request---work-in-progress).
    - [X] None
    - **OR**
        - [ ] Issue filed: 
- **User-facing changes**
    - [X] Not Applicable
    - **OR**
        - [ ] Issue filed: 
    - **OR**
- **Tooling**
    - [X] I have considered the user experience from a tooling perspective and don't believe tooling will be impacted.
    - **OR**

### DIFF
--- a/src/code/V2QueryBuilder.cs
+++ b/src/code/V2QueryBuilder.cs
@@ -1,3 +1,6 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
 using System;
 using System.Collections;
 using System.Collections.Generic;
--- a/src/code/V2QueryBuilder.cs
+++ b/src/code/V2QueryBuilder.cs
@@ -19,6 +19,8 @@ namespace Microsoft.PowerShell.PSResourceGet.Cmdlets
 
         internal bool ShouldEmitEmptyFilter = false;
 
+        internal string SearchTerm;
+
         internal NuGetV2QueryBuilder()
         {
 
@@ -39,6 +41,10 @@ namespace Microsoft.PowerShell.PSResourceGet.Cmdlets
             if (FilterBuilder.CriteriaCount > 0 || ShouldEmitEmptyFilter)
             {
                 QueryParameters["$filter"] = FilterBuilder.BuildFilterString();
+            }
+
+            if (SearchTerm != null) {
+                QueryParameters["searchTerm"] = SearchTerm;
             }
 
             foreach (var parameter in AdditionalParameters)
--- a/src/code/V2QueryBuilder.cs
+++ b/src/code/V2QueryBuilder.cs
@@ -1,0 +1,146 @@
+using System;
+using System.Collections;
+using System.Collections.Generic;
+using System.Collections.Specialized;
+using System.Configuration.Assemblies;
+using System.IO;
+using System.Linq;
+using System.Text;
+using System.Web;
+
+namespace Microsoft.PowerShell.PSResourceGet.Cmdlets
+{
+    internal class NuGetV2QueryBuilder
+    {
+
+        internal Dictionary<string, string> AdditionalParameters { get; private set; }
+
+        internal NuGetV2FilterBuilder FilterBuilder { get; private set; }
+
+        internal bool ShouldEmitEmptyFilter = false;
+
+        internal NuGetV2QueryBuilder()
+        {
+
+            FilterBuilder = new NuGetV2FilterBuilder();
+            AdditionalParameters = new Dictionary<string, string> { };
+        }
+
+        internal NuGetV2QueryBuilder(Dictionary<string, string> parameters) : this()
+        {
+            AdditionalParameters = new Dictionary<string, string>(parameters);
+        }
+        internal string BuildQueryString()
+        {
+
+            var QueryParameters = HttpUtility.ParseQueryString("");
+
+
+            if (FilterBuilder.CriteriaCount > 0 || ShouldEmitEmptyFilter)
+            {
+                QueryParameters["$filter"] = FilterBuilder.BuildFilterString();
+            }
+
+            foreach (var parameter in AdditionalParameters)
+            {
+                QueryParameters[parameter.Key] = parameter.Value;
+            }
+
+            return QueryParameters.ToString();
+
+        }
+
+    }
+
+    internal class NuGetV2FilterBuilder
+    {
+
+        internal NuGetV2FilterBuilder()
+        {
+
+        }
+
+        private HashSet<String> FilterCriteria = new HashSet<String>();
+
+        /// <summary>
+        ///     Convert the builder's provided set of filter criteria into an OData-compatible <c>filter</c> string.
+        /// </summary>
+        /// <remarks>
+        ///     Criteria order is not guaranteed.
+        /// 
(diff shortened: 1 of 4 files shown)


sean-r-williams:
@microsoft-github-policy-service agree company="Bungie, Inc."

anamnavi:
@sean-r-williams thanks for creating this PR, it looks really thorough and the context was well explained :) The team will take a deeper look into it an get back by this week with any suggestions/questions/follow up.

sean-r-williams:
Thanks @anamnavi! It looks like CodeQL is falling for some reason, but I'm not authorized to view the workflow logs. I have a feeling this might be due to the new dependency on System.Web, but I'm unsure. 

If someone from your team is able to summarize the errors coming back from CodeQL, I might be able to resolve the errors myself.

sean-r-williams:
Scratch that, it sems I forgot to add a file when I renamed a method. If someone could re-run Azure Pipelines, that'd be appreciated.

anamnavi:
no worries! if that doesn't work I can take a look at the CodeQL errors/see if it has to do with things on our end.

sean-r-williams:
> @sean-r-williams thanks for creating this PR, it looks really thorough and the context was well explained :) The team will take a deeper look into it an get back by this week with any suggestions/questions/follow up.

@anamnavi Gentle reminder on this - the current implementation of v2 API calls blocks dependency installation on Artifactory. 

I'd like to make sure either this PR or #1644 are included in 1.0.5 (or whatever the next patch or minor release happens to be), and that said release is made available in the near future.

src/code/V2QueryBuilder.cs
```diff
@@ -0,0 +1,224 @@
...
+
+        /// <summary>
+        ///     Construct a new <seealso cref="NuGetV2QueryBuilder"/> with a user-specified collection of query parameters.
+        /// </summary>
+        /// <param name="parameters">
+        ///     The set of additional parameters to provide.
+        /// </param>
+        internal NuGetV2QueryBuilder(Dictionary<string, string> parameters) : this()
+        {
+            AdditionalParameters = new Dictionary<string, string>(parameters);
```
alerickson, line 60: Can we have `AdditionalParameters = parameters` to save on the new allocation of Dictionary?

src/code/V2ServerAPICalls.cs
```diff
@@ -827,12 +877,30 @@ private string FindAllFromTypeEndPoint(bool includePrerelease, bool isSearchingM
         {
             _cmdletPassedIn.WriteDebug("In V2ServerAPICalls::FindAllFromTypeEndPoint()");
             string typeEndpoint = _isPSGalleryRepo && !isSearchingModule ? "/items/psscript" : String.Empty;
-            string paginationParam = _isPSGalleryRepo ? $"&$orderby=Id desc&$inlinecount=allpages&$skip={skip}&$top=6000" : $"&$inlinecount=allpages&$skip={skip}&$top=6000";
+
+            var queryBuilder = new NuGetV2QueryBuilder(new Dictionary<string, string>{
+                { "$inlinecount", "allpages" },
+                { "$skip", skip.ToString()},
+                { "$top", "100"}
```
alerickson:
```suggestion
                { "$top", "6000"}
```

alerickson (approved):

PR #1682, opened 2024-07-31
Publish-PSResource: Pack and Push Feature

### LINKED ISSUE #1635: Allow for Publishing "Packed" Modules to PSGallery

# PSResourceGet Pack and Push
 
## Goals
 
1. Allow for publishing signed PowerShell modules as nupkgs to PSGallery and ACR (what defines a module? It needs a psd1)
2. Allow for publishing nuspec dependencies into ACR
3. Create a new cmd for packing a PowerShell module
   
## "Pack" design
 
`Compress-PSResource
        [-Path] <String[]>
        [-DestinationPath] <String>
        [-PassThru]
        [-WhatIf]
        [-Confirm]
        [<CommonParameters>]`
 
## "Push" design
 
Add an additional parameter set which takes a path to a nupkg
 
`Publish-PSResource -NupkgPath "myPathtoNupkg" -Repository [Container] [PSGallery] -APIkey -SkipDependencyCheck`

## Screening for Modules

At this time PSGallery will only support this feature for modules, not arbitrary nupkgs

_No response_

jshigetomi:
# PR Summary

Fixes: #1635 

Decomposed Publish-PSResource to Pack and Push model with an addition of Compress-PSResource command.

Compress-PSResource requires a path to a PS module/script and a destination path to save the resulting nupkg.

Added -Nupkg parameter to Publish-PSResource that pushes the .nupkg file to a specified repository.

Publish-PSResource retains old functionality

## PR Context

This change is to meet an SFI2 requirement to ensure signed packages. This PR enables publishing signed packages through Publish-PSResource by first packing (Compress-PSResource) then signing (User's Choice) then pushing(Publish-PSResource -Nupkg <PathToNupkg>.

## PR Checklist

    - [x] None
    - **OR**
        - [ ] Issue filed: 
- **User-facing changes**
    - [ ] Not Applicable
    - **OR**
        - [ ] Issue filed: 
    - **OR**
- **Tooling**
    - **OR**


test/PublishPSResourceTests/PublishPSResource.Tests.ps1
```diff
@@ -321,9 +320,25 @@ Describe "Test Publish-PSResource" -tags 'CI' {
...
+        Publish-PSResource -NupkgPath $expectedPath -Repository $testRepository2
+        $expectedPath = Join-Path -Path $script:repositoryPath2  -ChildPath "$script:PublishModuleName.$version.nupkg"
+        Get-ChildItem $script:repositoryPath2 | Should -Be $expectedPath
+    }
+
     <# The following tests are related to passing in parameters to customize a nuspec.
      # These parameters are not going in the current release, but is open for discussion to include in the future.
+
     It "Publish a module with -Nuspec" {
+        Wait-Debugger
```
alerickson: Can remove this line

test/PublishPSResourceTests/CompressPSResource.Tests.ps1
```diff
@@ -0,0 +1,179 @@
...
+        # sign the nupkg
+        $nupkgPath = Join-Path -Path $script:repositoryPath2 -ChildPath "$script:PublishModuleName.$version.nupkg"
+        Set-AuthenticodeSignature -FilePath $nupkgPath -Certificate $testCert
+
+        # Verify the file was signed
+        $signature = Get-AuthenticodeSignature -FilePath $nupkgPath
+        $signature.Status | Should -Be 'Valid'
+	}
+    #>
+}
```
alerickson:
```suggestion
}

```

src/code/CompressPSResource.cs
```diff
@@ -0,0 +1,71 @@
...
+
+        protected override void EndProcessing()
+        {
+            _psResourceHelper.PackResource();
+        }
+
+        #endregion
+
+    }
+}
```
alerickson:
```suggestion
}

```

src/code/CompressPSResource.cs
```diff
@@ -0,0 +1,71 @@
...
+
+namespace Microsoft.PowerShell.PSResourceGet.Cmdlets
+{
+    /// <summary>
+    /// Compresses a module, script, or nupkg to a designated repository.
+    /// </summary>
+    [Cmdlet(VerbsData.Compress,
+        "PSResource",
+        SupportsShouldProcess = true)]
+    [Alias("cmpres")]
```
alerickson: I think we should actually change "cmpres" to "cmres" (even thought imo the former is more obvious in what's an alias for).  
 
If you run `Get-Verb compress` you can see what PowerShell thinks the appropriate alias prefix should be (`cm`).

src/code/CompressPSResource.cs
```diff
@@ -0,0 +1,71 @@
...
+
+        #region Method Overrides
+
+        protected override void BeginProcessing()
+        {
+            // Create a respository store (the PSResourceRepository.xml file) if it does not already exist
+            // This is to create a better experience for those who have just installed v3 and want to get up and running quickly
+            RepositorySettings.CheckRepositoryStore();
+
+            _psResourceHelper = new PSResourceHelper(
```
alerickson: This should be called something more descriptive, like `CompressHelper`

alerickson: After looking at the other files, I think it should be called `PublishHelper` since `publish` uses all of these methods but `compress` only uses the Pack related methods.

src/code/CompressPSResource.cs
```diff
@@ -0,0 +1,71 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using Microsoft.PowerShell.PSResourceGet.UtilClasses;
+using System;
```
alerickson: Looks like all the `System` directives are no longer needed here

src/code/PublishPSResource.cs
```diff
@@ -2,21 +2,11 @@
 // Licensed under the MIT License.
 
 using Microsoft.PowerShell.PSResourceGet.UtilClasses;
-using NuGet.Commands;
-using NuGet.Common;
-using NuGet.Configuration;
-using NuGet.Packaging;
 using NuGet.Versioning;
```
alerickson: `NuGet.Versioning` is no longer needed as well

src/code/PSResourceHelper.cs
```diff
@@ -0,0 +1,1214 @@
...
+        }
+
+        #endregion
+
+        #region Public Methods
+
+        internal void PackResource() 
+        {
+            // Returns the name of the file or the name of the directory, depending on path
+            if (!_cmdlet.ShouldProcess(string.Format("{0} '{1}' from the machine", _callerCmdlet, resolvedPath)))
```
alerickson: I think this wording is a bit off, it looks something like:
`What if: Performing the operation "Compress-PSResource" on target "CompressPSResource 'C:\code\testPSResourceGet' from the machine".` right now

src/code/PSResourceHelper.cs
```diff
@@ -0,0 +1,1214 @@
...
+            Path = path;
+            DestinationPath = destinationPath;
+            SkipModuleManifestValidate = skipModuleManifestValidate;
+            _cancellationToken = cancellationToken;
+            _isNupkgPathSpecified = isNupkgPathSpecified;
+        }
+
+        #endregion
+
+        #region Public Methods
```
alerickson:
```suggestion
        #region Internal Methods
```

src/code/PSResourceHelper.cs
```diff
@@ -0,0 +1,1214 @@
...
+                    _cmdlet.WriteError(new ErrorRecord(
+                        new ArgumentException($"No file with a .psd1 extension was found in '{pathToModuleManifestToPublish}'. Please specify a path to a valid module manifest."),
+                        "moduleManifestNotFound",
+                        ErrorCategory.ObjectNotFound,
+                        this._cmdlet));
+
+                    return;
+                }
+
+                if (_callerCmdlet == CallerCmdlet.CompressPSResource)
```
alerickson: This is an open question: do we want to skip validating the module manifest if we're just packing the module?

src/code/PSResourceHelper.cs
```diff
@@ -0,0 +1,1214 @@
...
+                            new ArgumentException("Error occured while creating directory to publish: " + e.Message),
+                            "ErrorCreatingDirectoryToPublish",
+                            ErrorCategory.InvalidOperation,
+                            this._cmdlet));
+                    }
+                }
+
+                if (_callerCmdlet == CallerCmdlet.PublishPSResource)
+                {
+                    outputNupkgDir = System.IO.Path.Combine(outputDir, "nupkg");
```
alerickson: Instead of having lines 293 to 300, consider adding a parameter to `PackResource()` that has the value for `outputNupkgDir`

src/code/PSResourceHelper.cs
```diff
@@ -0,0 +1,1214 @@
...
+            {
+                _cmdlet.ThrowTerminatingError(new ErrorRecord(
+                    e,
+                    $"{_callerCmdlet}Error",
+                    ErrorCategory.NotSpecified,
+                    this._cmdlet));
+            }
+            finally
+            {
+                if(_callerCmdlet == CallerCmdlet.CompressPSResource)
```
alerickson: I think we always want to delete `outputDir` when we're done with it

src/code/PSResourceHelper.cs
```diff
@@ -0,0 +1,1214 @@
...
+                        _cmdlet.WriteError(pushNupkgContainerRegistryError);
+                        // exit out of processing
+                        return;
+                    }
+                }
+                else
+                {
+                    if(_isNupkgPathSpecified)
+                    {
+                        outputNupkgDir = pathToNupkgToPublish;
```
alerickson: Also consider having `ouputNupkgDir` as a parameter that's passed in to PushResource()

src/code/PSResourceHelper.cs
```diff
@@ -0,0 +1,1214 @@
...
+            }
+        }
+
+        #endregion
+
+        #region Private Methods
+
+        private bool PackNupkg(string outputDir, string outputNupkgDir, string nuspecFile, out ErrorRecord error)
+        {
+            _cmdlet.WriteDebug("In PublishPSResource::PackNupkg()");
```
alerickson: Update `PublishPSResource` to the name of this class.  This happens in a few other places as well

alerickson (changes requested):
Looks really good, just made some small suggestions here and there.

src/code/PSResourceHelper.cs
jshigetomi: The CompressPSResource path deletes it here. But the PublishPSResource path still needs it for push to preserve the previous functionality. The PushResource method finally block handles that.

src/code/CompressPSResource.cs
jshigetomi: I didn't know you could do that! I will use for any aliases I have to come up with in the future.

src/code/CompressPSResource.cs
jshigetomi: Confirm
*Are you sure you want to perform this action?
Performing the operation "Compress-PSResource" on target "CompressPSResource 'C:\Program
Files\PowerShell\7\Modules\PowerShellGet' from the machine".
[Y] Yes  [A] Yes to All  [N] No  [L] No to All  [S] Suspend  [?] Help (default is "Y"):

more like this maybe?

src/code/PSResourceHelper.cs
jshigetomi: Confirm
**Are you sure you want to perform this action?
Performing the operation "Compress-PSResource" on target "CompressPSResource 'C:\Program
Files\PowerShell\7\Modules\PowerShellGet' from the machine".**
[Y] Yes  [A] Yes to All  [N] No  [L] No to All  [S] Suspend  [?] Help (default is "Y"):

more like this maybe?

src/code/PSResourceHelper.cs
jshigetomi: I will remove this check because it is redundant. Compress-PSResource has no way of changing SkipModuleManifestValidate parameter because did not add it as a parameter.

I think adding an option to check for manifest could be useful to users who want a manifest file. Publish provides the option as well. It would be annoying if the user could only check with publish after they had already packed it.

alerickson: I guess there's a couple things here-- first whether validating the module manifest should happen or not.  I think it should by default always occur, but we can add a parameter to `Compress-PSResource` to skip validation, similar to `Publish-PSResource` (https://github.com/PowerShell/PSResourceGet/blob/d0e71b2cf558822c040143f76803d6c5efb65584/src/code/PublishPSResource.cs#L82).  

Second, if we by default always validate the module manifest then the default value for `SkipModuleManifestValidate` should be false so the module manifest is always checked unless otherwise specified not to (https://github.com/PowerShell/PSResourceGet/blob/5b64c5eb1ccd13074d42d0aaea48c2fa5f6d0f69/src/code/PSResourceHelper.cs#L59)

src/code/PSResourceHelper.cs
jshigetomi: I've added it to their unique constructors for the PublishHelper class.

outputDir is created in PSResource helper so I would have to create it Publish and Compress, then pass it to PackResource(). Then in the Publish path I would have to assign it as a member because pack and push both use it.

src/code/PSResourceHelper.cs
jshigetomi: passing outputNupkgDir makes it difficult to maintain old PublishResource Behavior. All the outputNupkgDir logic needs to be done in Publish and not in helper and that means in Compress as well.

adityapatwardhan:
Closing and re-opening to re-trigger CI

jshigetomi:
Fixed a test that was strangely passing on local machine but failing on git.

The problem was with 
>Get-ChildItem $script:repositoryPath2 | Should -Be $expectedPath
changed to 
>(Get-ChildItem $script:repositoryPath2).FullName | Should -Be $expectedPath

test/PublishPSResourceTests/PublishPSResource.Tests.ps1
```diff
@@ -567,6 +567,8 @@ Describe "Test Publish-PSResource" -tags 'CI' {
 
         $expectedPath = Join-Path -Path $script:repositoryPath  -ChildPath "$scriptName.$scriptVersion.nupkg"
         (Get-ChildItem $script:repositoryPath).FullName | Should -Be $expectedPath
+
+        Remove-Item -Path $scriptPath
```
alerickson: This should probably be moved into the "After All" or "After Each" (whichever is appropriate), just to isolate the test functionality itself and have clean up done outside of test

src/code/PublishHelper.cs
```diff
@@ -0,0 +1,1211 @@
...
+            }
+        }
+
+        #endregion
+
+        #region Private Methods
+
+        private bool PackNupkg(string outputDir, string outputNupkgDir, string nuspecFile, out ErrorRecord error)
+        {
+            _cmdlet.WriteDebug($"In {_callerCmdlet}::PackNupkg()");
```
alerickson: These debug messages should specify "PublishHelper" instead of "_callerCmdlet" to stay consistent with how we're using the debug/verbose messages elsewhere in the repo

test/PublishPSResourceTests/PublishPSResource.Tests.ps1
jshigetomi: This test doesn't delete the test.ps1 file it creates. It only happen for this test.
Maybe it can be like the first test where it removes it in the test?
```pwsh
    It "Publish module with required module not installed on the local machine using -SkipModuleManifestValidate" {
        # Skip the module manifest validation test, which fails from the missing manifest required module.
        $testModulePath = Join-Path -Path $TestDrive -ChildPath ModuleWithMissingRequiredModule
        Publish-PSResource -Path $testModulePath -Repository $testRepository2 -Confirm:$false -SkipDependenciesCheck -SkipModuleManifestValidate

        $expectedPath = Join-Path -Path $script:repositoryPath2  -ChildPath 'ModuleWithMissingRequiredModule.1.0.0.nupkg'
        $publishedModuleFound = Test-Path -Path $expectedPath
        $publishedModuleFound | Should -BeTrue

        if ($publishedModuleFound) {
            Remove-Item $expectedPath -Force -ErrorAction SilentlyContinue
        }
    }
```

src/code/PublishHelper.cs
jshigetomi: done. Should I also change the errors as well like...

```csharp
                    _cmdlet.WriteError(new ErrorRecord(
                        new ArgumentException($"The resource repository '{Repository}' is not a registered. Please run 'Register-PSResourceRepository' in order to publish to this repository."),
                        "RepositoryNotFound",
                        ErrorCategory.ObjectNotFound,
                        this._cmdlet));

                    return;
```

should this._cmdlet be instead this? that refers to the PublishHelper?

test/PublishPSResourceTests/PublishPSResource.Tests.ps1
alerickson: Oh, sorry I was just looking at the line kind of out of context.  Really the entire line should be removed because `$scriptPath` will be deleted during clean up in the "After Each" block already: https://github.com/PowerShell/PSResourceGet/blob/0b4d77fa43f905284acee01db56f12eea3d3e795/test/PublishPSResourceTests/PublishPSResource.Tests.ps1#L108

src/code/PublishHelper.cs
alerickson: You can just leave it as `this` without the `.cmdlet` (eg: https://github.com/PowerShell/PSResourceGet/blob/e79f6a242f2e90545457b04336b937a47cfcce42/src/code/FindHelper.cs#L110).  And looking at that example, could you change `_cmdlet` to `_cmdletPassedIn`?  Staying consistent with the naming convention just helps us in the future if there's changes we need to make to the entire code base.  It's easier to search and replace one term instead of looking up the different permutations

alerickson (approved):

src/code/.vs/Microsoft.PowerShell.PSResourceGet/v17/.suo
alerickson, line 1: Can you add this file back just to keep this PR clean with changes related to Publish/Compress?

test/testFiles/testScripts/test.ps1
alerickson, line 1: We actually don't need the script as a separate file, the cmdlet `New-PSScriptFileInfo` can create a new script.  The testScripts directory just contains invalid or complicated to create scripts (scenarios that would fail if trying to use `New-PSScriptFileInfo`)

alerickson: The docs for the cmdlet are here: https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.psresourceget/new-psscriptfileinfo?view=powershellget-3.x

jshigetomi: This is the test script that is created in that one test where I put remove-item. It was added here on accident when rebasing the PR. I will remove it.

src/code/.vs/Microsoft.PowerShell.PSResourceGet/v17/.suo
jshigetomi: Add the file back.

aluty:
@adityapatwardhan Checking back after a while.  I see "preview" has an official release now.

- [PowerShell Gallery | Microsoft.PowerShell.PSResourceGet 1.1.0](https://www.powershellgallery.com/packages/Microsoft.PowerShell.PSResourceGet/1.1.0) (2025-01-09)

Tested and documented somewhere?

PR #1730, opened 2024-10-20
Implement Group Policy settings for PSResource Repository

### LINKED ISSUE #1696: Group Policy for a PSRepository Allow List/Registered List

The group policy will register repositories as the only repositories allowed to be installed from if the policy is enabled. 
If the policy is enabled no other repositories will be allowed. 

The policy will take in a list of repositories, at minimum a name;uri pair or a full hashtable (or json may be preferred) of repository information. 

The policy will initially be Windows only but we will explore enabling on non-Windows via Intune policy and file system. 

The policy will be checked at each call to a repository, we will assume same privileges for all operations. 

There will also be a function/api that will expose if the policy is enabled and what repository Uris are available so that other tools which call PSResourceGet (such as WinGet) can integrate to this functionality. 

If an exact repo is specified that is not allowed it will be a terminating error. If no repo is specified all allowed repos will be seared and it will be a non-terminating error. 

image

adityapatwardhan:
# PR Summary

Implement Group Policy settings for PSResource Repository. 
Policy allows administrator to define an allow list for Uris that are allowed to be used for finding and installing resources.

The format for the policy is as follows:

`Name=PSGallery;Uri=https://www.powershellgallery.com/api/v2/`

## PR Context

Fixes #1696 

## PR Checklist

    - [x] None
    - **OR**
        - [X] Issue filed: https://github.com/MicrosoftDocs/PowerShell-Docs-PSGet/issues/163
- **User-facing changes**
    - [x] Not Applicable
    - **OR**
        - [ ] Issue filed: 
    - **OR**
- **Tooling**
    - **OR**


doBuild.ps1
```diff
@@ -34,9 +34,19 @@ function DoBuild
     Copy-Item -Path "./LICENSE" -Dest "$BuildOutPath"
 
     # Copy notice
-    Write-Verbose -Verbose -Message "Copying ThirdPartyNotices.txt to '$BuildOutPath'"
+    Write-Verbose -Verbose -Message "Copying InstallPSResourceGetPolicyDefinitions.ps1 to '$BuildOutPath'"
```
alerickson: Should this verbose logging be added to line 39 instead (ie leave the verbose messaging for coping the notice and add a new verbose statement for copying policy definitions)?

alerickson: Oh I see it already exists on line 41, so this statement should be reverted to what it previously was (verbose message for ThirdPartyNotices.txt)

src/code/ServerFactory.cs
```diff
@@ -1,11 +1,12 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
-using Microsoft.PowerShell.PSResourceGet.UtilClasses;
+using System;
 using System.Collections;
+using System.Linq;
```
alerickson: Are these 2 additions still needed?

alerickson (approved):

doBuild.ps1
adityapatwardhan: Updated it to be correct. `Notice.txt`

src/code/ServerFactory.cs
adityapatwardhan: Removed all unnecessary

PR #1750, opened 2024-11-11
Bugfix for local repo casing issue on Linux

anamnavi:
For the FindNameHelper() the regex was:
test_local_mod.\d+\.\d+\.\d+(?:[a-zA-Z0-9-.]+|.\d)?.nupkg

The first '.' matches any character whereas our intention with the regex is to match the literal '.' so I added a backslash before the '.' to escape it. If the package being searched for is `test_local_mod` it was matching test_local_mod.1.0.0.nupkg and also test_local_mod_2.1.0.0.nupkg where the '_' character was getting matched by '.' in regex 
 and `2.1.0.0` was getting considered as the version.

The main issue Amber had found was that for FindVersionHelper() the path for the .nupkg was being constructed and we use lowercasing so on Linux that path will not be correct. Instead, once we find a matching package (by name and version) we should use the `path` given from the file system that has the correct casing.
# PR Summary

## PR Checklist

    - [ ] None
    - **OR**
        - [ ] Issue filed: 
- **User-facing changes**
    - [ ] Not Applicable
    - **OR**
        - [ ] Issue filed: 
    - **OR**
- **Tooling**
    - **OR**

### DIFF
--- a/src/code/LocalServerApiCalls.cs
+++ b/src/code/LocalServerApiCalls.cs
@@ -260,7 +260,7 @@ namespace Microsoft.PowerShell.PSResourceGet.Cmdlets
             string actualPkgName = packageName;
 
             // this regex pattern matches packageName followed by a version (4 digit or 3 with prerelease word)
-            string regexPattern = $"{packageName}" + @"\.\d+\.\d+\.\d+(?:[a-zA-Z0-9-.]+|.\d)?.nupkg";
+            string regexPattern = $"{packageName}" + @"(\.\d+){1,3}(?:[a-zA-Z0-9-.]+|.\d)?\.nupkg";
             Regex rx = new Regex(regexPattern, RegexOptions.Compiled | RegexOptions.IgnoreCase);
             _cmdletPassedIn.WriteDebug($"package file name pattern to be searched for is: {regexPattern}");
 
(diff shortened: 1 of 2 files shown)


test/FindPSResourceTests/FindPSResourceLocal.Tests.ps1
```diff
@@ -60,6 +60,13 @@ Describe 'Test Find-PSResource for local repositories' -tags 'CI' {
         $res.Version | Should -Be "1.0.0"
     }
 
+    It "find resource given specific Name with incorrect casing and Version (should return correct casing)" {
+        # FindName()
```
anamnavi:
```suggestion
        # FindVersion()
```

src/code/LocalServerApiCalls.cs
```diff
@@ -260,7 +260,7 @@ private FindResults FindNameHelper(string packageName, string[] tags, bool inclu
             string actualPkgName = packageName;
 
             // this regex pattern matches packageName followed by a version (4 digit or 3 with prerelease word)
-            string regexPattern = $"{packageName}" + @".\d+\.\d+\.\d+(?:[a-zA-Z0-9-.]+|.\d)?.nupkg";
+            string regexPattern = $"{packageName}" + @"\.\d+\.\d+\.\d+(?:[a-zA-Z0-9-.]+|.\d)?.nupkg";
```
adityapatwardhan: Does this account for 2 digit versions

alerickson:
Other than the digit issue, everything else looks good.

src/code/LocalServerApiCalls.cs
anamnavi: it does now

alerickson:
Do we still need 
`Match match = matches[0];` on line 276
and
`Capture group = groups[0];` on line 284? 
 Looks like both variables aren't being used anywhere

alerickson (approved):

PR #1761, opened 2024-12-06
Modify filter query parameter

### LINKED ISSUE #1656: Unable to update PSResourceGet via Artifactory remote repo
### Prerequisites

- [X] Make sure you are able to repro it on the [latest released version](https://www.powershellgallery.com/packages/Microsoft.PowerShell.PSResourceGet)

### Steps to reproduce

* Configure Artifactory with a remote repository targeting PSGallery.
* Connect PSResourceGet to this repository. (Because PSGallery is v2-only, the connection to Artifactory within PSResourceGet must also be `v2`).
* Update PSResourceGet with `Update-PSResource Microsoft.PowerShell.PSResourceGet`.

### Expected behavior

```console
PS> Update-PSResource Microsoft.PowerShell.PSResourceGet
PS> # update completes without issues
```

### Actual behavior

```console
PS> update-psresource Microsoft.PowerShell.PSResourceGet -Repository Artifactory
update-psresource : Package does not exist on the server
At line:1 char:1
+ update-psresource Microsoft.PowerShell.PSResourceGet -Repository Arti ...
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : ObjectNotFound: (Microsoft.Power...lets.FindHelper:FindHelper) [Update-PSResource], ResourceNotFoundException
    + FullyQualifiedErrorId : FindNameConvertToPSResourceFailure,Microsoft.PowerShell.PSResourceGet.Cmdlets.UpdatePSResource
```

### Error details

_No response_

### Environment data

```powershell
ModuleType Version    PreRelease Name                                ExportedCommands
---------- -------    ---------- ----                                ----------------
Binary     1.0.4.1               Microsoft.PowerShell.PSResourceGet  {Find-PSResource, Get-InstalledPSResource, Get-PSResourceRepository, Get-PSScriptFileInfo…}


PSVersion                      7.4.2
PSEdition                      Core
GitCommitId                    7.4.2
OS                             Microsoft Windows 10.0.19045
Platform                       Win32NT
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0…}
PSRemotingProtocolVersion      2.3
SerializationVersion           1.1.0.1
WSManStackVersion              3.0
```

### Visuals

This seems to be the combination of two separate issues:
* PSResourceGet changes behavior in calls like `FindPackageById()` if we detect the upstream repo is a JFrog repo. After discussing with JFrog support, they confirmed that NuGet calls like this are being passed unfettered to the upstream. This means that PSGallery is receiving subtly different API call parameters when it's being reflected through Artifactory.
* PSGallery, if provided with the exact filter string `IsLatestVersion`, is returning unexpected results. I filed PowerShell/PowerShellGallery#273 to track this.

anamnavi:
Modify $filter query parameters IsLatestVersion and IsAbsoluteVersion to be `IsLatestVersion eq true` or `IsAbsoluteVersion eq true` for Artifactory to resolve Artifactory issue. For ADO Search() based queries including `eq true` returns a BadRequest error so it's not included.

Based on the API call Artifactory is performing as supplied here: https://github.com/PowerShell/PowerShellGallery/issues/273#issuecomment-2257318777

the count returns 1 but the response doesn't include a package, however including the `eq true` actually honors the filter and the response is successful in getting a package from PSGallery (from our testing).

# PR Summary

## PR Context
Resolves #1656 

## PR Checklist

    - [ ] None
    - **OR**
        - [ ] Issue filed: 
- **User-facing changes**
    - [ ] Not Applicable
    - **OR**
        - [ ] Issue filed: 
    - **OR**
- **Tooling**
    - **OR**

### DIFF
--- a/src/code/V2ServerAPICalls.cs
+++ b/src/code/V2ServerAPICalls.cs
@@ -913,7 +913,6 @@ namespace Microsoft.PowerShell.PSResourceGet.Cmdlets
             }
 
             // JFrog/Artifactory requires an empty search term to enumerate all packages in the feed
-            // It also deems 'IsLatestVersion eq true' and 'IsAbsoluteLatestVersion eq true' in the filter to be a bad request, so we use 'IsLatestVersion' or 'IsAbsoluteLatestVersion' only
             if (_isJFrogRepo) {
                 queryBuilder.SearchTerm = "''";
 
@@ -925,6 +924,7 @@ namespace Microsoft.PowerShell.PSResourceGet.Cmdlets
                 }
             }
             else {
+                // For ADO, 'IsLatestVersion eq true' and 'IsAbsoluteLatestVersion eq true' in the filter create a bad request error, so we use 'IsLatestVersion' or 'IsAbsoluteLatestVersion' only
                 if (includePrerelease) {
                     queryBuilder.AdditionalParameters["includePrerelease"] = "true";
                     filterBuilder.AddCriterion("IsAbsoluteLatestVersion");
@@ -962,13 +962,13 @@ namespace Microsoft.PowerShell.PSResourceGet.Cmdlets
                 queryBuilder.AdditionalParameters["$orderby"] = "Id desc";
             }
 
-            // JFrog/Artifactory deems 'IsLatestVersion eq true' and 'IsAbsoluteLatestVersion eq true' in the filter to be a bad request, so we use 'IsLatestVersion' or 'IsAbsoluteLatestVersion' only
             if (includePrerelease) {
                 queryBuilder.AdditionalParameters["includePrerelease"] = "true";
                 if (_isJFrogRepo) {
                     filterBuilder.AddCriterion("IsAbsoluteLatestVersion eq true");
                 }
                 else {
+                    // For ADO, 'IsLatestVersion eq true' and 'IsAbsoluteLatestVersion eq true' in the filter create a bad request error, so we use 'IsLatestVersion' or 'IsAbsoluteLatestVersion' only
                     filterBuilder.AddCriterion("IsAbsoluteLatestVersion");
                 }
             } else {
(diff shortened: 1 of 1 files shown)


alerickson (approved):

sean-r-williams:
I can confirm that this does mitigate the issue with Artifactory as described in #1656.

My one point of feedback here is that restricting this change to Artifactory might not fully capture the problem/solution at hand.

The "real" bug here is Gallery's poor handling of `IsLatestVersion` and so on (I think we already aligned on that in the other issues). This shows up in Artifactory because AF proxies NuGet v2 API calls unedited to upstream remotes. However, I don't think Artifactory is the only NuGet provider that does this (in the context of remote NuGet feeds). I provided more context in https://github.com/PowerShell/PSResourceGet/issues/1656#issuecomment-2547431238.

Some suggestions:
* Adjust the scope of this change from `_isJfrogRepo` to `!_isADORepo` - that is, turn it on for any package management provider where it doesn't actively break things
* Add some commentary to indicate that this isn't an Artifactory/JFrog problem, but rather "any PMP that passes queries upstream unedited" (Does Sonatype Nexus have this problem? Are there other PMPs customers use? How do they handle remote-feed proxying?)

anamnavi:
@sean-r-williams happy new year and my apologies for the late reply. Thank you for the detailed feedback, while I agree that this behavior could happen with other PMPs, our concern is that we don't have enough data on those other ones like Sonatype Nexus to enable this behavior for any PMP that is not ADO, nor test feeds to check if it would break scenarios on those feeds. I'll add a comment saying what Artifactory (and other PMPs) may do and we can come back to it as we get feedback from users using other PMPs.

PR #1769, opened 2025-01-03
Fix code scanning alert no. 5: Information exposure through transmitted data

adityapatwardhan:
Fixes [https://github.com/PowerShell/PSResourceGet/security/code-scanning/5](https://github.com/PowerShell/PSResourceGet/security/code-scanning/5)

To fix the problem, we need to ensure that sensitive information such as passwords is not transmitted in plain text. Instead, we should use secure methods to handle and transmit such data. One way to achieve this is by encrypting the sensitive information before transmission and decrypting it at the receiving end. Additionally, we should avoid logging or displaying sensitive information in error messages.

In this specific case, we will modify the code to encrypt the `content` variable before it is assigned to `request.Content`. We will also ensure that any sensitive information is not included in error messages.

_Suggested fixes powered by Copilot Autofix. Review carefully before merging._

### DIFF
--- a/src/code/ContainerRegistryServerAPICalls.cs
+++ b/src/code/ContainerRegistryServerAPICalls.cs
@@ -906,7 +906,8 @@ namespace Microsoft.PowerShell.PSResourceGet
                     return null;
                 }
 
-                request.Content = new StringContent(content);
+                string encryptedContent = EncryptContent(content);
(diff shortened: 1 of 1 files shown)


alerickson:
Closing, need to rethink approach for this issue.

PR #1948, opened 2026-02-21
Publish-PSResource: honor ExternalModuleDependencies from PrivateData.PSData during dependency validation

PrzemyslawKlys:
## Summary
  Fix `Publish-PSResource` dependency validation so `ExternalModuleDependencies` are respected for module manifests when declared under `PrivateData.PSData`.

  ## Problem
  `Publish-PSResource` excluded external dependencies only when `ExternalModuleDependencies` was present at the top level of parsed metadata.
  For module manifests, this value is typically stored under `PrivateData.PSData.ExternalModuleDependencies`, so the dependency check could still fail and incorrectly require
  `-SkipDependenciesCheck`.

  ## Root Cause
  `ParseRequiredModules()` only looked at:
  - `parsedMetadataHash["ExternalModuleDependencies"]`

  and did not also inspect:
  - `parsedMetadataHash["PrivateData"]["PSData"]["ExternalModuleDependencies"]`

  Also, dependency key matching was case-sensitive.

  ## Changes
  - Updated `ParseRequiredModules()` in `src/code/PublishHelper.cs` to:
    - read external module dependencies from both top-level metadata and `PrivateData.PSData`
    - use robust nested hashtable extraction
    - normalize matching via case-insensitive dependency map keys
  - Added regression test in:
    - `test/PublishPSResourceTests/PublishPSResource.Tests.ps1`
    - test name: `Publish a module with ExternalModuleDependencies that are not published`

  ## Validation
  - Added publish regression test covering module manifest external dependency behavior.

### DIFF
--- a/src/code/PublishHelper.cs
+++ b/src/code/PublishHelper.cs
@@ -1195,7 +1195,7 @@ namespace Microsoft.PowerShell.PSResourceGet.Cmdlets
             //  c. A string array of module names
             //  d. A single string module name
 
-            var dependenciesHash = new Hashtable();
+            var dependenciesHash = new Hashtable(StringComparer.OrdinalIgnoreCase);
             foreach (var reqModule in requiredModules)
             {
                 if (LanguagePrimitives.TryConvertTo<Hashtable>(reqModule, out Hashtable moduleHash))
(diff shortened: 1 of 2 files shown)


PrzemyslawKlys:
I've updated PR with additional test proving this is actually needed. Right now if you publish something with External Dependencies in place - you won't be able to download it without issues. This seems critical to me.

PrzemyslawKlys:
@anamnavi could you please review this - it's quite critical dependency issue that breaks any module that uses external dependencies on publish

alerickson:
Running tests here: https://github.com/PowerShell/PSResourceGet/pull/1972

PrzemyslawKlys:
@alerickson i checked the failing tests and they don't seem to be related to my work but pester itself. Am I supposed to fix it? Or should I ignore it?

PR #2040, opened 2026-09-14
(DSC) Idiomaticize resource schemas

michaeltlombardi:
# PR Summary

This change:

1. Removes `null` from the `type` keyword for optional resource properties that don't explicitly model `null` as a valid state.
1. Updates the `Microsoft.PowerShell.PSResourceGet/Repository` schema to define `required` at the top-level to indicate that `name` is always required and simplifies the `allOf` keyword for the remaining case (when `_exist` is `true` then `uri` is required).
1. Updates the serialization for the resource types to idiomatically omit optional undefined properties and consolidates standardized trace messaging for the serialization process.
1. Updates the `export` functionality to gracefully handle cases where a scope has zero installed PSResources instead of emitting an unhandled (non-terminating) error and returning invalid data that fails schema validation in DSC.

## PR Context

Prior to this change, there were a few issues with the DSC resource schemas regarding the idiomatic definitions and semantics of JSON schema:

1. Optional properties should not define the `type` keyword with `null` as a valid data type unless `null` is an explicitly modeled state for the property.

   In JSON Schema, the absence of a property in an object is semantically distinct from a property explicitly set to `null`.

   Instead, we control the optionality of properties through the `required` or `dependentRequired` keywords.
1. The schema for `Microsoft.PowerShell.PSResourceGet/Repository` didn't define `required` at the top-level. Instead, it used the `allOf` keyword to indicate whether the `name` and `uri` properties are required or just `name`.

   When a property is always required, we should hoist it to the top-level `required` keyword array. We can still use the `allOf` to conditionally extend the schema as needed.

   Parsing a JSON Schema is difficult for consumers, moreso when it requires unrolling schema composition. We should surface as much information to the consumer in the simplest form possible.

## PR Checklist

    - [x] None
    - **OR**
        - [ ] Issue filed: 
- **User-facing changes**
    - [x] Not Applicable
    - **OR**
        - [ ] Issue filed: 
    - **OR**
- **Tooling**
    - **OR**


Gijsreyn:
@michaeltlombardi - again, well done on the work you did here! I did a quick review, and yes, I agree with the direction. I checked the branch against what the DSC engine (v3.3.0) actually enforces and found a few things worth fixing before merging. Also, a small heads-up: it might need a rebase after a couple of PRs land in `master`. Two points:

1. `Repository` schema: name-only input becomes invalid (regression)

`"if": { "properties": { "_exist": { "const": true } } }` passes vacuously when `_exist` is absent, so `then` requires `uri` for every input that omits `_exist`. DSC validates the instance against the resource schema before `get`, `set` and `delete`, so these now fail (checked with `Test-Json`):

| Input | master | this PR |
| --- | --- | --- |
| `{"name":"x"}` (typical `dsc resource get` / `delete`) | valid | **invalid** |
| `{"name":"PSGallery","trusted":true}` | valid | **invalid** |
| `{"name":"x","_exist":true}` (no `uri`) | invalid | invalid |
| `{"name":"x","_exist":false}` | valid | valid |

Requiring `_exist` inside the `if` keeps your simplification and restores today's behaviour:

```json
"required": ["name"],
"allOf": [
  {
    "if": { "properties": { "_exist": { "const": true } }, "required": ["_exist"] },
    "then": { "required": ["uri"] }
  }
]
```

`uri` has to stay optional for `PSGallery` in particular: `Set-PSResourceRepository -Name PSGallery -Uri ...` is rejected by the cmdlet.

Minor:** for a repository that does not exist, `Repository.ToData()` still emits `trusted: false, priority: 0`. Since the PR is about omitting undefined state, returning only `name` + `_exist: false` . Shouldn't this be the idiomatic shape?

